PR #17, opened 2017-01-06
Rawspeed: DngDecoder: add support for deflate decompression

anarsoul:
Tested with files generated by hdrmerge, with this change darktable is able to import them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 option(WITH_PTHREADS "Enable threading support through pthread. Highly recommended" ON)
 option(WITH_OPENMP "Enable OpenMP support. Only used for tools, not the library itself." ON)
+option(WITH_ZLIB "Enable ZLIB support for DNG deflate support" ON)
 option(USE_XMLLINT "Run xmllint to test if data/cameras.xml is valid" ON)
 option(USE_IWYU "Run iwyu tool when compiling sources" OFF)
 option(BUILD_TESTING "Build the testing tree." ON)
--- a/RawSpeed/CMakeLists.txt
+++ b/RawSpeed/CMakeLists.txt
@@ -30,6 +30,17 @@ include_directories(SYSTEM ${Pugixml_INCLUDE_DIRS})
 find_package(JPEG REQUIRED)
 include_directories(SYSTEM ${JPEG_INCLUDE_DIRS})
 
+if (WITH_ZLIB)
+  find_package(ZLIB REQUIRED)
+  if(NOT ZLIB_FOUND)
+    message(SEND_ERROR "Did not find ZLIB! Either make it find ZLIB, or pass -DWITH_ZLIB=OFF to disable ZLIB.")
+  endif()
+  add_definitions(-DHAVE_ZLIB)
+  include_directories(SYSTEM ${ZLIB_INCLUDE_DIRS})
+else()
+  message(STATUS "ZLIB is disabled, DNG deflate support won't be available.")
+endif()
+
 CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)
 
 FILE(GLOB RAWSPEED_SOURCES
@@ -110,6 +121,10 @@ endif(WIN32)
 add_library(rawspeed OBJECT ${RAWSPEED_SOURCES})
 
 list(APPEND RAWSPEED_LIBS ${Pugixml_LIBRARIES} ${JPEG_LIBRARIES})
+
+if(WITH_ZLIB AND ZLIB_FOUND)
+  list(APPEND RAWSPEED_LIBS ${ZLIB_LIBRARIES})
+endif()
 
 if(WITH_PTHREADS AND Threads_FOUND)
   list(APPEND RAWSPEED_LIBS ${CMAKE_THREAD_LIBS_INIT})
--- a/RawSpeed/DngDecoder.cpp
+++ b/RawSpeed/DngDecoder.cpp
@@ -60,7 +60,13 @@ RawImage DngDecoder::decodeRawInternal() {
     try {
       isSubsampled = (*i)->getEntry(NEWSUBFILETYPE)->getInt() & 1; // bit 0 is on if image is subsampled
     } catch (TiffParserException) {}
-    if ((compression != 7 && compression != 1 && compression != 0x884c) || isSubsampled) {  // Erase if subsampled, or not JPEG or uncompressed
+    if (!(compression == 7 ||
+#ifdef HAVE_ZLIB
+          compression == 8 ||
+#endif
+          compression == 1 ||
+          compression == 0x884c) ||
+        isSubsampled) {  // Erase if subsampled, or not deflated, JPEG or uncompressed
       i = data.erase(i);
     } else {
       ++i;
@@ -81,6 +87,8 @@ RawImage DngDecoder::decodeRawInternal() {
   if (raw->hasEntry(SAMPLEFORMAT))
     sample_format = raw->getEntry(SAMPLEFORMAT)->getInt();
 
+  const int compression = raw->getEntry(COMPRESSION)->getShort();
+
   if (sample_format == 1)
     mRaw = RawImage::create(TYPE_USHORT16);
   else if (sample_format == 3)
@@ -98,8 +106,8 @@ RawImage DngDecoder::decodeRawInternal() {
   if (sample_format == 1 && bps > 16)
     ThrowRDE("DNG Decoder: Integer precision larger than 16 bits currently not supported.");
 
-  if (sample_format == 3 && bps != 32)
-    ThrowRDE("DNG Decoder: Float point must be 32 bits per sample.");
+  if (sample_format == 3 && bps != 32 && compression != 8)
+    ThrowRDE("DNG Decoder: Uncompressed float point must be 32 bits per sample.");
 
   try {
     mRaw->dim.x = raw->getEntry(IMAGEWIDTH)->getInt();
@@ -108,10 +116,8 @@ RawImage DngDecoder::decodeRawInternal() {
     ThrowRDE("DNG Decoder: Could not read basic image information.");
   }
 
-  int compression = -1;
-
   try {
-    compression = raw->getEntry(COMPRESSION)->getShort();
+
     if (mRaw->isCFA) {
 
       // Check if layout is OK, if present
@@ -229,17 +235,23 @@ RawImage DngDecoder::decodeRawInternal() {
       } catch (TiffParserException) {
         ThrowRDE("DNG Decoder: Unsupported format, uncompressed with no strips.");
       }
-    } else if (compression == 7 || compression == 0x884c) {
+    } else if (compression == 7 || compression == 8 || compression == 0x884c) {
       try {
         // Let's try loading it as tiles instead
 
-        mRaw->setCpp(raw->getEntry(SAMPLESPERPIXEL)->getInt());
         mRaw->createData();
 
-        if (sample_format != 1)
-           ThrowRDE("DNG Decoder: Only 16 bit unsigned data supported for compressed data.");
+        if (compression == 8 && sample_format != 3)
+           ThrowRDE("DNG Decoder: Only float format is supported for deflate-compressed data.");
+        else if (compression != 8 && sample_format != 1)
+           ThrowRDE("DNG Decoder: Only 16 bit unsigned data supported for JPEG-compressed data.");
 
         DngDecoderSlices slices(mFile, mRaw, compression);
+        if (raw->hasEntry(PREDICTOR)) {
+            uint32 predictor = raw->getEntry(PREDICTOR)->getInt();
+            slices.mPredictor = predictor;
+        }
+        slices.mBps = raw->getEntry(BITSPERSAMPLE)->getInt();
         if (raw->hasEntry(TILEOFFSETS)) {
           uint32 tilew = raw->getEntry(TILEWIDTH)->getInt();
           uint32 tileh = raw->getEntry(TILELENGTH)->getInt();
@@ -259,7 +271,7 @@ RawImage DngDecoder::decodeRawInternal() {
 
           for (uint32 y = 0; y < tilesY; y++) {
             for (uint32 x = 0; x < tilesX; x++) {
-              DngSliceElement e(offsets->getInt(x+y*tilesX), counts->getInt(x+y*tilesX), tilew*x, tileh*y);
+              DngSliceElement e(offsets->getInt(x+y*tilesX), counts->getInt(x+y*tilesX), tilew*x, tileh*y, tilew, tileh);
               e.mUseBigtable = tilew * tileh > 1024 * 1024;
               slices.addSlice(e);
             }
@@ -279,7 +291,7 @@ RawImage DngDecoder::decodeRawInternal() {
 
           uint32 offY = 0;
           for (uint32 s = 0; s < counts->count; s++) {
-            DngSliceElement e(offsets->getInt(s), counts->getInt(s), 0, offY);
+            DngSliceElement e(offsets->getInt(s), counts->getInt(s), 0, offY, mRaw->dim.x, yPerSlice);
             e.mUseBigtable = yPerSlice * mRaw->dim.y > 1024 * 1024;
             offY += yPerSlice;
 
--- a/RawSpeed/DngDecoderSlices.cpp
+++ b/RawSpeed/DngDecoderSlices.cpp
@@ -143,6 +143,207 @@ my_error_throw (j_common_ptr cinfo)
   ThrowRDE("JPEG decoder error!");
 } 
 
+#ifdef HAVE_ZLIB
+// decodeFPDeltaRow(): MIT License, copyright 2014 Javier Celaya <jcelaya@gmail.com>
+static inline void decodeFPDeltaRow(unsigned char *src, unsigned char *dst,
+                             size_t tileWidth, size_t realTileWidth,
+                             unsigned int bytesps, int factor) {
+  // DecodeDeltaBytes
+  for (size_t col = factor; col < realTileWidth * bytesps; ++col) {
+    src[col] += src[col - factor];
+  }
+  // Reorder bytes into the image
+  // 16 and 32-bit versions depend on local architecture, 24-bit does not
+  if (bytesps == 3) {
+    for (size_t col = 0; col < tileWidth; ++col) {
+      dst[col * 3] = src[col];
+      dst[col * 3 + 1] = src[col + realTileWidth];
+      dst[col * 3 + 2] = src[col + realTileWidth * 2];
+    }
+  } else {
+    if (getHostEndianness() == little) {
+      for (size_t col = 0; col < tileWidth; ++col) {
+        for (size_t byte = 0; byte < bytesps; ++byte)
+          dst[col * bytesps + byte] =
+              src[col + realTileWidth * (bytesps - byte - 1)];
+      }
+    } else {
+      for (size_t col = 0; col < tileWidth; ++col) {
+        for (size_t byte = 0; byte < bytesps; ++byte)
+          dst[col * bytesps + byte] = src[col + realTileWidth * byte];
+      }
+    }
+  }
+}
+
+static inline uint32 fp16ToFloat(ushort16 fp16) {
+  // IEEE-754-2008: binary16:
+  // bit 15 - sign
+  // bits 14-10 - exponent (5 bit)
+  // bits 9-0 - fraction (10 bit)
+  //
+  // exp = 0, fract = +-0: zero
+  // exp = 0; fract != 0: subnormal numbers
+  //                      equation: -1 ^ sign * 2 ^ -14 * 0.fraction
+  // exp = 1..30: normalized value
+  //              equation: -1 ^ sign * 2 ^ (exponent - 15) * 1.fraction
+  // exp = 31, fract = +-0: +-infinity
+  // exp = 31, fract != 0: NaN
+
+  uint32 sign = (fp16 >> 15) & 1;
+  uint32 fp16_exponent = (fp16 >> 10) & ((1 << 5) - 1);
+  uint32 fp16_fraction = fp16 & ((1 << 10) - 1);
+
+  // Normalized or zero
+  // binary32 equation: -1 ^ sign * 2 ^ (exponent - 127) * 1.fraction
+  // => exponent32 - 127 = exponent16 - 15, exponent32 = exponent16 + 127 - 15
+  uint32 fp32_exponent = fp16_exponent + 127 - 15;
+  uint32 fp32_fraction = fp16_fraction << (23 - 10); // 23 is binary32 fraction size
+
+  if (fp16_exponent == 31) {
+    // Infinity or NaN
+    fp32_exponent = 255;
+  } else if (fp16_exponent == 0 && fp16_fraction == 0) {
+    // +-Zero
+    fp32_exponent = 0;
+    fp32_fraction = 0;
+  } else if (fp16_exponent == 0 && fp16_fraction != 0) {
+    // Subnormal numbers
+    // binary32 equation: -1 ^ sign * 2 ^ (exponent - 127) * 1.fraction
+    // binary16 equation: -1 ^ sign * 2 ^ -14 * 0.fraction, we can represent it
+    // as a normalized value in binary32, we have to shift fraction until we get
+    // 1.new_fraction and decrement exponent for each shift
+    fp32_exponent = -14 + 127;
+    while (!(fp32_fraction & (1 << 23))) {
+      fp32_exponent -= 1;
+      fp32_fraction <<= 1;
+    }
+    fp32_fraction &= ((1 << 23) - 1);
+  }
+  return (sign << 31) | (fp32_exponent << 23) | fp32_fraction;
+}
+
+static inline uint32 fp24ToFloat(uint32 fp24) {
+  // binary24: Not a part of IEEE754-2008, but format is obvious,
+  // see https://en.wikipedia.org/wiki/Minifloat
+  // bit 23 - sign
+  // bits 22-16 - exponent (7 bit)
+  // bits 15-0 - fraction (16 bit)
+  //
+  // exp = 0, fract = +-0: zero
+  // exp = 0; fract != 0: subnormal numbers
+  //                      equation: -1 ^ sign * 2 ^ -62 * 0.fraction
+  // exp = 1..126: normalized value
+  //              equation: -1 ^ sign * 2 ^ (exponent - 63) * 1.fraction
+  // exp = 127, fract = +-0: +-infinity
+  // exp = 127, fract != 0: NaN
+
+  uint32 sign = (fp24 >> 23) & 1;
+  uint32 fp24_exponent = (fp24 >> 16) & ((1 << 7) - 1);
+  uint32 fp24_fraction = fp24 & ((1 << 16) - 1);
+
+  // Normalized or zero
+  // binary32 equation: -1 ^ sign * 2 ^ (exponent - 127) * 1.fraction
+  // => exponent32 - 127 = exponent24 - 64, exponent32 = exponent16 + 127 - 63
+  uint32 fp32_exponent = fp24_exponent + 127 - 63;
+  uint32 fp32_fraction = fp24_fraction << (23 - 16); // 23 is binary 32 fraction size
+
+  if (fp24_exponent == 127) {
+    // Infinity or NaN
+    fp32_exponent = 255;
+  } else if (fp24_exponent == 0 && fp24_fraction == 0) {
+    // +-Zero
+    fp32_exponent = 0;
+    fp32_fraction = 0;
+  } else if (fp24_exponent == 0 && fp24_fraction != 0) {
+    // Subnormal numbers
+    // binary32 equation: -1 ^ sign * 2 ^ (exponent - 127) * 1.fraction
+    // binary24 equation: -1 ^ sign * 2 ^ -62 * 0.fraction, we can represent it
+    // as a normalized value in binary32, we have to shift fraction until we get
+    // 1.new_fraction and decrement exponent for each shift
+    fp32_exponent = -62 + 127;
+    while (!(fp32_fraction & (1 << 23))) {
+      fp32_exponent -= 1;
+      fp32_fraction <<= 1;
+    }
+    fp32_fraction &= ((1 << 23) - 1);
+  }
+  return (sign << 31) | (fp32_exponent << 23) | fp32_fraction;
+}
+
+static inline void expandFP16(unsigned char *dst, int width) {
+  ushort16 *dst16 = (ushort16 *)dst;
+  uint32 *dst32 = (uint32 *)dst;
+
+  for (off_t x = width - 1; x >= 0; x--)
+    dst32[x] = fp16ToFloat(dst16[x]);
+}
+
+static inline void expandFP24(unsigned char *dst, int width) {
+  uint32 *dst32 = (uint32 *)dst;
+  dst += (width - 1) * 3;
+  for (off_t x = width - 1; x >= 0; x--) {
+    dst32[x] = fp24ToFloat((dst[0] << 16) | (dst[1] << 8) | dst[2]);
+    dst -= 3;
+  }
+}
+
+void DngDecoderSlices::decodeDeflate(const DngSliceElement &e,
+                                     unsigned char *uBuffer) {
+  if (!mFile->isValid(e.byteOffset, e.byteCount))
+    ThrowRDE("DngDecoderSlices::decodeDeflate: tile offset plus size is longer "
+             "than file. Truncated file.");
+
+  unsigned long dstLen = e.width * e.height * 4;
+  const unsigned char *cBuffer = mFile->getData(e.byteOffset, e.byteCount);
+  int predFactor = 0;
+
+  switch (mPredictor) {
+  case 3:
+    predFactor = 1;
+    break;
+  case 34894:
+    predFactor = 2;
+    break;
+  case 34895:
+    predFactor = 4;
+    break;
+  default:
+    predFactor = 0;
+    break;
+  }
+
+  int err = uncompress(uBuffer, &dstLen, cBuffer, e.byteCount);
+  if (err != Z_OK)
+    ThrowRDE("DngDecoderSlices::decodeDeflate: failed to uncompress tile: %d",
+             err);
+  int bytesps = mBps / 8;
+  size_t thisTileLength =
+      e.offY + e.height > (uint32)mRaw->dim.y ? mRaw->dim.y - e.offY : e.height;
+  size_t thisTileWidth =
+      e.offX + e.width > (uint32)mRaw->dim.x ? mRaw->dim.x - e.offX : e.width;
+  for (size_t row = 0; row < thisTileLength; ++row) {
+    unsigned char *src = uBuffer + row * e.width * bytesps;
+    unsigned char *dst = (unsigned char *)mRaw->getData() +
+                         ((e.offY + row) * mRaw->pitch +
+                          e.offX * sizeof(float) * mRaw->getCpp());
+    if (predFactor)
+      decodeFPDeltaRow(src, dst, thisTileWidth, e.width, bytesps, predFactor);
+    switch (bytesps) {
+    case 2:
+      expandFP16(dst, thisTileWidth);
+      break;
+    case 3:
+      expandFP24(dst, thisTileWidth);
+      break;
+    case 4:
+    default:
+      // No need to expand FP32
+      break;
+    }
+  }
+}
+#endif
 
 void DngDecoderSlices::decodeSlice(DngDecoderThread* t) {
   if (compression == 7) {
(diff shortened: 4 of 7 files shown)


houz:
I wonder what the future policy wrt. external dependencies is? So far adding zlib was a no-go which is why we never added deflate support. Has that changed with us taking over? Or should we rather look for one of those minimalistic implementations that fit into a single file and can be bundled?

RawSpeed/DngDecoder.cpp
```diff
@@ -81,9 +81,16 @@ RawImage DngDecoder::decodeRawInternal() {
   if (raw->hasEntry(SAMPLEFORMAT))
     sample_format = raw->getEntry(SAMPLEFORMAT)->getInt();
 
-  if (sample_format == 1)
+  int compression = -1;
+  compression = raw->getEntry(COMPRESSION)->getShort();
+  if (sample_format == 3 && bps != 32 && compression != 8)
+    ThrowRDE("DNG Decoder: Float point must be 32 bits per sample.");
+
+  // FIXME: We're creating USHORT16 image for all deflate images
```
LebedevRI: That's not rawspeed's problem. Just provide non-normalized float image.

RawSpeed/DngDecoder.cpp
```diff
@@ -81,9 +81,16 @@ RawImage DngDecoder::decodeRawInternal() {
   if (raw->hasEntry(SAMPLEFORMAT))
     sample_format = raw->getEntry(SAMPLEFORMAT)->getInt();
 
-  if (sample_format == 1)
+  int compression = -1;
+  compression = raw->getEntry(COMPRESSION)->getShort();
```
LebedevRI: And if there is no such tag?

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,159 @@ my_error_throw (j_common_ptr cinfo)
...
+    } else {
+    for (size_t col = 0; col < tileWidth; ++col) {
+      for (size_t byte = 0; byte < bytesps; ++byte)
+        dst[col*bytesps + byte] = src[col + realTileWidth*byte];
+        }
+      }
+    }
+}
+
+// From DNG SDK dng_utils.h
```
LebedevRI: And what is the legal status of such codedrop?

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -160,6 +313,27 @@ void DngDecoderSlices::decodeSlice(DngDecoderThread* t) {
...
       }
     }
+    /* Deflate compression */
+  } else if (compression == 8) {
+    unsigned char *uBuffer = NULL;
+    float *floatBuffer = NULL;
+    while (!t->slices.empty()) {
+      DngSliceElement e = t->slices.front();
+      if (!uBuffer)
+        uBuffer = new unsigned char [e.width * e.height * sizeof(float)];
```
LebedevRI: Let's not take any chances and reorder it like: `sizeof(float) * e.width * e.height`

RawSpeed/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ include_directories(SYSTEM ${Pugixml_INCLUDE_DIRS})
 find_package(JPEG REQUIRED)
 include_directories(SYSTEM ${JPEG_INCLUDE_DIRS})
 
+find_package(ZLIB REQUIRED)
```
LebedevRI: @houz is right. But now, it is up to me i guess.
I consider zlib to be ok as a dep. I may need to revisit this once i finally start to look at fuzzing, but for now it is ok.

What i do however think, is that it should be an *option*, i.e. it should be able to pass `-DWITH_ZLIB=OFF` to cmake, and that code will not be compiled.

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,159 @@ my_error_throw (j_common_ptr cinfo)
   ThrowRDE("JPEG decoder error!");
 } 
 
+static void decodeFPDeltaRow(unsigned char * src, unsigned char * dst, size_t tileWidth, size_t realTileWidth, unsigned int bytesps, int factor) {
```
LebedevRI: Where this is from?

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,159 @@ my_error_throw (j_common_ptr cinfo)
...
+  int bytesps = mBps / 8;
+  size_t thisTileLength = e.offY + e.height > (uint32)mRaw->dim.y ? mRaw->dim.y - e.offY : e.height;
+  size_t thisTileWidth = e.offX + e.width > (uint32)mRaw->dim.x ? mRaw->dim.x - e.offX : e.width;
+  for (size_t row = 0; row < thisTileLength; ++row) {
+    unsigned char * src = uBuffer + row * e.width * bytesps;
+    if (predFactor)
+      decodeFPDeltaRow(src, (unsigned char *)floatBuffer, thisTileWidth, e.width, bytesps, predFactor);
+    expandFloats((unsigned char *)floatBuffer, thisTileWidth, bytesps);
+    unsigned short *dst = (unsigned short *)((unsigned char *)mRaw->getData() + ((e.offY + row) * mRaw->pitch + e.offX * sizeof(ushort16) * mRaw->getCpp()));
+    // FIXME: convert float to ushort16, because darktable doesn't support non-normalized float32 images
```
LebedevRI: Nope.

LebedevRI (changes requested):
Looks, promising, i'm not against zlib as a dependency.

RawSpeed/DngDecoder.cpp
anarsoul: I tried, darktable can't handle it - it displays it all white.

LebedevRI: I know.

LebedevRI: (i'll handle the darktable-side patch)

RawSpeed/DngDecoderSlices.cpp
anarsoul: From RawTherapee, dcraw.cc, it's license says:

   No license is required to download and use dcraw.c.  However,
   to lawfully redistribute dcraw, you must either (a) offer, at
   no extra charge, full source code* for all executable files
   containing RESTRICTED functions, (b) distribute this code under
   the GPL Version 2 or later, (c) remove all RESTRICTED functions,
   re-implement them, or copy them from an earlier, unrestricted
   Revision of dcraw.c, or (d) purchase a license from the author.

I chose option (c), since these functions are not restricted.

RawSpeed/DngDecoderSlices.cpp
anarsoul: From DNG SDK license agreement:

1. LICENSE GRANT
Software License.  Subject to the restrictions below and other terms of this Agreement, Adobe hereby grants you a non-exclusive, worldwide, royalty free license to use, reproduce, prepare derivative works from, publicly display, publicly perform, distribute and sublicense the Software for any purpose.

2. RESTRICTIONS AND OWNERSHIP
You will not remove any copyright or other notice included in the Software or Documentation and you will include such notices in any copies of the Software that you distribute in human-readable format.

I'll move it to another file and will add Adobe copyright note. Looks like RawTherapee violates Adobe license.

RawSpeed/DngDecoderSlices.cpp
anarsoul: OK

RawSpeed/CMakeLists.txt
anarsoul: OK

RawSpeed/DngDecoderSlices.cpp
LebedevRI: Can such code even be used in LGPL library?

RawSpeed/DngDecoderSlices.cpp
LebedevRI: TBN, it is not from `dcraw.c`
https://github.com/Beep6581/RawTherapee/blob/880d0305486d8d9f6a4bb71c9f42a45921bfe48e/rtengine/dcraw.patch#L3173

anarsoul: It's from dcraw.cc, not from dcraw.c

RawSpeed/DngDecoderSlices.cpp
anarsoul: Both RawTherapee and hdrmerge use parts of DNG SDK, but of course it's not an excuse for darktable. As far as I can tell, DNG SDK license doesn't contradict LGPL, but I'm not a lawyer, so I can't tell for sure.

RawSpeed/DngDecoderSlices.cpp
LebedevRI: https://github.com/jcelaya/hdrmerge/issues/103#issuecomment-270953456
So `MIT License, copyright 2014 Javier Celaya <jcelaya@gmail.com>`
Now, can it be used in with LGPL?

anarsoul: Yes, it can be relicensed under LGPL, explanation is here:
http://softwareengineering.stackexchange.com/questions/105912/can-you-change-code-distributed-under-the-mit-license-and-re-distribute-it-unde

LebedevRI: The GPL and LGPL licenses are not synonyms, they are 2 different things, and i see nothing about LGPL on that page.
But yes, i do think that this MIT code can be used.
Do put `MIT License, copyright 2014 Javier Celaya <jcelaya@gmail.com>` as comment before the function.

anarsoul: Of course they are not. From the link above:

> Note that the MIT license explicitly includes the right to sublicense. Quoting: "the rights to use, copy, modify, merge, publish,distribute, sublicense, and/or sell"

> When I sublicense code, I cannot grant rights that I didn't originally have. In the case of the GPL, I am explicitly forbidden to sublicense only some rights. But neither in law nor in the MIT license do I have an obligation to sublicense all rights as a whole.

LGPL is more permissive than GPL, i.e. you can sublicense more rights that you already have. So I see no problem in sublicensing the code under LGPL.

LebedevRI: OK.

RawSpeed/DngDecoderSlices.cpp
LebedevRI: I'd personally try really hard to find a better-licensed version of this function.
After all, it looks like it simply converts half-float to full float.
Probably OpenEXR has such a function.

anarsoul: I agree. Will do.

RawSpeed/DngDecoderSlices.cpp
anarsoul: Multiplication is commutative, what's the difference?

anarsoul:
I rewrote fp-conversion functions from scratch and addressed all other comments, please review it again.

anarsoul:
Please note that I tested it only with float to ushort16 conversion patch.

RawSpeed/DngDecoderSlices.cpp
LebedevRI: At least in C, `typeof(sizeof())` is `size_t`, but are `width` and `height` of `size_t` type too?
If not, the multiplication can somewhat easily overflow, i believe.

anarsoul: Fair enough

anarsoul: Fixed and pushed.

RawSpeed/DngDecoder.cpp
```diff
@@ -60,7 +60,7 @@ RawImage DngDecoder::decodeRawInternal() {
     try {
       isSubsampled = (*i)->getEntry(NEWSUBFILETYPE)->getInt() & 1; // bit 0 is on if image is subsampled
     } catch (TiffParserException) {}
-    if ((compression != 7 && compression != 1 && compression != 0x884c) || isSubsampled) {  // Erase if subsampled, or not JPEG or uncompressed
+    if ((compression != 7 && compression != 8 && compression != 1 && compression != 0x884c) || isSubsampled) {  // Erase if subsampled, or not deflated, JPEG or uncompressed
```
LebedevRI: If without ZLIB, deflated should be erased.

RawSpeed/DngDecoder.cpp
```diff
@@ -230,17 +234,23 @@ RawImage DngDecoder::decodeRawInternal() {
       } catch (TiffParserException) {
         ThrowRDE("DNG Decoder: Unsupported format, uncompressed with no strips.");
       }
-    } else if (compression == 7 || compression == 0x884c) {
+    } else if (compression == 7 || compression == 8 || compression == 0x884c) {
```
LebedevRI, line 238: If without ZLIB, deflated should be ignored.

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,204 @@ my_error_throw (j_common_ptr cinfo)
   ThrowRDE("JPEG decoder error!");
 } 
 
+#ifdef HAVE_ZLIB
+static void decodeFPDeltaRow(unsigned char * src, unsigned char * dst, size_t tileWidth, size_t realTileWidth, unsigned int bytesps, int factor)
```
LebedevRI: no i mean, put the exact quote from author right before this function
it should be something like 
```c++
// decodeFPDeltaRow(): MIT License, copyright 2014 Javier Celaya <jcelaya@gmail.com>
static void decodeFPDeltaRow(unsigned char * src, unsigned char * dst, size_t tileWidth, size_t realTileWidth, unsigned int bytesps, int factor)
```

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,204 @@ my_error_throw (j_common_ptr cinfo)
...
+    } else {
+    for (size_t col = 0; col < tileWidth; ++col) {
+      for (size_t byte = 0; byte < bytesps; ++byte)
+        dst[col*bytesps + byte] = src[col + realTileWidth*byte];
+        }
+      }
+    }
+}
+
+static uint32_t fp16ToFloat(uint16_t fp16)
```
LebedevRI: Awesome, no more adobe code with questionable license!

RawSpeed/StdAfx.h
```diff
@@ -107,6 +107,9 @@ extern "C" {
 #include <list>
 using namespace std;
 
+// ZLIB
+#include <zlib.h>
```
LebedevRI, line 112:
```
#ifdef HAVE_ZLIB
```

LebedevRI (changes requested):
Getting there :)

CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 option(WITH_PTHREADS "Enable threading support through pthread. Highly recommended" ON)
 option(USE_XMLLINT "Run xmllint to test if data/cameras.xml is valid" ON)
 option(BUILD_TESTING "Build the testing tree." ON)
+option(WITH_ZLIB "Enable ZLIB support for DNG deflate support" ON)
```
LebedevRI: While there, please move this new line to after first two `option(WITH_`
Just OCD :)

RawSpeed/DngDecoder.cpp
anarsoul: In my opinion that's meaningless, since it decoding will fail later, and all extra ifdefs make code readability worse, but OK.

RawSpeed/DngDecoder.cpp
anarsoul: OK

RawSpeed/DngDecoderSlices.cpp
anarsoul: OK

CMakeLists.txt
anarsoul: OK

RawSpeed/StdAfx.h
anarsoul: OK

RawSpeed/DngDecoder.cpp
LebedevRI: Hmm, now that i think the second time, for now just do https://github.com/darktable-org/rawspeed/pull/17/files/4f2bdbbf76cb8c23be314d0b76effafd397f206e#r95054866

anarsoul:
All comments are addressed now.

RawSpeed/DngDecoder.cpp
```diff
@@ -230,17 +238,23 @@ RawImage DngDecoder::decodeRawInternal() {
...
         // Let's try loading it as tiles instead
 
-        mRaw->setCpp(raw->getEntry(SAMPLESPERPIXEL)->getInt());
         mRaw->createData();
 
-        if (sample_format != 1)
-           ThrowRDE("DNG Decoder: Only 16 bit unsigned data supported for compressed data.");
+        if (compression == 8 && sample_format != 3)
+           ThrowRDE("DNG Decoder: Only float format is supported for deflate-compressed data.");
+        else if (compression != 8 && sample_format != 1)
```
LebedevRI, line 246: But `compression != 8` only says that it is not deflate compression, not that it is jpeg compression?

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,204 @@ my_error_throw (j_common_ptr cinfo)
   ThrowRDE("JPEG decoder error!");
 } 
 
+#ifdef HAVE_ZLIB
+// decodeFPDeltaRow(): MIT License, copyright 2014 Javier Celaya <jcelaya@gmail.com>
```
LebedevRI, line 147: Finally :)

RawSpeed/DngDecoderSlices.cpp
LebedevRI: Note that i did not actually check that the function is implemented correctly. One day this will get to have an unit-test...

RawSpeed/DngDecoder.cpp
anarsoul: We're in a branch for jpeg and deflate. What else it could be?

RawSpeed/DngDecoderSlices.cpp
anarsoul: I checked that it returns the same value as Adobe's implementation for all possible fp16 and fp24 values. The only difference is handling +-inf and NaN, Adobe converts them to +-max and 0 respectively, my implementation leaves them as is.

Anyway, unit-testing this code is pretty useless: for normalized values it's just a matter of testing how well compiler translates shifts to machine code, and for non-normalized it's essentially means reimplementing part of these functions in a test. The question is who will watch the watchman^Wtest ;)

LebedevRI: Potentially having NAN's in input data will seriously hurt all the further algorithms that e.g. blur

anarsoul: Handle it later in darktable? IIRC DNG 1.4 spec says nothing about clamping infinity or treating NaN as zero.

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,204 @@ my_error_throw (j_common_ptr cinfo)
...
+  // 16 and 32-bit versions depend on local architecture, 24-bit does not
+  if (bytesps == 3) {
+    for (size_t col = 0; col < tileWidth; ++col) {
+      dst[col*3] = src[col];
+      dst[col*3 + 1] = src[col + realTileWidth];
+      dst[col*3 + 2] = src[col + realTileWidth*2];
+    }
+  } else {
+    union X { unsigned int x; unsigned char c; };
+    if (((union X){1}).c) {
```
axxel: I suggest to use the existing `getHostEndianness()` from Common.h instead of introducing a new way to determine the endianness.

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,204 @@ my_error_throw (j_common_ptr cinfo)
...
+    while (!(fp32_fraction & (1 << 23))) {
+      fp32_exponent -= 1;
+      fp32_fraction <<= 1;
+    }
+    fp32_fraction &= ((1 << 23) - 1);
+  }
+  return (sign << 31) | (fp32_exponent << 23) | fp32_fraction;
+}
+
+static uint32_t fp24ToFloat(uint32_t fp24)
```
axxel: Question: have you tried if hinting your compiler to `inline` this function helps him to actually to it? Removing a function call in the innermost loop might be worth it.

axxel:
As a general comment regarding the used types (`uint32_t`, etc.): I would suggest to stick to the custom types defined in Common.h:101ff just to be consistent with the rest of the code base.

Note: In my opinion, the chosen names _suck_ and I will definitely suggest to rename them later. Maybe exactly to the C99 standard types used here, but maybe to something more concise (I like the type names in rust a lot). Anyway, having it all consistent is a value in itself.

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,204 @@ my_error_throw (j_common_ptr cinfo)
...
+      dst[col*3 + 1] = src[col + realTileWidth];
+      dst[col*3 + 2] = src[col + realTileWidth*2];
+    }
+  } else {
+    union X { unsigned int x; unsigned char c; };
+    if (((union X){1}).c) {
+    for (size_t col = 0; col < tileWidth; ++col) {
+      for (size_t byte = 0; byte < bytesps; ++byte)
+        dst[col*bytesps + byte] = src[col + realTileWidth*(bytesps-byte-1)];  // Little endian
+    }
```
LebedevRI: Mis-indented `}`

axxel: it's actually the whole block of 4 lines that is misindented. same for the `else` block. but: a general clang-format whitespace cleanup will follow later anyway, right? (but only **after** my infrastructure rework! :))

RawSpeed/DngDecoderSlices.cpp
LebedevRI: Currently we simply assume that the [float] input is always ok, and only special-case the code inside the pipe that produces NAN's to not produce them.

RawSpeed/DngDecoder.cpp
LebedevRI: Aha, okay

RawSpeed/DngDecoderSlices.cpp
LebedevRI: While i do understand that this is basically a vendordrop from rt/jcelaya, i have to agree, please use `getHostEndianness()` here.

RawSpeed/DngDecoderSlices.cpp
LebedevRI: And since this is just one commit, also *please* *try* to [format](https://github.com/darktable-org/rawspeed/blob/develop/.clang-format) your new functions properly.
Only these few new functions, not the whole file/etc.

RawSpeed/DngDecoderSlices.cpp
anarsoul: I don't see anything that handles NaN's for uncompressed data (compression == 1) in rawspeed.

RawSpeed/DngDecoderSlices.cpp
```diff
@@ -143,6 +143,204 @@ my_error_throw (j_common_ptr cinfo)
...
+  if (err != Z_OK)
+    ThrowRDE("DngDecoderSlices::decodeDeflate: failed to uncompress tile: %d", err);
+  int bytesps = mBps / 8;
+  size_t thisTileLength = e.offY + e.height > (uint32)mRaw->dim.y ? mRaw->dim.y - e.offY : e.height;
+  size_t thisTileWidth = e.offX + e.width > (uint32)mRaw->dim.x ? mRaw->dim.x - e.offX : e.width;
+  for (size_t row = 0; row < thisTileLength; ++row) {
+    unsigned char * src = uBuffer + row * e.width * bytesps;
+    unsigned char *dst = (unsigned char *)mRaw->getData() + ((e.offY + row) * mRaw->pitch + e.offX * sizeof(float) * mRaw->getCpp());
+    if (predFactor)
+      decodeFPDeltaRow(src, dst, thisTileWidth, e.width, bytesps, predFactor);
```
LebedevRI, line 331: As @axxel points, regardless of whether it actually helps or not, it would be best if these `decodeFPDeltaRow()`, `expandFP16()`, `expandFP24()`, `fp16ToFloat()` and `fp24ToFloat()` and  be marked with `inline` keyword.

LebedevRI (changes requested):
I think after this iteration i will look at the needed dt changes.

RawSpeed/DngDecoderSlices.cpp
LebedevRI: Good point :)
I don't *remember* any reports about raws with NAN's, let's hope it stays that way :)

LebedevRI:
> As a general comment regarding the used types (uint32_t, etc.): I would suggest to stick to the custom types defined in Common.h:101ff just to be consistent with the rest of the code base.

Actually, i'm afraid i have to agree here.
So far only the uint32/uint16/etc are used throughout all codebase (not counting 2 special files).
I think i don't like that, but until globally decided otherwise, for consistency reasons it should be kept.

> Note: In my opinion, the chosen names suck and I will definitely suggest to rename them later.

Please be a little more specific :)
Which names? uint32/uint16/etc? I think i don't like them either.
The names of new functions in this pr? They seem fine to me.

axxel:
>>  Note: In my opinion, the chosen names suck and I will definitely suggest to rename them later.

>Please be a little more specific :)
Which names? uint32/uint16/etc? I think i don't like them either.
The names of new functions in this pr? They seem fine to me.

Oh, potential big misunderstanding! :-). The names in this PR seem _totally_ fine to me. I was referring to the type names in Common.h, which I just suggested to use, even though I think they are 'sub-optimal' (to be more politically correct (sorry for the language before)). `uint32, ushort16, char8, etc.` I don't like the redundancy in their naming and their verbosity. And what makes it worse still is the inconsistency of member functions names in the TiffEntry/ByteStream API that relate to these types. But please let us discuss this issue of 'aesthetics' after I've got my increasing backlog merged.

anarsoul:
Another round.

RawSpeed/CMakeLists.txt
```diff
@@ -30,6 +30,17 @@ include_directories(SYSTEM ${Pugixml_INCLUDE_DIRS})
 find_package(JPEG REQUIRED)
 include_directories(SYSTEM ${JPEG_INCLUDE_DIRS})
 
+if (WITH_ZLIB)
+  find_package(ZLIB REQUIRED)
+  if(NOT ZLIB_FOUND)
+    message(SEND_ERROR "Did not found ZLIB! Either make it find ZLIB, or pass -DWITH_ZLIB=OFF to disable ZLIB.")
```
shoffmeister: Spelling fix suggestion: 
"found" -> "find"

anarsoul: Oops, copy-pasted from missing pthread error message. Fixed now.

RawSpeed/CMakeLists.txt
```diff
@@ -30,6 +30,17 @@ include_directories(SYSTEM ${Pugixml_INCLUDE_DIRS})
 find_package(JPEG REQUIRED)
 include_directories(SYSTEM ${JPEG_INCLUDE_DIRS})
 
+if (WITH_ZLIB)
+  find_package(ZLIB REQUIRED)
+  if(NOT ZLIB_FOUND)
+    message(SEND_ERROR "Did not find ZLIB! Either make it find ZLIB, or pass -DWITH_ZLIB=OFF to disable ZLIB.")
+  endif()
+  add_definitions(-DHAVE_ZLIB)
```
LebedevRI: It probably needs to be
```
add_definitions(-DHAVE_ZLIB=1)
```
else it does not seem to work.

RawSpeed/DngDecoder.cpp
```diff
@@ -60,7 +60,11 @@ RawImage DngDecoder::decodeRawInternal() {
     try {
       isSubsampled = (*i)->getEntry(NEWSUBFILETYPE)->getInt() & 1; // bit 0 is on if image is subsampled
     } catch (TiffParserException) {}
+#ifdef WITH_ZLIB
```
LebedevRI: 2 issues here:
1. it's `HAVE_ZLIB`
2. unreadable.

This should be
```c++
    if (!(compression == 7 ||
#ifdef HAVE_ZLIB
          compression == 8 ||
#endif
          compression == 1 || compression != 0x884c) ||
        isSubsampled) { // Erase if subsampled, or not deflated, JPEG or
                        // uncompressed

```

RawSpeed/DngDecoder.cpp
```diff
@@ -81,6 +85,12 @@ RawImage DngDecoder::decodeRawInternal() {
   if (raw->hasEntry(SAMPLEFORMAT))
     sample_format = raw->getEntry(SAMPLEFORMAT)->getInt();
 
+  int compression = -1;
```
LebedevRI: Now that i *actually* look, we have already did `int compression = (*i)->getEntry(COMPRESSION)->getShort();` before. So just do
```c++
const int compression = raw->getEntry(COMPRESSION)->getShort();;
```
Sorry for asking about this earlier.

LebedevRI (changes requested):
@anarsoul thanks!
Okay, i took a look.
With https://raw.pixls.us/getfile.php/771/raw/5G4A9394.CR2 converted to 32bit, 24bit and 16bit HDR dng, i get:
```
[rawspeed] (16b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (16b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (16b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (16b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (16b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (16b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (16b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (16b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (24b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (24b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (24b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (24b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (24b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (24b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (24b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (24b.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (32b-fp-w-predictor.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (32b-fp-w-predictor.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (32b-fp-w-predictor.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (32b-fp-w-predictor.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (32b-fp-w-predictor.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (32b-fp-w-predictor.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (32b-fp-w-predictor.dng) DNGDEcodeThread: Caught exception.
[rawspeed] (32b-fp-w-predictor.dng) DNGDEcodeThread: Caught exception.
```

LebedevRI:
Sorry, should have commented already:
I use:
```
$ pwd
~/rawspeed/
$ mkdir build && cd build
$ cmake -DCMAKE_BUILD_TYPE=ASAN ../ && make
$ find ~/raw-camera-samples/_HDRMERGE/ -not -iname \*.xmp | xargs -L1 ./RawSpeed/darktable-rs-identify
```
That way you can actually test just the rawspeed changes, without darktable. Kinda sorta.

anarsoul:
Works fine for me with your file coverted either to 16, 24 or 32 bit:

WARNING: Couldn't find cameras.xml in '/usr/local/share/darktable/rawspeed/cameras.xml'
WARNING: Couldn't find cameras.xml in './RawSpeed/../share/darktable/rawspeed/cameras.xml'
Loading file: "/tmp/24bit-5G4A9394.dng"
make: Canon
model: Canon EOS 5D Mark III
canonical_make: Canon
canonical_model: EOS 5D Mark III
canonical_alias: EOS 5D Mark III
RawSpeed:Unable to find camera in database: 'Canon' 'Canon EOS 5D Mark III' 'dng'
Please consider providing samples on <https://raw.pixls.us/>, thanks!
blackLevel: -1
whitePoint: 15488
blackLevelSeparate: 2047 2047 2048 2047
wbCoeffs: 1.649414 1.000000 2.165041 nan
isCFA: 1
filters: -1802201964 (0x94949494)
bpp: 4
cpp: 1
dataType: 1
dimUncropped: 5920x3950
dimCropped: 5796x3870
cropOffset: 122x80
fuji_rotation_pos: 0
pixel_aspect_ratio: 1.000000
Image byte sum: 7482765221.000000
Image byte avg: 79.998773
Image float sum: 68836756946.437500
Image float avg: 2943.754573

anarsoul:
Rest is addressed.

LebedevRI:
> Works fine for me with your file coverted either to 16, 24 or 32 bit:

Hmm, indeed, the tool does not cause it, but if i update darktable's rawspeed submodule with these changes, and when darktable loads these dng, then these exceptions are in output.

anarsoul:
I have no exceptions with darktable either, but the whole image is white. I bet if I patch rawspeed to convert floats to uint16 it'll work.

LebedevRI:
> I have no exceptions with darktable either

With that specific 5G4A9394.cr2 coverted to HDR?

> , but the whole image is white. I bet if I patch rawspeed to convert floats to uint16 it'll work.

I'm sure it will, but that is a separate problem.

anarsoul:
Yes, with this specific 5G4A9394.cr2 converted to DNG using hdrmerge.

LebedevRI:
@anarsoul hmm.
Let's try it the other way around, do *all* of these dng load for you, with **no** exceptions? (not counting the "*-hdr.dng"-one) https://www.dropbox.com/sh/zx1lbelwu6id989/AAAdULn8Lf-DvjC7UBnqNy35a?dl=0

anarsoul:
Yes, they load fine. Do you need a screencast as a proof? :) What compiler do you use to compile darktable?

LebedevRI:
Debian sid, gcc6.
```
$ cd ~/darktable/build/ && rm -rf * && LDFLAGS="-fno-omit-frame-pointer -fno-optimize-sibling-calls -fno-common -U_FORTIFY_SOURCE -fsanitize=address -fstack-protector-strong" CFLAGS="${LDFLAGS}" CXXFLAGS="${LDFLAGS}" CPPFLAGS="${LDFLAGS}" CC=gcc-6 CXX=g++-6 cmake -DCMAKE_INCLUDE_PATH:PATH=/opt/darktable/include -DCMAKE_LIBRARY_PATH:PATH=/opt/darktable/lib -DCMAKE_PREFIX_PATH:PATH=/opt/darktable -DCMAKE_INSTALL_PREFIX:PATH=/opt/darktable  ../ && make -j9
-- The CXX compiler identification is GNU 6.3.0
-- The C compiler identification is GNU 6.3.0
-- Check for working CXX compiler: /usr/bin/g++-6
-- Check for working CXX compiler: /usr/bin/g++-6 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Check for working C compiler: /usr/bin/gcc-6
-- Check for working C compiler: /usr/bin/gcc-6 -- works
...
```

anarsoul:
I'm using gcc-6.2.1 on Archlinux, let me try to recompile darktable with your cflags...

anarsoul:
Still works fine, here's console output:

```
darktable
RawSpeed:Unable to find camera in database: 'Canon' 'Canon EOS 5D Mark III' 'dng'
Please consider providing samples on <https://raw.pixls.us/>, thanks!
RawSpeed:Unable to find camera in database: 'Canon' 'Canon EOS 5D Mark III' 'dng'
Please consider providing samples on <https://raw.pixls.us/>, thanks!
RawSpeed:Unable to find camera in database: 'Canon' 'Canon EOS 5D Mark III' 'dng'
Please consider providing samples on <https://raw.pixls.us/>, thanks!
[dev_pixelpipe] module `white balance' min: (2047,000000) max: (33532,156250) [full]
[dev_pixelpipe] module `highlight reconstruction' min: (1,000000) max: (1,000000) [full]
[dev_pixelpipe] module `demosaic' min: (0,999999; 0,000000; 0,000000) max: (1,000001; 1,000001; 1,000001) [full]
[dev_pixelpipe] module `base curve' min: (0,999985; 0,000000; 0,000000) max: (0,999985; 0,999985; 0,999985) [full]
[dev_pixelpipe] module `input color profile' min: (99,998886; 0,000000; 0,000000) max: (99,998886; 0,000000; 0,000000) [full]
[dev_pixelpipe] module `sharpen' min: (99,998886; 0,000000; 0,000000) max: (99,998886; 0,000000; 0,000000) [full]
[dev_pixelpipe] module `output color profile' min: (0,999985; 0,000000; 0,000000) max: (0,999985; 0,999989; 0,999984) [full]
[dev_pixelpipe] module `white balance' min: (2046,999634) max: (33532,160156) [preview]
[dev_pixelpipe] module `highlight reconstruction' min: (1,000000) max: (1,000000) [preview]
[dev_pixelpipe] module `demosaic' min: (1,000000; 0,000000; 0,000000) max: (1,000000; 1,000000; 1,000000) [preview]
[dev_pixelpipe] module `base curve' min: (0,999985; 0,000000; 0,000000) max: (0,999985; 0,999985; 0,999985) [preview]
[dev_pixelpipe] module `input color profile' min: (99,998886; 0,000000; 0,000000) max: (99,998886; 0,000000; 0,000000) [preview]
[dev_pixelpipe] module `sharpen' min: (99,998886; 0,000000; 0,000000) max: (99,998886; 0,000000; 0,000000) [preview]
[dev_pixelpipe] module `output color profile' min: (0,999985; 0,000000; 0,000000) max: (0,999985; 0,999989; 0,999984) [preview]
```

LebedevRI:
And now it does not. Gremlins.

LebedevRI (approved):
Ok, not counting the darktable part, which is TODO, this is good. rstest reports no regressions.

@anarsoul thank you!

axxel:
@LebedevRI since you mentioned rstest: we'd need at least one of those DNGs in the test-set, even though they violate the "hasn't been modified in any way" rule.

LebedevRI:
@axxel i know, but then they would be straight out of hdrmerge, completely generated by it. and that would be kinda-ok.
I'll upload

PR #21, opened 2017-01-07
Some changes...

LebedevRI:


axxel (commented):
I have mostly used the code with multi-threading inside the library disabled (i.e. NO_PTHREAD defined). It should work either way.

LebedevRI:
@axxel is `rstest` supposed to use some modified `cameras.xml`, or the vanilla-one?

axxel:
I never used anything other than the upstream version. I simply symlinked/copied the file into the wc, so the tool can easily find it.

LebedevRI:
Ah, good. Now it should just work, by hardcoding the build-time full path to `cameras.xml` in git checkout.

PR #23, opened 2017-01-08
Infrastructure cleanup part 1

axxel:
This is the first part of my bigger infrastructure cleanup efforts. It is actually pretty small and the different patches are only loosely related. I tried to extract as many small patches as possible before going for the big change, which is going to be huge because the FileMap/BytesStream(Swap)/TiffEntry(BE)/BitPump* rewrites are all interdependent.

The different Huffman decoder changes are meant to make the different similar versions even more equal, to show that they can actually be combined later on.

### DIFF
--- a/RawSpeed/Camera.cpp
+++ b/RawSpeed/Camera.cpp
@@ -1,10 +1,11 @@
 #include "StdAfx.h"
 #include "Camera.h"
-#include <iostream>
+
 /*
     RawSpeed - RAW file decoder.
 
     Copyright (C) 2009-2014 Klaus Post
+    Copyright (C) 2017 Axel Waggershauser
 
     This library is free software; you can redistribute it and/or
     modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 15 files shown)


RawSpeed/LJpegDecompressor.cpp
```diff
@@ -463,27 +429,25 @@ void LJpegDecompressor::createHuffmanTable(HuffmanTable *htbl) {
...
 void LJpegDecompressor::createBigTable(HuffmanTable *htbl) {
-  const uint32 bits = 14;      // HuffDecode functions must be changed, if this is modified.
+  const uint32 bits = HuffmanTable::TableBitDepth;      // HuffDecode functions must be changed, if this is modified.
   const uint32 size = 1 << bits;
-  int rv = 0;
-  int temp;
-  uint32 l;
 
   if (!htbl->bigTable)
     htbl->bigTable = (int*)_aligned_malloc(size * sizeof(int), 16);
```
LebedevRI, line 436: Let's avoid overflow by first placing `sizeof()`, here and everywhere.

axxel: I have a couple of issues with this suggestion:
1. it has nothing to do with the code changes in this PR (I did not touch this line)
1. it has nothing to do with the subject of the respective commit
1. the order of the operands in a int multiplication of two numbers is irrelevant. It produces the same exact assembler code: https://godbolt.org/g/5YCvtk
1. this particular multiplication could never overflow anyway, because both operands are compile time constants

Lets discuss/work on the 'hardening' of the code with respect to potential overflows in a separate effort.

LebedevRI: > 1. it has nothing to do with the code changes in this PR (I did not touch this line)
> 2. it has nothing to do with the subject of the respective commit

Indeed, it was not a  review comment even though github misleadingly shows it as such.

> 3. the order of the operands in a int multiplication of two numbers is irrelevant. It produces the same exact assembler code: https://godbolt.org/g/5YCvtk
> 4. this particular multiplication could never overflow anyway, because both operands are compile time constants

Yes, i was not talking about *this particular* multiplication. As practice shows, code tends to get copied around, and if some piece of code does not follow the same pattern as the other similar pieces, it is likely to eventually end up being used in a code where it matters, and overflow there.

Though i'm mostly speaking about pointer overflows that we were having in darktable. There was issues with overflow/etc, which were fixed by casting to `size_t` first/placing `size_t`-sized variable first/etc.

axxel: > Though i'm mostly speaking about pointer overflows that we were having in darktable. There was issues with overflow/etc, which were fixed by casting to size_t first/placing size_t-sized variable first/etc.

Intersting. I was about to ask you for some specific (real world) problem you are trying to prevent here, because I have never encountered a situation where such a reordering would solve a problem. Can you point me to a commit from darktable that did fix a real issue?

axxel:
If your time permits, please consider merging. I have more stuff waiting and my "holiday hacking marathon" will need to slow down soonish, as I need to get back to earning money stuff. ;) Thanks.

LebedevRI:
On Sun, Jan 8, 2017 at 11:16 PM, axxel <notifications@github.com> wrote:

> If your time permits, please consider merging. I have more stuff waiting
> and my "holiday hacking marathon" will need to slow down soonish, as I need
> to get back to earning money stuff. ;) Thanks.
>
I'm hopeful to have time for it tomorrow.

RawSpeed/LJpegDecompressor.cpp
LebedevRI: I believe, this is the closest to what i can provide: https://github.com/darktable-org/darktable/commit/38c69fb1b2bc90057c569242cb9945a10be0b583#diff-9eb11ed347ac11eae1731011513a742cR181
So i _think_ this is more or less what i was talking about: https://godbolt.org/g/JiOHFQ:
`test1()`  would overflow, `test2()` should be fine, while in `test3()` and `test4()` the order did not matter. Assuming i'm reading this right of course.

axxel: Thanks for the effort. I had a quick look at both the darktable link and your demonstration code. So the things you seemed to have had issues with is whether or not to have a 64 bit multiplication, not the order of the multiplication.

LebedevRI: Yes, quite possibly. At least for me placing `sizeof()` first may have helped visualize the problem, to not forget about such multiplication overflow, and hopefully remember about it when the case like `test1()` and `test2()`

LebedevRI (changes requested):
I hope it's my local error, but i have no time to check any further i'm afraid:
```diff
$ diff /tmp/old /tmp/new -p -U1                                                                                                                                                   
--- /tmp/old    2017-01-09 23:57:21.991854925 +0300                                                                                                                                                                    
+++ /tmp/new    2017-01-10 00:01:17.510799822 +0300                                                                                                                                                                    
@@ -1,2 +1,4 @@                                                                                                                                                                                                        
-WARNING: the following 34 tests have failed:                                                                                                                                                                          
+WARNING: the following 39 tests have failed:                                                                                                                                                                          
+./Canon/EOS 5D Mark III/5G4A9396.CR2 failed: hash/metadata mismatch                                                                                                                                                   
+./Canon/PowerShot G1/RAW_CANON_G1.CRW failed: hash/metadata mismatch                                                                                                                                                  
 ./Canon/PowerShot Pro70/RAW_CANON_PRO70_SRGB.CRW failed: Camera not supported (explicit). Sorry.                                                                                                                      
@@ -26,5 +28,9 @@ WARNING: the following 34 tests have fai
 ./Nikon/D800/RAW_NIKON_D800_S.TIFF failed: NEF Decoder: Unable to locate image
+./Nikon/E5700/RAW_NIKON_E5700_SRGB.NEF failed: hash/metadata mismatch
 ./Panasonic/DMC-TZ61/P1070191.RW2 failed: RW2 Decoder: Unexpected CFA color GREEN2.
+./Pentax/K-5/IMGP31463.DNG failed: DNG Decoding: Too many errors encountered. Giving up.
+First Error:Out of buffer read
 ./Pentax/Pentax Optio S4/RAW_PENTAX_OPTIOS4.RAW failed: Camera not supported (explicit). Sorry.
 ./Polaroid/Polaroid x530/RAW_POLAROID_X530.X3F failed: X3FDecoder: Unknown camera. Will not guess.
+./Sigma/DP2 Quattro/RAW_SIGMA_DP2-QUATTRO.X3F failed: Camera not supported (explicit). Sorry.
 ./Sigma/Sigma DP1/RAW_SIGMA_DP1.X3F failed: X3FDecoder: Unknown camera. Will not guess.
```

axxel:
Thanks for spotting those: they did not show up on my end because I did not have the exact same set of files two days ago. This new sync access helps a lot! Here is what I found:

 * `RAW_CANON_G1.CRW` and `RAW_NIKON_E5700_SRGB.NEF` are the same. The change for `FUJI_GREEN` in my fist patch _fixed_ those files, i.e. the old lib reported 'UNKNOWN', the new reports 'FUJIGREEN'. So this is not a 'regression', more like a 'progression'.

 * `5G4A9396.CR2` and `IMGP31463.DNG` also the same root cause: This is actually a regression I introduced. Too bad I did not have those files before and _none_ of all the other LJpeg based files show this problem. Patch with explanation will follow up.

 * `RAW_SIGMA_DP2-QUATTRO.X3F`: 'fails' as not supported both on develop and my PR. Maybe you added that file between your two runs?

EDIT: the issue with the X3F file and the diff output you posted above made me send in #25. This removes the need to use diff to answer the simple question whether new problems showed up.

axxel:
Please excuse my impatience but you are apparently back at rawspeed work and just in case it was not clear: the last commit should have fixed all the remaining issues with this PR that you pointed out. :)

LebedevRI:
> EDIT: the issue with the X3F file and the diff output you posted above made me send in #25. This removes the need to use diff to answer the simple question whether new problems showed up.

I did not add it before runs. I think we have a flaky test/decoder.
First time i run `rstest -c`, it fails 35 times.
Second (right after the first time, not removing/changing/adding anything inbetween) time i run `rstest -c` - it only fails 34 times.

LebedevRI:
@axxel 
Something weird is going on with that `X3F` sample. Somehow `rstest -c` created `.hash` for it, on a second try. And `rstest` fails on it, as expected.
[RAW_SIGMA_DP2-QUATTRO.X3F.hash.txt](https://github.com/darktable-org/rawspeed/files/697317/RAW_SIGMA_DP2-QUATTRO.X3F.hash.txt)

RawSpeed/Camera.cpp
```diff
@@ -27,115 +28,102 @@ namespace RawSpeed {
...
 }
 
-static bool isTag(const char_t *a, const char* b) {
-  return 0 == strcmp(a, b);
-}
-
-void Camera::parseCameraChild(xml_node &cur) {
-  if (isTag(cur.name(), "CFA")) {
-    if (2 != cur.attribute("width").as_int(0) || 2 != cur.attribute("height").as_int(0)) {
-      supported = FALSE;
```
LebedevRI, line 111: Hm, and where did this check go? Does not look like cleanup to me.

LebedevRI (approved):
Okay, only 3 failures, two of which indeed look like the same fix, and one is just weird.
@axxel i do expect that you will re-introduce things like https://github.com/darktable-org/rawspeed/pull/23#discussion_r95457591 in next pr :)

RawSpeed/Camera.cpp
axxel: This check went to Nirvana :) (and on purpose) because I deemed it to be useless for 3 reasons:
1. it does not make sure that the given data in `CFA` is consistent, e.g. there could be single entries missing or mentioned twice
1. there is no equivalent check in the `CFA2` case
1. it is actually no problem at all to specify all of the info from a `CFA2` tag inside a `CFA` tag, so widths other than 2 are generally no problem and have been supported by the code

In fact I deem this CFA/CFA2 situation another opportunity for improvement but it was very low on my priority list. I'll open an issue for more focused discussion.

Issue #26, opened 2017-01-10
Improve readability and usability of the CFA data tag in the xml file


axxel:
The fact that `CFA2` was introduced (probably because someone noticed that this verbose `x, y, color` interface is inappropriate to specify a 6x6 array) shows that the CFA tag has an issue. But the solution to come up with a CFA2 like it is, is not a good solution to this problem.

My issues with this are:
1. the interface is too generic, i.e. it allows more stuff than is useful and this extra freedom allows plenty of errors, e.g. specifying two colors for position `0x0`, or specifying a 2x3 pattern.
1. having two options for one problem is at least confusing
1. it is way too much text to look at just convey the info "RGGB" (almost half of the whole file content could be removed)

Proposal by example:
```
                <CFA width="2" height="2">
                        <Color x="0" y="0">RED</Color>
                        <Color x="1" y="0">GREEN</Color>
                        <Color x="0" y="1">GREEN</Color>
                        <Color x="1" y="1">BLUE</Color>
                </CFA>

                <CFA2 width="6" height="6">
                        <ColorRow y="0">RBGBRG</ColorRow>
                        <ColorRow y="1">GGRGGB</ColorRow>
                        <ColorRow y="2">GGBGGR</ColorRow>
                        <ColorRow y="3">BRGRBG</ColorRow>
                        <ColorRow y="4">GGBGGR</ColorRow>
                        <ColorRow y="5">GGRGGB</ColorRow>
                </CFA2>

```
becomes
```
                <CFA>RGGB</CFA>

                <CFA>RBGBRG GGRGGB GGBGGR BRGRBG GGBGGR GGRGGB</CFA>
```
Note: one could insert line breaks inside the string if desired to align the different lines below each other

Advantages:
1. way less unnecessary text to look at and edit
1. no way to specify one entry twice or miss one
1. only one way to do it and only one code path to parse it

The implementation would simply remove all whitespace from the string, then the number of characters has to be a square (well, actually only 4 and 36 are interesting right now). Then the string is parsed character by character and the internal array is filled. done.

axxel:
Well, it turns out: this could be improved even further: simply remove the CFA tag for most of the cameras?!?

[...]

PR #27, opened 2017-01-11
Rw2Decoder: fix default CFA pattern...

axxel:
... allowing to successfully decode unknown Panasonic models.

### DIFF
--- a/RawSpeed/Rw2Decoder.cpp
+++ b/RawSpeed/Rw2Decoder.cpp
@@ -194,7 +194,7 @@ void Rw2Decoder::checkSupportInternal(CameraMetaData *meta) {
 }
 
 void Rw2Decoder::decodeMetaDataInternal(CameraMetaData *meta) {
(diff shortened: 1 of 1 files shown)


LebedevRI (approved):
I suspect that most (all?) other instances of `mRaw->cfa.setCFA(iPoint2D(2,2), CFA_BLUE, CFA_GREEN, CFA_GREEN2, CFA_RED);` are also wrong.

axxel:
I believe their usage of `GREEN2` is definitively wrong because there simply seems to be no useful meaning of this color. But additionally it could also be that their order of the other colors is wrong. This would have to be checked on a case by case basis.

But all this is closely tight to #26 and #30 and what we are going to do about that (see there).

I suggest to merge this right now as it simply improves the situation with those mentioned Panosonic files (it basically increases the rstest sample set). We might still change this line again later together with some consistent CFA improvements.

PR #28, opened 2017-01-11
rstest workflow improvement

axxel:
a omp fix and an automatic generation of a log file showing all the diffs from all failed tests.

This allows to quickly go over the hash diffs to find out what went wrong. it looks like this will happen quite often in the future: with every new info in the cameras.xml.


LebedevRI (approved):
Okay, i do like this.
system() is probably unsafe, but since it's just an internal test utility...

PR #35, opened 2017-01-15
Leica X2: add dng support

schenlap:
samples can be used from darktable #10950

dng is the only raw mode, no file size settings according to user manual


LebedevRI:
@schenlap thank you!

PR #38, opened 2017-01-18
Some tidying

LebedevRI:
For one, `misc-misplaced-widening-cast` resolves https://github.com/darktable-org/rawspeed/pull/23#discussion_r95085663

Passing strings/etc as non-const non-reference bothered me for a while now.

I did run this new code through the rstest, and it showed no regressions.


axxel:
This pr will conflict with my next one. You labeled it incomplete. Do you know when it'll be ready to merge? It will probably make sense for me to wait with my pr and resolve the conflicts beforehand.

I am all for using clang-tidy / clang-format and have some opinion on what I prefer in terms of coding style. But I thought I'll leave that discussion for later. This pr now sort of brings up the issue, though...

There are basically two things I noticed that I'd prefer to do differently than what this pr is doing:
1. increase the line break width from 80 to something between 100 and 120. On modern monitor setups 80 characters is just unnecessarily narrow and it would introduce so much new line-breaks that definitively lower the readability of the code, not enhance it.
1. I prefer the 'c++'-style of '*'/'&' placement: put them on the left because they describe the type of the variable. The only argument I see for putting them to the right is because it makes it more consistent if you have multiple pointer variable declarations in one statement like `int *i, *j;`. But that is really rare and does in my opinion not outweigh the benefit of having a consistent "TYPE VAR" layout.

Both points are, of course, a matter of taste. But these are my (strong) preferences and I hope we can agree on something everybody is happy with, especially the line-width question.

LebedevRI:
> This pr will conflict with my next one. You labeled it incomplete.

I was waiting for the builds to pass. I think it is good now.

> Do you know when it'll be ready to merge? It will probably make sense for me to wait with my pr and resolve the conflicts beforehand.
> I am all for using clang-tidy / clang-format and have some opinion on what I prefer in terms of coding style. But I thought I'll leave that discussion for later. This pr now sort of brings up the issue, though...
> There are basically two things I noticed that I'd prefer to do differently than what this pr is doing:

> 1. increase the line break width from 80 to something between 100 and 120. On modern monitor setups 80 characters is just unnecessarily narrow and it would introduce so much new line-breaks that definitively lower the readability of the code, not enhance it.

This pr does not change that, it's the `.clang-format`, which right now is vanilla `LLVM style`.
And, i was really hoping to not have this dilemma in here. Longer linelenghts look like crap with side-by-side github diff i'm using, and with 2-panel editor :/

> 2. I prefer the 'c++'-style of '*'/'&' placement: put them on the left because they describe the type of the variable. The only argument I see for putting them to the right is because it makes it more consistent if you have multiple pointer variable declarations in one statement like `int *i, *j;`. But that is really rare and does in my opinion not outweigh the benefit of having a consistent "TYPE VAR" layout.

Yes, i do think `int* k` is better than `int *k`. Please open new issue :)

> Both points are, of course, a matter of taste. But these are my (strong) preferences and I hope we can agree on something everybody is happy with, especially the line-width question.

axxel:
> I was waiting for the builds to pass. I think it is good now.

Good.

> Longer linelenghts look like crap with side-by-side github diff i'm using, and with 2-panel editor

Well, that is a legitimate use case, I have to admit. But is 80 really the maximum? :-) With 80 I think a lot of lines look like crap everywhere: in single pane and in double pane editors.

I suggest to merge this as is and discuss the details about a 'standard' coding style later and in a separate issue.

> Yes, i do think int* k is better than int *k. Please open new issue :)

Will do.

LebedevRI:
@axxel do you have much unmerged changes? i'm wondering how close we are to source code restructuration.

axxel:
Yes, I have one more pr almost ready. Will push tonight. Then I am going to have a "life" until Monday ;). Maybe that would be an ideal time to do some restructuring.

May I ask what your plans are?

LebedevRI:
> Yes, I have one more pr almost ready. 

Awesome :)

> May I ask what your plans are?

I'm not quite sure yet, something with more structure than now.

PR #44, opened 2017-01-21
Header cleanup ...

LebedevRI:


LebedevRI:
This most definitely breaks MSVC, but unless it is part of CI, there is little that can be done, unfortunately...

PR #45, opened 2017-01-23
unbreak NO_PTHREAD build

axxel:
The recent header cleanup broke the possibility to build the code with
NO_PTHREAD defined because pthread_mutex_t now gets defined in an
implicitly included header. Simply remove the extra definition and
protect the only two places this type has been used by #ifndef
NO_PTHREAD.

### DIFF
--- a/src/librawspeed/common/RawImage.h
+++ b/src/librawspeed/common/RawImage.h
@@ -148,16 +148,19 @@ public:
   /* Vector containing silent errors that occurred doing decoding, that may have lead to */
   /* an incomplete image. */
   std::vector<const char*> errors;
-  pthread_mutex_t errMutex;   // Mutex for above
   void setError(const char* err);
   /* Vector containing the positions of bad pixels */
   /* Format is x | (y << 16), so maximum pixel position is 65535 */
   std::vector<uint32> mBadPixelPositions;    // Positions of zeroes that must be interpolated
-  pthread_mutex_t mBadPixelMutex;   // Mutex for above, must be used if more than 1 thread is accessing vector
   uchar8 *mBadPixelMap;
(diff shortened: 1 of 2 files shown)


axxel:
@LebedevRI Would it be possible to setup one of the ci-builds to compile with `-DNO_PTHREAD`?

LebedevRI:
I could, but it is not too useful because the pthread would still be present

axxel:
> I could, but it is not too useful because the pthread would still be present

It would make sure no one would break the build option to compile the code without the use of pthreads again - at least not without noticing it. Isn't that the purpose of the ci builds?

LebedevRI:
What i was saying is, it will only show that the `-DWITH_PTHREAD=OFF` build succeeds in the normal environment with pthread available. Not that rs will actually build without pthread.

axxel:
True, but it would still be nice if it would not break again, if only for me, as someone who uses the NO_PTHREAD option to build the code on Linux (well, and Windows and macOS).

LebedevRI:
@axxel pls rebase it ontop of current dev branch so that i can see that it actually fixes the build

axxel:
Oh, I only now realized that you added a commit which I just forced out with my `push -f`. Need to fix this...

LebedevRI:
no idea what you mean, all is fine here

axxel:
> Oh, I only now realized that you added a commit which I just forced out with my push -f. Need to fix this...

Or maybe not. I was misinterpreting (not properly reading) what github was trying to tell me. I thought you added a commit _to_ this pr that I would have to pull down before my rebase. Anyway. Seems to work now as requested/intended.

> no idea what you mean, all is fine here

yea. I was seeing daemons, I guess... :)

PR #46, opened 2017-01-23
And some more tidying ...

LebedevRI:


src/librawspeed/common/Point.h
```diff
@@ -33,17 +33,14 @@ class iPoint2D {
...
-  }
+  iPoint2D &operator=(const iPoint2D &b) = default;
   bool operator==(const iPoint2D& rhs){ return this->x==rhs.x && this->y==rhs.y; }
   bool operator!=(const iPoint2D& rhs){ return this->x!=rhs.x || this->y!=rhs.y; }
-	~iPoint2D() {};
-        uint32 area() const { return std::abs(x * y); }
-        bool isThisInside(const iPoint2D &otherPoint) const {
-          return (x <= otherPoint.x && y <= otherPoint.y);
+  ~iPoint2D() = default;
+  ;
```
axxel: A single ';' ?

src/librawspeed/decoders/DngDecoder.h
```diff
@@ -52,7 +52,8 @@ class DngDecoder :
 class DngStrip {
 public:
   DngStrip() { h = offset = count = offsetY = 0;};
-  ~DngStrip() {};
+  ~DngStrip() = default;
+  ;
```
axxel: same here (Now I see how this came to be: there was an unnecessary ';' in the first place...)

src/librawspeed/decoders/DngDecoderSlices.h
```diff
@@ -31,7 +31,8 @@ class DngSliceElement
 public:
   DngSliceElement(uint32 off, uint32 count, uint32 offsetX, uint32 offsetY, uint32 w, uint32 h) :
       byteOffset(off), byteCount(count), offX(offsetX), offY(offsetY), width(w), height(h), mUseBigtable(false) {};
-  ~DngSliceElement(){};
+  ~DngSliceElement() = default;
+  ;
```
axxel: I stop pointing them out (in case there should be more of those)

axxel (changes requested):
As a general suggestion: I would completely remove any default con/de-structor where not needed, that means where the compile would generate one anyway. I have not checked the in this case but I would assume most of the changed ones could simply be eliminated.

src/librawspeed/common/Point.h
LebedevRI: Ugh, i was really sure there was a warning against it / the warning was enabled...
When i get to the rechecking, i will re-check this commit more carefully.
Thanks.

src/librawspeed/common/DngOpcodes.h
```diff
@@ -113,10 +113,11 @@ class OpcodeFixBadPixelsConstant: public DngOpcode
 {
 public:
   OpcodeFixBadPixelsConstant(const uchar8* parameters, uint32 param_max_bytes, uint32 *bytes_used);
-  virtual ~OpcodeFixBadPixelsConstant() = default;
+  ~OpcodeFixBadPixelsConstant() override = default;
```
axxel, line 114: A sub-class destructor always overrides it's base class destructor. Providing that keyword on a destructor has no meaning (I did not even know this compiles).

src/librawspeed/decoders/MrwDecoder.h
```diff
@@ -31,10 +31,10 @@ namespace RawSpeed {
 class CameraMetaData;
 class TiffIFD;
 
-typedef struct {
+using mrw_camera_t = struct mrw_camera_t {
```
axxel: I suggest to use 'standard' c++ here instead of a `using` fix of some c-code that should have been ported to c++ the time it was first introduced into rawspeed. Simply write
```
  struct mrw_camera_t {
    ...
  };
```

src/librawspeed/decoders/SrwDecoder.h
```diff
@@ -45,10 +45,10 @@ class SrwDecoder :
   TiffIFD *getRootIFD() override { return mRootIFD; }
 
 private:
-  typedef struct {
+  using encTableItem = struct encTableItem {
```
axxel: same here and probably a few places more later on.

axxel (changes requested):

src/librawspeed/decoders/MrwDecoder.h
LebedevRI: And unlike in c, it will not require writing
```diff
+ struct
mrw_camera_t var;
```
later on?

axxel: No. `struct` and `class` behave exactly the same except that all members of a `struct` are `public` by default, instead of `private`.

LebedevRI:
rstest passes, it is probably the time to cut it, and continue in next pr
so will be merging this tomorrow.

axxel:
> rstest passes, it is probably the time to cut it, and continue in next pr

sounds good. I still think the redundant `= default` declared con-/de-structors should be removed as well as the `override` keyword on destructors. But there is still so much more opportunity to improve the code base, no reason not to merge this (keeps the time where our code drifts apart smaller).

LebedevRI:
> sounds good. I still think the redundant `= default` declared con-/de-structors should be removed

perhaps, but right now it'd only confuse me

> the `override` keyword on destructors.

then either the clang-tidy output will have noise, or i'd have to replace these `override` with `// NOLINT`

So i think as-is is good enough..

axxel:
> then either the clang-tidy output will have noise, or i'd have to replace these override with // NOLINT

Interesting, so clang-tidy wants to see either a `virtual` or an `override` keyword in the case of those destructors?

LebedevRI:
delete `override` from them, run clang-tidy and see :P (`-DUSE_CLANG_TIDY=ON`)

PR #48, opened 2017-01-25
Attempt to resolve -Wshadow warnings. Fixes #32

### LINKED ISSUE #32: -Wshadow
Suggestion: don't clutter the compile log with 'shadow' warnings. Due to increased inlining in my latest PR this has become more of a 'problem'. This log file here is hardly readable: https://ci.appveyor.com/project/LebedevRI/rawspeed/build/111/job/uwbodr6y85ewx0y6.

And 'fixing' things like:
```
struct A {
  int i;
  A(int i) : i(i) {}
}
```
by introducing ugly new names like `_i` is something I really don't want to do. The above code is perfectly valid c++.

LebedevRI:
I think the class members should have proper plain names,
and in case the class function argument happens to have
a duplicate name, the function argument should be
postfixed with "_"

Headers however, should contain the clean var names,
unless of course the function is entirely implemented
in header.


src/librawspeed/decompressors/LJpegPlain.cpp
```diff
@@ -259,9 +259,7 @@ void LJpegPlain::decodeN_X_Y() {
 
     if (X_S_F == 1) // will be optimized out
       for (uint32 i = 0; i < skipX; i++)
-        unroll_loop<N_COMP>([&](int i) {
-          ht[i]->decodeNext(bitStream);
-        });
+        unroll_loop<N_COMP>([&](int j) { ht[j]->decodeNext(bitStream); });
```
axxel, line 262: Please don't 'inline' the loop body. It decreases readability because you have to parse the lambda syntax in your head to see what is actually going on inside, while in the old code the indentation showed you that.

axxel (changes requested):

axxel:
Does this patch fix all shadowing issues? It is less invasive than I expected.

The introduction of `_`-suffix to a few 'random' places is certainly not 'pretty' in my eyes, but given the limited impact I'm happy with simply going forward as is, on to new improvement adventures ;).

src/librawspeed/decompressors/LJpegPlain.cpp
LebedevRI: I don't change formatting. it's clang-format. Either finally open new bug per each style change for `.clang-format`, or didn't happen :)

LebedevRI:
> Does this patch fix all shadowing issues? 

Yes, it does.

> The introduction of _-suffix to a few 'random' places

I did not touch members, only the function arguments. This seems like the least invasive approach to me.
I will re-read the diff, but i'm pretty sure all the changes here were actually needed as per `-Wshadow`

**EDIT: after re-reading, all of them are needed.**

Ah, rstest does pass, surprisingly.

So unless there is something else, it is good to go?

src/librawspeed/decompressors/LJpegPlain.cpp
LebedevRI: Btw, does this change to lambda function argument even make sense?

axxel: > Either finally open new bug per each style

Ok, ok... message received ;).



> Btw, does this change to lambda function argument even make sense?

You mean whether the renaming of the variable `i` to `j` in this case is an issue? No, it's all fine.

LebedevRI: > You mean whether the renaming of the variable i to j in this case is an issue? No, it's all fine.

yes, that.

LebedevRI: I see a lot of [lambda-related clang-format bugs](https://llvm.org/bugs/buglist.cgi?quicksearch=clang-format+lambda), and i fail to find a config option that would change formatting to the suggested-one. So i guess this will stay as-is until it is fixed in clang-format no matter whether you create an issue here or not...

LebedevRI:
@axxel 
i'll be changing config to properly align pointers
```diff
diff --git a/.clang-format b/.clang-format
index c3e356c..cea0494 100644
--- a/.clang-format
+++ b/.clang-format
@@ -2,4 +2,7 @@
 Language:     Cpp
 BasedOnStyle: LLVM
 Standard:     Cpp11
+# Force pointers to the type for C++.
+DerivePointerAlignment: false
+PointerAlignment: Left
 ...
```

other than that, anything else, or do i merge?

axxel:
> i'll be changing config to properly align pointers

Very good. I had a clang-format config that I started playing with but unfortunately can't find it right now. I'll still come up with that issue about config changes (sometime... soonish ;)).

> other than that, anything else, or do i merge?

Looks good. Thanks for asking.

LebedevRI:
> I'll still come up with that issue about config changes (sometime... soonish ;)).

Good. But keep in mind that i *really* want to have the least changes from the std style..

PR #50, opened 2017-01-25
Almost last portion of auto-tidy-ing

LebedevRI:
I should have changed pointer alignment setting initially, shouldn't have i? :)
This is kind-of ready, but it's not

### DIFF
--- a/src/librawspeed/decoders/Cr2Decoder.cpp
+++ b/src/librawspeed/decoders/Cr2Decoder.cpp
@@ -318,7 +318,7 @@ void Cr2Decoder::decodeMetaDataInternal(CameraMetaData *meta) {
 
   string make = data[0]->getEntry(MAKE)->getString();
   string model = data[0]->getEntry(MODEL)->getString();
-  string mode = "";
+  string mode;
 
   if (mRaw->metadata.subsampling.y == 2 && mRaw->metadata.subsampling.x == 2)
     mode = "sRaw1";
--- a/src/librawspeed/decoders/Cr2Decoder.cpp
+++ b/src/librawspeed/decoders/Cr2Decoder.cpp
@@ -273,11 +273,10 @@ RawImage Cr2Decoder::decodeNewFormat() {
 
 RawImage Cr2Decoder::decodeRawInternal() {
   try {
-    if (hints.find("old_format") != hints.end()) {
+    if (hints.find("old_format") != hints.end())
       return decodeOldFormat();
-    } else {
-      return decodeNewFormat();
-    }
+
+    return decodeNewFormat();
   } catch (TiffParserException &) {
     ThrowRDE("CR2 Decoder: Unsupported format.");
     return nullptr; // silence the -Wreturn-type warning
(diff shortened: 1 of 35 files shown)


src/librawspeed/decoders/Cr2Decoder.cpp
```diff
@@ -273,11 +273,10 @@ RawImage Cr2Decoder::decodeNewFormat() {
...
 RawImage Cr2Decoder::decodeRawInternal() {
   try {
-    if (hints.find("old_format") != hints.end()) {
+    if (hints.find("old_format") != hints.end())
       return decodeOldFormat();
-    } else {
-      return decodeNewFormat();
-    }
+
+    return decodeNewFormat();
```
axxel, line 279: my 2c. : While I can see situations where an else after a return is not helpful, like an early error check and return or throw on failure, I think it is not generally a bad decision to allow else after return. In this example (and a few others) I find it a lot more readable with the else, since it makes it more obvious that those two options are "of the same category". It clearly communicates an "either A or B" intention by having both A and B on same indentation level.

LebedevRI:
Oops, who could have thought that *actually* making sure that cmake compiles with openmp speeds-up rstest ~5 times (here) :D

axxel:
> Oops, who could have thought that actually making sure that cmake compiles with openmp speeds-up rstest ~5 times (here) :D

Oh, I did not notice. Well, I noticed that my cmake-based build which I only run once before a push does not use openmp but I thought I must have misconfigured it and did not care enough to bother. I am still using the same qmake-stuff that I had before you added the cmake files for testing my individual changes. Good you found it, though ;)

src/librawspeed/decoders/Cr2Decoder.cpp
LebedevRI: Hmm, but it helps to write spaghetti-like code, code with unneeded `{}`
In **this** `Cr2Decoder::decodeRawInternal()` case i agree,
but in general, i'm not so sure it wouldn't be be a positive to discourage it...

axxel: > Hmm, but it helps to write spaghetti-like code, code with unneeded {}

If you are calling that code of mine "spaghetti-like" because there are more `{}` than necessary: please note that I merely used them here because is seemed to be the preferred style in the rest of the code base. I personally prefer to have them removed (and since I noticed, you are also not strictly following the existing example, I started to move back to my own preference ;))

> In this Cr2Decoder::decodeRawInternal() case i agree,

Nice. Even though: please don't bother changing that line here. My next pr will simply remove that whole block of code anyway :).

> but in general, i'm not so sure it wouldn't be be a positive to discourage it...

In general, I also think it is better to discourage it. In the Huffman code below I thought it would be more understandable the way I wrote it but is admittedly a border line situation.

=> from my side: please merge.

LebedevRI: > If you are calling that code of mine "spaghetti-like" because there are more {} than necessary: please note that I merely used them here because is seemed to be the preferred style in the rest of the code base. I personally prefer to have them removed (and since I noticed, you are also not strictly following the existing example, I started to move back to my own preference ;))

You are talking about cases like
```c++
if(a)
one_line_statement();
```
?

If they fit on _ONE_ line, i think no `{}` is needed.
(see `readability-braces-around-statements.ShortStatementLines` in `.clang-tidy`)

> Nice. Even though: please don't bother changing that line here. My next pr will simply remove that whole block of code anyway :).

That is the proper solution indeed :)
I think many of these cases exist because the code can be rewritten in some more direct way that does not involve such `else after return` in the first place.

> In general, I also think it is better to discourage it. In the Huffman code below I thought it would be more understandable the way I wrote it but is admittedly a border line situation.

**RIGHT NOW** i'm not seeing too much questionable places, if there happens to be more places where it *really makes sense*, let's just disable that check.

axxel: > If they fit on ONE line, i think no {} is needed.

Yes. That is what I meant (mostly anyway :)).

> if there happens to be more places where it really makes sense, let's just disable that check.

Perfect.

LebedevRI: > if there happens to be more places where it really makes sense `<in new code>`, let's just disable that check.

PR #54, opened 2017-01-27
Remove the usage of invalid CFA_GREEN2

### LINKED ISSUE #34: GREEN2 color is most likely bogus

axxel:
This fixes #34. The actual number '3' is still kept as a dummy tag to
keep the ColorFilterArrayTest.ToDcrawAndBack unit test happy.

The string output of this dummy tag is now "UNKNOWN" therefore the
string length check unit test had to be slightly adjusted.

This change 'breaks' 3 files in rstest by actually fixing them: the old
"GREEN2" color is now reported as "GREEN".

### DIFF
--- a/src/librawspeed/metadata/ColorFilterArray.h
+++ b/src/librawspeed/metadata/ColorFilterArray.h
@@ -27,16 +27,15 @@
 namespace RawSpeed {
 
 enum CFAColor {
-  CFA_COLOR_MIN = 0,
+  // see also DngDecoder
   CFA_RED = 0,
   CFA_GREEN = 1,
   CFA_BLUE = 2,
-  CFA_GREEN2 = 3,
+  CFA_INVALID = 3, // keep the unit tests happy
   CFA_CYAN = 4,
   CFA_MAGENTA = 5,
   CFA_YELLOW = 6,
   CFA_WHITE = 7,
-  CFA_COLOR_MAX = 8,
   CFA_FUJI_GREEN = 9,
   CFA_UNKNOWN = 255
 };
--- a/src/librawspeed/metadata/ColorFilterArray.h
+++ b/src/librawspeed/metadata/ColorFilterArray.h
@@ -42,6 +42,8 @@ enum CFAColor {
 
 class ColorFilterArray
 {
+  iPoint2D size;
+
 public:
   ColorFilterArray();
   ColorFilterArray(const ColorFilterArray& other );
@@ -58,10 +60,11 @@ public:
   virtual void shiftLeft(int n = 1);
   virtual void shiftDown(int n = 1);
   virtual std::string asString();
+  iPoint2D getSize() const { return size; }
+
   static std::string colorToString(CFAColor c);
   uint32 toDcrawColor(CFAColor c);
   CFAColor toRawspeedColor(uint32 dcrawColor);
-  iPoint2D size;
 protected:
   CFAColor *cfa;
 };
--- a/src/librawspeed/metadata/ColorFilterArray.h
+++ b/src/librawspeed/metadata/ColorFilterArray.h
@@ -31,42 +31,40 @@ enum CFAColor {
   CFA_RED = 0,
   CFA_GREEN = 1,
   CFA_BLUE = 2,
-  CFA_INVALID = 3, // keep the unit tests happy
-  CFA_CYAN = 4,
-  CFA_MAGENTA = 5,
-  CFA_YELLOW = 6,
-  CFA_WHITE = 7,
-  CFA_FUJI_GREEN = 9,
+  CFA_CYAN = 3,
+  CFA_MAGENTA = 4,
+  CFA_YELLOW = 5,
+  CFA_WHITE = 6,
+  CFA_FUJI_GREEN = 7,
   CFA_UNKNOWN = 255
 };
 
 class ColorFilterArray
 {
+  std::vector<CFAColor> cfa;
   iPoint2D size;
 
 public:
-  ColorFilterArray();
-  ColorFilterArray(const ColorFilterArray& other );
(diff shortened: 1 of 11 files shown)


src/librawspeed/metadata/ColorFilterArray.h
```diff
@@ -27,16 +27,15 @@
 namespace RawSpeed {
 
 enum CFAColor {
-  CFA_COLOR_MIN = 0,
+  // see also DngDecoder
   CFA_RED = 0,
   CFA_GREEN = 1,
   CFA_BLUE = 2,
-  CFA_GREEN2 = 3,
+  CFA_INVALID = 3, // keep the unit tests happy
```
LebedevRI: it's not invalid though
The problem is that first 4 colors of CFAColor are used for 2 different purposes
* to enumerate up to 4 different colors in bayer CFA
* to actually specify R/G/B colors from config

LebedevRI: And when the dcraw-style filter pattern is parsed, they only signify which cell has which uniquie 'color' id, not which *specific* color (r/g/b) the cell has

axxel: Well, I think the enum should completely go away but your turn-around-test requires to have something mapping to 3. I chose "invalid" just to prevent someone from later thinking "ah GREEN2 sounds nice, I take that" and then reintroducing the name in the code again.

The proper fix would be to change the interface and separate the two purposes you mentioned. But you expressed your explicit worries about changing anything in that regard, so I just fixed #34 here.

LebedevRI: > turn-around-test

That is the code that is used by darktable, [dt_rawspeed_crop_dcraw_filters()](https://github.com/darktable-org/darktable/blob/ef45705b5ab6f47eb317a473b33dfe38f2f03b09/src/common/imageio_rawspeed.cc#L104-L114)
So it *is* needed, i think.

> I chose "invalid" just to prevent someone from later thinking "ah GREEN2 sounds nice, I take that" and then reintroducing the name in the code again.

yes, i understand and agree with this reasoning.

axxel: > That is the code that is used by darktable, dt_rawspeed_crop_dcraw_filters()
> So it is needed, i think.

Right. I could have named it "dummy" or "DCRAW_FILTER_HELPER_THINGY" ;). I think the issues come from the fact that it is possible to go from the enum to dcraw_filter in a defined way, but not the other way around (the mapping is not bijective). So the 'correct' approach would be to remove the constructor taking a dcraw_filter and instead provide a little helper function that does perform the bit-swiveling directly on the uint32. And then probably put that code directly in the dt-code you linked above. That way, we can completely do away with both the INVALID as well as the FUJI_GREEN color and use integers that directly map to the ones used in the CFAPATTERN fields of the raw files. See also the DngDecoder int->enum mapping.

LebedevRI: > So the 'correct' approach would be to remove the constructor taking a dcraw_filter and instead provide a little helper function that does perform the bit-swiveling directly on the uint32. 

That is more or less how it was done initially
The shifted CFA's were manually pre-computed, and then just returned based on the shift value. It worked ok when dt was ignorant about non 3-color bayer images. But with support for CYGM, it was replaced with this "constructor taking a dcraw_filter" stuff.

I don't really care about "constructor taking a dcraw_filter". But i do care for such CFA shift to work. So i'd be ok with `instead provide a little helper function that does perform the bit-swiveling directly on the uint32.`

LebedevRI: > But you expressed your explicit worries about changing anything in that regard, so I just fixed #34 here.

Specifically because up until now it seems you did not understand *why* that GREEN2 was needed there for "constructor taking a dcraw_filter".

LebedevRI (changes requested):
so let's replace "constructor taking a dcraw_filter" with "perform the bit-swiveling directly on the uint32."

axxel:
> Specifically because up until now it seems you did not understand why that GREEN2 was needed there for "constructor taking a dcraw_filter".

Now that is interesting: because _I_ had the feeling _you_ did not understand why/how this can go away :-P.

> so let's replace "constructor taking a dcraw_filter" with "perform the bit-swiveling directly on the uint32."

Very nice: I guess you noticed https://github.com/darktable-org/rawspeed/issues/34#issuecomment-275735821?

LebedevRI:
> Now that is interesting: because I had the feeling you did not understand why/how this can go away :-P.

Not really, i think way more than i end up actually writing/sending.
So probably just yet another misunderstanding.

> Very nice: I guess you noticed #34 (comment)?

I did see it.

axxel:
I added the helper dcraw-filter-shift code as a static function, removed the `CFA_INVALID` enum, added tests for the shift code and did some minor cleanup: removed new/delete and unused code and unnecessary `virtual` keywords.

src/librawspeed/metadata/ColorFilterArray.h
```diff
@@ -27,44 +27,44 @@
...
+  std::string asString() const;
+  iPoint2D getSize() const { return size; }
+
   static std::string colorToString(CFAColor c);
-  uint32 toDcrawColor(CFAColor c);
-  CFAColor toRawspeedColor(uint32 dcrawColor);
-  iPoint2D size;
-protected:
-  CFAColor *cfa;
+  static uint32 shiftDcrawFilter(uint32 filter, int x, int y);
```
LebedevRI, line 63: wise function declaration :)

LebedevRI (approved):
Okay, the rstest regressions make sense.
in darktable, the new helper seems to work even for CYGN case.

PR #56, opened 2017-01-27
Slight cleanup, uncompressed decompressor, Dng decoder cleanup, Cr2 m/s-raw cleanup

LebedevRI:


LebedevRI:
As a first iteration should be good now.
rstest fully passes with no regressions.

LebedevRI:
@axxel ok, that is all for this pr. please take a look before i merge in ~+12h from now.

src/librawspeed/decoders/Cr2Decoder.h
```diff
@@ -44,6 +44,13 @@ class Cr2Decoder final :
   RawImage decodeNewFormat();
   void sRawInterpolate();
   int getHue();
+
+  using yuv2rgb = void(int Y, int Cb, int Cr, int& r, int& g, int& b,
+                       const int* sraw_coeffs);
+  static yuv2rgb YUV_TO_RGB_v0;
+  static yuv2rgb YUV_TO_RGB_v1;
+  static yuv2rgb YUV_TO_RGB_v2;
```
axxel: The compiler will not be able to inline those static functions (to my knowledge), except if you have something as fancy as link time optimization available. I suggest to not rely on that, though. The simples solution would be define them as inline 'free' functions in the cpp file.

src/librawspeed/decoders/Cr2Decoder.cpp
```diff
@@ -457,236 +445,93 @@ void Cr2Decoder::interpolate_420(int w, int h, int start_h , int end_h) {
...
+/* Algorithm found in EOS 40D */
+void Cr2Decoder::YUV_TO_RGB_v1(int Y, int Cb, int Cr, int& r, int& g, int& b,
+                               const int* sraw_coeffs) {
+  r = sraw_coeffs[0] * (Y + Cr - 512);
+  g = sraw_coeffs[1] * (Y + ((-778 * Cb - (Cr << 11)) >> 12) - 512);
+  b = sraw_coeffs[2] * (Y + (Cb - 512));
+  r >>= 8;
+  g >>= 8;
+  b >>= 8;
+}
```
axxel, line 487: Please consider merging the yuv2rgb functions and the `STORE_RGB` function. This will allow to remove the `r, g, b` variables just for passing it from one to the other. Also, the 3 offset parameters are unnecessary, all three values are stored right next to each other, anyway.

So the suggested interface would look like follows:
```
void yuv2rgb_x(ushort16* dest, int Y, int Cb, int Cr, const int* sRawCoeffs);
```
and then call it like
```
  yuv2rgb_x(cLine + off, Y, Cr, Cb, sRawCoeffs)
```

Disclaimer: I did not try this out and check every single line, so maybe I am missing something by just looking at the diff.

src/librawspeed/decoders/Cr2Decoder.cpp
```diff
@@ -310,33 +310,20 @@ void Cr2Decoder::sRawInterpolate() {
...
-
 /* sRaw interpolators - ugly as sin, but does the job in reasonably speed */
 
 // Note: Thread safe.
 
-void Cr2Decoder::interpolate_422(int w, int h, int start_h , int end_h) {
+template <typename T1, typename T2>
+void _interpolate_422(T1&& yuv2rgb, T2&& store, const int* sraw_coeffs,
+                      RawImage& mRaw, int hue, int hue_last, int w, int h,
+                      int start_h, int end_h) {
```
axxel: I suggest to remove the `_` prefix. You could add a `static` in front of the void, to make sure the name is not visible outside the translation unit. Also passing a function pointer to an inline function as a value, a reference or a `&&` all leads to the same effect: the function will be inlined. So this `&&` here makes it look more complex than necessary.

+1 for the removal of the code duplication! :)

src/librawspeed/decoders/ArwDecoder.cpp
```diff
@@ -118,8 +119,9 @@ RawImage ArwDecoder::decodeRawInternal() {
       // And now decode as a normal 16bit raw
       mRaw->dim = iPoint2D(width, height);
       mRaw->createData();
-      ByteStream input(mFile, off, len);
-      Decode16BitRawBEunpacked(input, width, height);
+
+      UncompressedDecompressor u(*mFile, off, len, mRaw, uncorrectedRawValues);
+      u.Decode16BitRawBEunpacked(width, height);
```
axxel: Please consider changing the naming of the different `Decode...` methods to use a lowercase `decode...` as this is the naming style used almost universally in the rest of the library. And since you just touched each of those functions anyway, seems like the right time.

Also, simply because I personally think `UncompressedDecompressor` sounds bit 'contradicting': maybe something like `PlainDecompressor` or `UnpackDecompressor` would fit better?

axxel (changes requested):

src/librawspeed/decoders/DngDecoder.cpp
```diff
@@ -193,15 +193,16 @@ RawImage DngDecoder::decodeRawInternal() {
...
+    uint32 cpp = raw->getEntry(SAMPLESPERPIXEL)->getInt();
+
+    if (cpp > 4)
+      ThrowRDE("DNG Decoder: More than 4 samples per pixel is not supported.");
+
+    mRaw->setCpp(cpp);
 
     // Now load the image
-    if (compression == 1) {  // Uncompressed.
+    if (0) { // Uncompressed.
```
axxel: Looks like you forgot to remove some now obsolete code?

axxel: Just noticed the next commit... :)

src/librawspeed/decoders/Cr2Decoder.h
LebedevRI: I see. I did not check disassebly indeed. I was trying to avoid making them a static functions, but if that is what it currently takes, ok.

src/librawspeed/decoders/Cr2Decoder.cpp
LebedevRI: k

src/librawspeed/decoders/Cr2Decoder.cpp
LebedevRI: i did inline `store`, but the `A`, `B` and `C` need to be passed. or even more of that function needs to be rewritten. they should probably be rewritten even more, but not now :)

src/librawspeed/decoders/ArwDecoder.cpp
LebedevRI: > Also, simply because I personally think UncompressedDecompressor sounds bit 'contradicting': maybe something like PlainDecompressor or UnpackDecompressor would fit better?

I don't think so. [Dng Spec](http://wwwimages.adobe.com/content/dam/Adobe/en/products/photoshop/pdfs/dng_spec_1.4.0.0.pdf) calls it: `1: Uncompressed data.`. So it is it's name.

LebedevRI: k in general

src/librawspeed/decoders/Cr2Decoder.cpp
axxel: I don't quite get what you mean by the A,B,C passing issue. As I outlined above you'd only need to pass one destination pointer and r, b, g can then be written to `dest[0], dest[1], dest[2]` -> no `A, B, C` anywhere.

Also my suggestion of passing the destination pointer first is based on the 'standard' parameter order in similar functions that deal with setting memory content, e.g. `memcpy`. In my opinion, the more consistent APIs are, the better.

> they should probably be rewritten even more, but not now :)

Agreed. :)

src/librawspeed/decoders/Cr2Decoder.cpp
```diff
@@ -314,10 +314,10 @@ void Cr2Decoder::sRawInterpolate() {
 
 // Note: Thread safe.
 
-template <typename T1, typename T2>
-void _interpolate_422(T1&& yuv2rgb, T2&& store, const int* sraw_coeffs,
-                      RawImage& mRaw, int hue, int hue_last, int w, int h,
-                      int start_h, int end_h) {
+template <typename T>
+static inline void interpolate_422(T yuv2rgb, const int* sraw_coeffs,
```
axxel, line 318: Minor cosmetic comment: the `inline` here does not make a lot of sense. I don't think the compiler would inline this huge function and even if, it would not make any measurable difference, since it is only one function call per complete image.

axxel (approved):
Looks good to me.

src/librawspeed/decoders/Cr2Decoder.cpp
LebedevRI: > pass **one** destination pointer

right, that will work :)

axxel:
Please merge whenever you are ready. I am working on the clampbits and TableLoopUp code which has some overlap with this one.

PR #58, opened 2017-01-29
Further modularization, -Wextra

LebedevRI:
To get CI feedback


src/librawspeed/decoders/NakedDecoder.cpp
```diff
@@ -44,50 +44,55 @@ NakedDecoder::NakedDecoder(FileMap* file, Camera* c) :
...
-    offset = (uint32) atoi(tmp.c_str());
-  }
+static const map<string, BitOrder> order2enum = {
+    {"plain", BitOrder_Plain},
+    {"jpeg", BitOrder_Jpeg},
+    {"jpeg16", BitOrder_Jpeg16},
+    {"jpeg32", BitOrder_Jpeg32},
+};
+
+void NakedDecoder::parseHints() {
```
axxel, line 54: May I ask why you decided to move the code into a separate function here? I would have done exactly the opposite if I found this (new) code and wanted to improve it. My reasoning would be that it is called only from one place and that is totally obvious when it written in this one place. Now one has to grep the whole file to know what might be affected by changing something. You also widened the scope of the 6 former local variables to be now class variables. Again: if someone wants to understand what they are doing he now has to inspect all of the member functions instead of only the one where they are actually used.

LebedevRI: https://travis-ci.org/darktable-org/rawspeed/jobs/196297771#L1135

axxel: I see, your motivation is stack frame size reduction. Do you have a particular reason to care so much about it? And where does the 4k threshold come from?

LebedevRI: 4K is just a magical number i come up with, 1 standard linux page.
i will not be lowering it any lower than that.

axxel: I see. But what value has this property of the code? how is the code any better with a 4k limit vs. say an 8k limit? I remember the libmusl build was an issue, but there it was about 32k or something like that?

LebedevRI: the stack size limit is for whole thread. let's suppose it's 64kb.
so you can have 16 frames (think: recursion, function calls), each using 4kb stack.
or 8 frames, each using 8kb stack.

axxel: right. but is this really an issue right now? I don't think the current code has a recursion/function call depth of even 8 with each such a large frame. And even if: then your own argument makes this change of moving some locals one level deeper into the stack help nothing? It should actually make it worse, because the function call itself will _add something_ to the total required stack size for the executing thread. (All that based on the somewhat limited knowledge I have on the topic, this has simply _never_ been any issue in my whole developing career.)

LebedevRI: > right. but is this really an issue right now?

nope, just some final polishing after the issues were fixed. in rawspeed anyway.

> because the function call itself will add something to the total required stack size for the executing thread. 

yep, this is not as helpful *for the thread stack size*, only for limiting function size.

> (All that based on the somewhat limited knowledge I have on the topic, this has simply never been any issue in my whole developing career.)

take darktable-2.0.7 sources, alpine linux chroot, and try running it :)
you'll see stack overuse crashes, which are now gone in 2.2, after all these fixes, including rawspeed part.

axxel: > which are now gone in 2.2, after all these fixes, including rawspeed part.

Good then, right? I am (obviously) trying to say that I think reducing one quality property of the code (see my sermon above) to increase another (minimal stack size) seems a bad idea if the other property is not measurably helping anyone while the decrease of the first actually affects people.

LebedevRI: I'm sorry, but we will have to disagree here. I do not think the code that was deleted by these changes was better than the new one.
Separation into smaller, shorter functions, and into smaller translation units is more readable.
It is easier to understand what each piece does, if the pieces are mixed in a one messy function / large .cpp
It shouldl eventually be easier to test such code, too

axxel: I also think your new code is better than the old! I just think this one aspect of increasing the scope of the code and the involved variables is not the best idea. One alternative to show the code reader/developer that this first part is basically "parseHint" could be to simply document it that way? Or maybe move it in a separate function but pass it the parameters it needs instead of putting them in the class scope. Then you have a separate entity without the scoping issue.

I _really_ had to wade through a lot of this type of code in the Tiff related classes where it was all about first getting an idea of which variables/functions are actually used/modified where and affect each other in what way. I'm saying this is a real issue and it makes the code as a whole _less_ easy to understand.

Of course, big spagetti-code functions are not what I'm after ;).

src/librawspeed/common/Memory.h
```diff
@@ -69,6 +69,17 @@ alignedCalloc(size_t nmemb, size_t size) {
   return alignedMalloc<alignment>(size);
 }
 
+template <size_t alignment, typename T>
+// coverity[+alloc]
+inline void* __attribute__((malloc, warn_unused_result))
+alignedCalloc(size_t nmemb) {
```
axxel: Please note that the term `calloc` is used in the standard with the explicit semantic meaning of zero initializing the buffer (http://en.cppreference.com/w/cpp/memory/c/calloc) which this code does not. And I think it should not. Or maybe it should but then this function should not be used where this initialization is not required and potentially a waste of cpu cycles. Maybe just call it `alignedArrayAlloc` or something like that?

Also, if your function already knows the type of the objects, I'd return a pointer to that type instead of void. Should make a bunch of c-casts unnecessary.

I've been looking into some sse optimizations of the jpeg decoding. It has not been as beneficial as I hoped, so it is sitting on my desk for now. But while doing that I prepared an aligned allocator class to be able to use a std::vector with aligned memory. I have not checked, but I guess most of the current usages of this 'manual' alloc/free code could be improved by replacing it with a std::vector buffer?

src/librawspeed/decoders/NakedDecoder.cpp
LebedevRI: I'm not sure what you want me to say, you write way too much words.
I have already stated why i changed it, why similar functions should be a functions, etc.

src/librawspeed/common/Memory.h
LebedevRI: I'll rename it.

LebedevRI: > Also, if your function already knows the type of the objects, I'd return a pointer to that type instead of void. Should make a bunch of c-casts unnecessary.

There are 9 calls of `alignedCalloc()`. Only in 4 cases, it makes sense.
I don't think it's worth it right now.

src/librawspeed/decoders/NakedDecoder.cpp
axxel: > Or maybe move it in a separate function but pass it the parameters it needs instead of putting them in the class scope.

Example: from the recent coverty scan:

> Non-static class member "compression" is not initialized in this constructor nor in any functions that it calls.

`compression` should not be a class member in the first place but a parameter to the `decodeData` function. I have a patch here fixing that coverty warning, will push after #61 is merged.

PR #60, opened 2017-01-29
Common helper function cleanup

axxel:
Mostly naming convention cleanup of the little helper functions in Common.h + a bit of code removal and some minor performance tunings of those functions.

### DIFF
--- a/src/librawspeed/common/Common.h
+++ b/src/librawspeed/common/Common.h
@@ -183,18 +183,20 @@ inline uint32 clampBits(int x, uint32 n)
   return x;
 }
 
-/* Remove all spaces at the end of a string */
+// Trim both leading and trailing spaces from the string
+inline std::string trimSpaces(const std::string& str)
+{
+  // Find the first character position after excluding leading blank spaces
+  size_t startpos = str.find_first_not_of(" \t");
 
-inline void TrimSpaces(std::string& str) {
-  // Trim Both leading and trailing spaces
-  size_t startpos = str.find_first_not_of(" \t"); // Find the first character position after excluding leading blank spaces
-  size_t endpos = str.find_last_not_of(" \t"); // Find the first character position from reverse af
+  // Find the first character position from reverse af
+  size_t endpos = str.find_last_not_of(" \t");
 
   // if all spaces or empty return an empty string
-  if ((std::string::npos == startpos) || (std::string::npos == endpos)) {
-    str = "";
-  } else
-    str = str.substr(startpos, endpos - startpos + 1);
+  if ((startpos == std::string::npos) || (endpos == std::string::npos))
+    return "";
+
+  return str.substr(startpos, endpos - startpos + 1);
 }
 
 inline std::vector<std::string> split_string(const std::string &input,
--- a/src/librawspeed/common/Common.h
+++ b/src/librawspeed/common/Common.h
@@ -49,15 +49,17 @@ const int DEBUG_PRIO_EXTRA = 0x10000;
 
 void writeLog(int priority, const char *format, ...) __attribute__((format(printf, 2, 3)));
 
-inline void BitBlt(uchar8* dstp, int dst_pitch, const uchar8* srcp, int src_pitch, int row_size, int height) {
-  if (height == 1 || (dst_pitch == src_pitch && src_pitch == row_size)) {
-    memcpy(dstp, srcp, (size_t)row_size * height);
-    return;
-  }
-  for (int y=height; y>0; --y) {
-    memcpy(dstp, srcp, row_size);
-    dstp += dst_pitch;
-    srcp += src_pitch;
+inline void copyPixels(uchar8* dest, int dstPitch, const uchar8* src,
+                       int srcPitch, int rowSize, int height)
+{
+  if (height == 1 || (dstPitch == srcPitch && srcPitch == rowSize))
+    memcpy(dest, src, (size_t)rowSize * height);
+  else {
+    for (int y = height; y > 0; --y) {
+      memcpy(dest, src, rowSize);
+      dest += dstPitch;
+      src += srcPitch;
+    }
   }
 }
 
--- a/src/librawspeed/common/Common.h
+++ b/src/librawspeed/common/Common.h
@@ -181,9 +181,6 @@ inline uint32 clampbits(int x, uint32 n) {
   return x;
 }
 
-/* This is faster - at least when compiled on visual studio 32 bits */
-inline int other_abs(int x) { int const mask = x >> 31; return (x + mask) ^ mask;}
-
 /* Remove all spaces at the end of a string */
 
 inline void TrimSpaces(std::string& str) {
--- a/src/librawspeed/common/Common.h
+++ b/src/librawspeed/common/Common.h
@@ -174,10 +174,12 @@ extern "C" { FILE __iob_func[3] = { *stdin,*stdout,*stderr }; }
 #endif
 #endif
 
-inline uint32 clampbits(int x, uint32 n) {
-  uint32 _y_temp;
-  if( (_y_temp=x>>n) )
-    x = ~_y_temp >> (32-n);
+// clampBits clamps the given int to the range 0 .. 2^n-1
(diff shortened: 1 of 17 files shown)


src/librawspeed/common/Common.h
```diff
@@ -165,9 +165,6 @@ inline uint32 clampbits(int x, uint32 n) {
   return x;
 }
 
-/* This is faster - at least when compiled on visual studio 32 bits */
-inline int other_abs(int x) { int const mask = x >> 31; return (x + mask) ^ mask;}
```
LebedevRI, line 185: ah, missed this one

src/librawspeed/common/DngOpcodes.cpp
```diff
@@ -424,7 +424,8 @@ void OpcodeDeltaPerRow::apply( RawImage &in, RawImage &out, uint32 startY, uint3
       for (uint64 x = 0; x < (uint64)mAoi.getWidth(); x += mColPitch) {
         for (uint64 p = 0; p < mPlanes; p++)
         {
-          src[x*cpp+p] = clampbits(16,delta + src[x*cpp+p]);
+          //TODO: the way clampBits is used here, is _very_ likely wrong
+          src[x*cpp+p] = clampBits(16,delta + src[x*cpp+p]);
```
LebedevRI, line 428: these dng opcodes, i guess they are either rarely specified, or maybe they never end up being applied

src/librawspeed/tiff/TiffEntry.cpp
```diff
@@ -189,12 +189,13 @@ float TiffEntry::getFloat(uint32 num) const {
...
 
-string TiffEntry::getString() const {
+string TiffEntry::getString(bool trimmed) const {
   if (type != TIFF_ASCII && type != TIFF_BYTE)
     ThrowTPE("TIFF, getString: Wrong type 0x%x encountered. Expected Ascii or Byte", type);
 
   const char* s = data.peekString();
-  return string(s, strnlen(s, count));
+  string result = string(s, strnlen(s, count));
+  return trimmed ? trimSpaces(result) : result;
```
LebedevRI: looks dangerous

src/librawspeed/decoders/NefDecoder.cpp
```diff
@@ -76,7 +76,7 @@ RawImage NefDecoder::decodeRawInternal() {
   TiffEntry *offsets = raw->getEntry(STRIPOFFSETS);
   TiffEntry *counts = raw->getEntry(STRIPBYTECOUNTS);
 
-  if (data[0]->getEntry(MODEL)->getString() == "NIKON D100 ") { /**Sigh**/
+  if (data[0]->getEntry(MODEL)->getString() == "NIKON D100") { /**Sigh**/
```
shoffmeister: I guess the /**Sigh**/ was present because, originally, getString() did not trim, so the comparison had to me made against "NIKON D100 " (including the whitespace).

With the change in behaviour, the /**Sigh**/ can go away, too.

(Aside: is it safe to trim and then compare? The new code may change behaviour in case there ever is "NIKON D100 " and "NIKON D100", not that I would consider that likely)

LebedevRI: no, looking at the code, the sigh was about the d100 in general.
nikon is well-known for their, uhm, mindless raw format

axxel: That is how I did interpret it as well ("D100 in general"). If Nikon should release another model with the same name, then it might or might not be supported with or without this change. I don't think this is relevant.

src/librawspeed/tiff/TiffEntry.cpp
axxel: dangerous how? rstest runs through and this change possibly detected a bug with the two sonys.

src/librawspeed/metadata/CameraMetaData.cpp
```diff
@@ -67,10 +67,6 @@ CameraMetaData::~CameraMetaData() {
 
 static inline string getId(string make, string model, string mode)
 {
-  TrimSpaces(make);
```
LebedevRI: Now this is just wrong.
You do see that it is called from `addCamera()`, right?

src/librawspeed/decoders/RawDecoder.cpp
```diff
@@ -125,8 +125,6 @@ void RawDecoder::decodeUncompressed(TiffIFD *rawIFD, BitOrder order) {
 
 bool RawDecoder::checkCameraSupported(CameraMetaData *meta, string make,
                                       string model, const string &mode) {
-  TrimSpaces(make);
```
LebedevRI: But it is not guaranteed that the function is called with already-trimmed values?
E.g. see aridecoder, it calls with some string

src/librawspeed/tiff/TiffEntry.h
```diff
@@ -78,7 +78,7 @@ class TiffEntry
   ushort16 getShort(uint32 num=0) const;
   short16 getSShort(uint32 num=0) const;
   float getFloat(uint32 num=0) const;
-  std::string getString() const;
+  std::string getString(bool trimmed = true) const;
```
LebedevRI: Okay, so i think this is the only thing in this pr i don't quite agree with.
Trimming the strings from exif/makernotes by default seems dangerous.

LebedevRI (changes requested):
How about undoing the `TrimSpaces()` changes?

LebedevRI:
I assume this is the very last few function renames?
Maybe should look into [readability-identifier-naming](http://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html) clang-tidy check.

src/librawspeed/decoders/RawDecoder.cpp
axxel: > E.g. see aridecoder, it calls with some string

But if a programmer used that function and passed it an invalid compile-time string (including a space at the end), would the right thing to do in that case not be to fix that string where it is specified instead of fixing it during runtime?

Is seems quite obvious to me that this particular trim here _is_ the 'default' trim someone came up with to deal with the problem of excessive spaces in tiff entries on a more or less central position. I simply put it, where it belongs.

src/librawspeed/tiff/TiffEntry.h
axxel: Would you please explain to me what the danger is exactly?

Please note:
* rstest is happy
* if someone wrote some new code and called this function and trimming was not what he wanted, then he can call the function an pass 'false'.
* based on the code it is more likely that some new code actually needs yet another trim call and the developer forgets about it until he notices it later.

Alternative suggestion, which makes it more obvious that trimming is happening: have two functions, one called `getTrimmedString()` and use that. We'd get about the same effect, just more characters.

axxel:
> I assume this is the very last few function renames?

Well, it is the last from my side for now. I don't know if there are any more somewhere.

src/librawspeed/tiff/TiffEntry.h
LebedevRI: In any case, `getId()` _needs_ to keep calling trim.
It is called from `addCamera()`, which could be called not by rawspeed, and unless the data gets the same treatment (read: trimming), the results may be unexpected.

axxel: > In any case, getId() needs to keep calling trim.

Ok, so I'll add that back. Still: `getString(bool b=true)` or `getTrimmedString()`?

LebedevRI: I'd pick `getTrimmedString()`, because it is **much** easier to search for.

LebedevRI:
I'm waiting for this pr, and then i'll do full IWYU pass.

axxel:
Done

LebedevRI (approved):
Ok, so no auto-trimming :)

axxel:
> Ok, so no auto-trimming :)

Yes and no: see `TiffRootIFD::getID()` :). But I agree that it is a lot better how this turned out now than what I first proposed.

PR #61, opened 2017-01-30
Continue several open ends of the infrastructure cleanup effort

axxel:
1) Continuation of the cleanup of the TiffIFD object/memory handling, that I left off a few weeks back. 
1) Consolidate common code of the Tiff based decoders, introducing AbstractTiffDecoder
1) Splitting LJpegPlain code in two separate use cases: Cr2 and DNG

Details please see the individual commit messages.

### DIFF
--- a/src/librawspeed/decoders/ArwDecoder.h
+++ b/src/librawspeed/decoders/ArwDecoder.h
@@ -23,26 +23,23 @@
 
 #include "common/Common.h"       // for uint32
 #include "common/RawImage.h"     // for RawImage
-#include "decoders/RawDecoder.h" // for RawDecoder, RawDecoderThread (ptr o...
+#include "decoders/AbstractTiffDecoder.h"
 #include "io/FileMap.h"          // for FileMap
 
 namespace RawSpeed {
 
 class ByteStream;
 class CameraMetaData;
-class TiffIFD;
 
-class ArwDecoder :
-  public RawDecoder
+class ArwDecoder final : public AbstractTiffDecoder
 {
 public:
-  ArwDecoder(TiffIFD *rootIFD, FileMap* file);
-  ~ArwDecoder() override;
+  using AbstractTiffDecoder::AbstractTiffDecoder;
+
   RawImage decodeRawInternal() override;
   void checkSupportInternal(CameraMetaData *meta) override;
   void decodeMetaDataInternal(CameraMetaData *meta) override;
   void decodeThreaded(RawDecoderThread *t) override;
-  TiffIFD *getRootIFD() override { return mRootIFD; }
 
 protected:
   int getDecoderVersion() const override { return 1; }
@@ -51,9 +48,8 @@ protected:
   void DecodeUncompressed(TiffIFD* raw);
   void SonyDecrypt(uint32 *buffer, uint32 len, uint32 key);
   void GetWB();
-  TiffIFD *mRootIFD;
   ByteStream *in;
-  int mShiftDownScale;
+  int mShiftDownScale = 0;
 };
 
 } // namespace RawSpeed
--- a/src/librawspeed/decoders/ArwDecoder.h
+++ b/src/librawspeed/decoders/ArwDecoder.h
@@ -45,6 +45,7 @@ public:
   TiffIFD *getRootIFD() override { return mRootIFD; }
 
 protected:
+  int getDecoderVersion() const override { return 1; }
   void DecodeARW(ByteStream &input, uint32 w, uint32 h);
   void DecodeARW2(ByteStream &input, uint32 w, uint32 h, uint32 bpp);
   void DecodeUncompressed(TiffIFD* raw);
--- a/src/librawspeed/decoders/ArwDecoder.h
+++ b/src/librawspeed/decoders/ArwDecoder.h
@@ -40,7 +40,6 @@ public:
     : AbstractTiffDecoder(move(root), file) {}
 
   RawImage decodeRawInternal() override;
-  void checkSupportInternal(CameraMetaData *meta) override;
   void decodeMetaDataInternal(CameraMetaData *meta) override;
   void decodeThreaded(RawDecoderThread *t) override;
 
--- a/src/librawspeed/decoders/ArwDecoder.h
+++ b/src/librawspeed/decoders/ArwDecoder.h
@@ -34,7 +34,10 @@ class CameraMetaData;
 class ArwDecoder final : public AbstractTiffDecoder
 {
 public:
-  using AbstractTiffDecoder::AbstractTiffDecoder;
+  // please revert _this_ commit, once IWYU can handle inheriting constructors
+  // using AbstractTiffDecoder::AbstractTiffDecoder;
+  ArwDecoder(TiffRootIFDOwner&& root, FileMap* file)
+    : AbstractTiffDecoder(move(root), file) {}
 
   RawImage decodeRawInternal() override;
   void checkSupportInternal(CameraMetaData *meta) override;
--- a/src/librawspeed/decoders/Cr2Decoder.cpp
+++ b/src/librawspeed/decoders/Cr2Decoder.cpp
@@ -49,7 +49,6 @@ namespace RawSpeed {
 
 Cr2Decoder::Cr2Decoder(TiffIFD *rootIFD, FileMap* file) :
     RawDecoder(file), mRootIFD(rootIFD) {
-  decoderVersion = 8;
 }
 
 Cr2Decoder::~Cr2Decoder() {
--- a/src/librawspeed/decoders/Cr2Decoder.cpp
+++ b/src/librawspeed/decoders/Cr2Decoder.cpp
@@ -157,36 +157,23 @@ RawImage Cr2Decoder::decodeRawInternal() {
 }
 
 void Cr2Decoder::checkSupportInternal(CameraMetaData *meta) {
-  vector<TiffIFD*> data = mRootIFD->getIFDsWithTag(MODEL);
-  if (data.empty())
-    ThrowRDE("CR2 Support check: Model name not found");
-  if (!data[0]->hasEntry(MAKE))
-    ThrowRDE("CR2 Support: Make name not found");
-  string make = data[0]->getEntry(MAKE)->getString();
-  string model = data[0]->getEntry(MODEL)->getString();
-
+  auto id = mRootIFD->getID();
   // Check for sRaw mode
   if (mRootIFD->getSubIFDs().size() == 4) {
     TiffEntry* typeE = mRootIFD->getSubIFDs()[3]->getEntryRecursive(CANON_SRAWTYPE);
     if (typeE && typeE->getInt() == 4) {
-      this->checkCameraSupported(meta, make, model, "sRaw1");
+      checkCameraSupported(meta, id, "sRaw1");
       return;
     }
   }
 
-  this->checkCameraSupported(meta, make, model, "");
+  checkCameraSupported(meta, id, "");
 }
 
 void Cr2Decoder::decodeMetaDataInternal(CameraMetaData *meta) {
   int iso = 0;
   mRaw->cfa.setCFA(iPoint2D(2,2), CFA_RED, CFA_GREEN, CFA_GREEN, CFA_BLUE);
-  vector<TiffIFD*> data = mRootIFD->getIFDsWithTag(MODEL);
 
-  if (data.empty())
-    ThrowRDE("CR2 Meta Decoder: Model name not found");
-
-  string make = data[0]->getEntry(MAKE)->getString();
-  string model = data[0]->getEntry(MODEL)->getString();
   string mode;
 
   if (mRaw->metadata.subsampling.y == 2 && mRaw->metadata.subsampling.x == 2)
@@ -243,7 +230,7 @@ void Cr2Decoder::decodeMetaDataInternal(CameraMetaData *meta) {
     mRaw->setError(e.what());
     // We caught an exception reading WB, just ignore it
   }
-  setMetaData(meta, make, model, mode, iso);
+  setMetaData(meta, mode, iso);
 }
 
 int Cr2Decoder::getHue() {
--- a/src/librawspeed/decoders/Cr2Decoder.cpp
+++ b/src/librawspeed/decoders/Cr2Decoder.cpp
@@ -47,16 +47,6 @@ using namespace std;
 
 namespace RawSpeed {
 
-Cr2Decoder::Cr2Decoder(TiffIFD *rootIFD, FileMap* file) :
-    RawDecoder(file), mRootIFD(rootIFD) {
-}
-
-Cr2Decoder::~Cr2Decoder() {
-  if (mRootIFD)
-    delete mRootIFD;
-  mRootIFD = nullptr;
-}
-
 RawImage Cr2Decoder::decodeOldFormat() {
   uint32 offset = 0;
   if (mRootIFD->getEntryRecursive(CANON_RAW_DATA_OFFSET))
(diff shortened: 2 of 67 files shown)


axxel:
forgot to `git add` the one new file. pushed again.

LebedevRI:
-ETOOMUCHPR
all this rawspeed clean up, instead of being fun expirience for me, becomes an tedious job to review, with much less time to actually do the stuff i wanted to do

axxel:
> -ETOOMUCHPR

To make sure I don't misunderstand what you just said: Are you trying to _actively_ discouraging me to contribute?

> becomes an tedious job to review

Well, you could limit the time spent on reviewing and simply trust more in the work I do. I think I can claim I proofed my worth as a developer by getting rid of thousands of lines of code, killing quite a few bugs along the way and improving performance substantially while I was at it... Maybe my prs were not completely flawless (and I always appreciate constructive feedback), but neither were yours. And I'd say it was you who put himself in the 'single maintainer seat'.

> with much less time to actually do the stuff i wanted to do

I don't want to prevent in you doing that in any way.

Sorry, but after what I just wrote yesterday about looking forward to keep on work with you, even though we have or different opinions, this is quite an unpleasant surprise.

src/librawspeed/decoders/ArwDecoder.h
```diff
@@ -23,36 +23,33 @@
...
-class TiffIFD;
 
-class ArwDecoder :
-  public RawDecoder
+class ArwDecoder final : public TiffBasedDecoder
 {
 public:
-  ArwDecoder(TiffIFD *rootIFD, FileMap* file);
-  ~ArwDecoder() override;
+  using TiffBasedDecoder::TiffBasedDecoder;
```
LebedevRI: This sadly really breaks IWYU at the moment, https://github.com/include-what-you-use/include-what-you-use/issues/402 :((
Not that i want to insist on avoiding it

src/librawspeed/decoders/TiffBasedDecoder.h
```diff
@@ -0,0 +1,40 @@
...
+
+#pragma once
+
+#include "decoders/RawDecoder.h" // for RawDecoder
+#include "io/FileMap.h"          // for FileMap
+#include "tiff/TiffIFD.h"        // for TiffRootIFDOwner
+
+namespace RawSpeed {
+
+class TiffBasedDecoder : public RawDecoder
```
LebedevRI: Okay, so the fact that this introduces some structure is really awesome!

But, i don't agree with the naming.
Does that mean that `TiffBasedDecoder` creates a decoder (i.e. one of the cr2/dng/etc) from tiff data?
No, the other way around. All the decoders based on `TiffBasedDecoder` are tiff-based.
I think this duality is confusing

LebedevRI (commented):
Introduction of structure is good.
C++ constructor inheritance unfortunately breaks IWYU. I don't know if because of that i want to avoid it though.
Naming is up for debate...

src/librawspeed/decoders/TiffBasedDecoder.h
axxel: > All the decoders based on TiffBasedDecoder are tiff-based.

Exactly! Hence, the name!?! I don't see the problem here. But I'm fine with another suitable name. The name should contain the strings "Tiff" and "Decoder" and some hint at the fact that it is not a complete decoder by itself but a base class for others. Do you have a suggestion?

axxel:
> I don't know if because of that i want to avoid it though.

I'm not sure what you are trying to say here. You approve of disapprove the code improvements? Just in case you are against using this language feature because of the bug in the IWYU tool:
 * it is the right thing to do in terms of available language features
 * the tool-guys acknowledged that it is their bug and they seem to be already looking into it

You could of course add another pr that replaces all the inheriting constructors with manual pass-though constructors for the time being. And once your tool has been fixed you can simply revert that one commit.

LebedevRI:
> I'm not sure what you are trying to say here. You approve of disapprove the code improvements? Just in case you are against using this language feature because of the bug in the IWYU tool:

That is exactly the decision i'm trying to make.

axxel:
> That is exactly the decision i'm trying to make.

I read that statement as "I disapprove of the use of the feature because of the bug in the tool.". Fine. I could add the suggested commit that can later be reverted.

Please try to be a bit more constructive and make alternative suggestions and if I already presented an alternative then provide some feedback to that.

src/librawspeed/decoders/TiffBasedDecoder.h
LebedevRI: So why do you not like `Abstract` prefix so much?

axxel: That is fine with me. To explicitly spell it out: `AbstractTiffEncoder`?

My 'problem' with this prefix is just that I think the prefix is justified when you want to explicitly state that this class is abstracting away some behavior. While that is true for the suggested class, it is not really because of the Tiff stuff but the RawEncoder stuff. The tiff-part does not add any abstract methods. But that is really a minute detail and subjective as well. I'm totall fine with `AbstractTiffEncoder`.

EDIT: `AbstractTiffDecoder`, not _en_-coder of course...

LebedevRI: Yes, `AbstractTiffDecoder`.
I do undestand that part about tiff, but still, this class is used as a base for the actual decoders, and is not (as far as i can see?) used as a decoder by itself. So *some* indication of that is needed. I think such a prefix makes sense...

axxel: Very good. To make some progress and get this off my table, I already did that also rebased to latest develop and added the mentioned 'old-school pass-through constructor' workaround for the IWYU bug.

LebedevRI: Cool! Hopefully will be merging this and #60 tomorrow...
I believe there are a few more `using X::X;` in the codebase, namely in decoders.

src/librawspeed/decoders/DngDecoderSlices.cpp
```diff
@@ -158,10 +159,9 @@ void DngDecoderSlices::decodeSlice(DngDecoderThread* t) {
     while (!t->slices.empty()) {
       DngSliceElement e = t->slices.front();
       t->slices.pop();
-      LJpegPlain l(*mFile, e.byteOffset, e.byteCount, mRaw);
-      l.mDNGCompatible = mFixLjpeg;
+      LJpegDecompressor d(*mFile, e.byteOffset, e.byteCount, mRaw);
```
LebedevRI, line 162: Nice! :)

axxel: Glad you like it. As I said: I'd like to get this off my plate. Once this is merged, I'll wrap up the other open pr and maybe 1-2 additional 'loose ends' from the work done earlier.

LebedevRI:
by all means, do continue piling things up
the bigger the diff, the more things you can pile :)

LebedevRI:
do tell when you are done, and i can start looking at the diff

axxel:
> the bigger the diff, the more things you can pile :)

:) You've seen most of the stuff already. The LJpegPlain split patch was part of the other PR. The last 3 are rather small. Only new/big is the `Consolidate more common code of the Tiff based decoders` from last night.

I just want to complete some TODOs that I came across during the past weeks, e.g. fixing the inconsistent (old) getter function names of DataBuffer vs TiffEntry and similar cosmetic stuff like that. I can continue until I'm done or take a brake now. Whatever you prefer.

src/librawspeed/decompressors/AbstractLJpegDecompressor.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include "io/IOException.h"               // for IOException
+#include <array>                          // for array
+#include <memory>                         // for unique_ptr, allocator
+#include <vector>                         // for vector
+
+namespace RawSpeed {
+
+void AbstractLJpegDecompressor::decode() {
+  if (getNextMarker(false) != M_SOI)
+    ThrowRDE("LJpegDecompressor: Image did not start with SOI. Probably not an LJPEG");
```
shoffmeister, line 37: Would LJpegDecompressor in the throw expression message have to be adjusted to AbstractLJpegDecompressor? 

(Systematic change)

LebedevRI:
I preferred to merge this today, and expressed this intent yesterday (i think i did?)
I even pretty much finished going through the diff

So not to waste anyone's time, please do pick a point where to stop adding changes and tell me that it is ready :)

axxel:
> I preferred to merge this today, and expressed this intent yesterday (i think i did?)

yes, you did.

> tell me that it is ready :)

It is ready.

src/librawspeed/decompressors/AbstractLJpegDecompressor.cpp
axxel: Thanks for spotting this. My QtCreator->refactor->rename tool does not touch strings, of course. But the whole exception handling and error message situation could use some love in general. The _real_ issue here is this manually typed typename in the first place (in my opinion). So maybe the best / most pragmatic approach would be to address that in general (at another time)?

shoffmeister: Personally, I'd approach this as follows:
1.) fix the string immediately (aka "do not lie") in that class
2.) review whether printing the class name adds value at all (uninformed opinion - no, it does not; debug builds ought to throw the file name which might be good enough?). If there is value in the name, http://stackoverflow.com/questions/1666802/is-there-a-class-macro-in-c seems to be one way to go. That looks like pain, though.

(Disclaimer: my C++ is awfully rusty, I have no mental model of the source code at large, and I barely remember anything about failure output modes on various OSes. Buyer beware :) )

axxel: @LebedevRI please let me know if you wait for me to replace the class names in the error strings before you want to merge.

> review whether printing the class name adds value at all (uninformed opinion - no, it does not

I agree with you: it does not. I have a somewhat informed opinion about it and will add an issue to discuss this subject on its own thread.

LebedevRI: No, do not change anything.
I'll be fixing that along with slight deduplication manually.

src/librawspeed/decoders/RawDecoder.h
```diff
@@ -170,12 +170,12 @@ class RawDecoder
...
   FileMap *mFile;
 
-  /* Decoder version - defaults to 0, but can be overridden by decoders */
+  /* Decoder version */
   /* This can be used to avoid newer version of an xml file to indicate that a file */
   /* can be decoded, when a specific version of the code is needed */
   /* Higher number in camera xml file: Files for this camera will not be decoded */
   /* Higher number in code than xml: Image will be decoded. */
-  int decoderVersion;
+  virtual int getDecoderVersion() const = 0;
```
LebedevRI, line 178: does this ensure that it must be implemented by *each* decoder?

axxel: Yes, the compilation fails if this method is not implemented in each sub-class (or sub-sub-class, etc.) EDIT: 'pure virtual function' is the term to google for.

src/librawspeed/decoders/Cr2Decoder.cpp
```diff
@@ -131,11 +130,10 @@ RawImage Cr2Decoder::decodeNewFormat() {
   TiffEntry* offsets = raw->getEntry(STRIPOFFSETS);
   TiffEntry* counts = raw->getEntry(STRIPBYTECOUNTS);
 
-  LJpegPlain l(*mFile, offsets->getInt(), counts->getInt(), mRaw);
-  l.addSlices(s_width);
+  Cr2Decompressor d(*mFile, offsets->getInt(), counts->getInt(), mRaw);
```
LebedevRI, line 133: All these decompressors have basically the same constructor, which makes sense.
There should be `AbstractDecompressor`, similar to the `AbstractTiffDecoder`

LebedevRI (approved):

src/librawspeed/decoders/Cr2Decoder.cpp
axxel: Hm, I just had a look at the new `JpegDecompressor` and `DeflateDecompressor` and I think the design to wrap the decode function into such a class makes the code more complicated that helpful. The decode function needs a couple parameters. Now some are passed to the constructor of the class, some are passed to the member function. This is arbitrary. Also the use case is the same every time:
```
   SomeDecompressor d(some, parameters);
   d.decode( some, more, parameters ); // executed exactly once
   // destroy d
```
For the `Decoders` it is something completely different, because you really have an abstraction there: the client gets a pointer to some decoder implementation and the type `RawDecoder` does not know which it is. With the decompressors, everything is fully known.

I suggest to model this functionality as a dead simple function call. See `NikonDecompressor`.

LebedevRI: *with current* code.
If it is a function, how will it be able to cache memory allocations e.g.?
Yes, it should not allocate memory each time it is called.
In fact, down the road, i will highly prefer to fix all the allocations, i.e. they should only happen in the main thread.

axxel: > If it is a function, how will it be able to cache memory allocations e.g.?

Not at all (without nasty `static` tricks of course). But it does not - that is what I tried to explicitly hint at with the `// executed exactly once` comment.

> i.e. they should only happen in the main thread.

Something like that will certainly require substantial changes. I think it is better to then introduce exactly what is necessary to achieve that than to now introduce something that is not used now but then might or might not be used in the current form later.

LebedevRI:
Hmm, that is a bit too much new static analysis issues...

src/librawspeed/decoders/Cr2Decoder.cpp
LebedevRI: >` // executed exactly once`

Are you talking in general, or about these {jpeg,deflate} decompressors?
They are not called exactly once, they are called from several worker threads, once per slice.

LebedevRI: (hint: DeflateDecompressor already does it, at least partially)

axxel: > Are you talking in general, or about these {jpeg,deflate} decompressors?

About all of them, actually. The difference with the LJpeg decompressors is that there you actually have a use for the internal modeling of the `AbstractLJpegDecompressor` class, because the different variants share common code. But the usage of such code could as well be modeled with a single function:
```
void decompressCr2( const Buffer& data, Buffer::size_type offset,
                      Buffer::size_type size, const RawImage& img,
                      vector<int> slicesW)
{
   Cr2Decompressor(data, offset, size, img) d;
   d.decode(slicesW);
}
```

I meant in the current code there is an object instantiation and then exactly one call to a method of that object. Whether there are several threads doing both at the same time is irrelevant to my point.

axxel:
> Hmm, that is a bit too much new static analysis issues...

I had quick glance over them and at least some are not 'new'. I noticed this in the last report e-mail already. No idea how coverty is defining 'new'. Anyway: I'll take the opportunity and fix the ones from the report that don't have anything to do with the testing code.

axxel:
I went through all the (non testing) issues and none of them were new. I'm even convinced that most of the memory issues (BAD_FREE, LEAK, etc.) are actually false alarms. Maybe coverty has an issue with r-value references and move semantics. A real leak or double free would have been detected by ASAN.

LebedevRI:
> Maybe coverty has an issue with r-value references and move semantics. 

That would be really bad :(

> A real leak or double free would have been detected by ASAN.

True

LebedevRI:
It seems like using gcc-6 for coverity scan helped, a lot of issues got marked as fixed.

axxel:
Nice! :).

PR #65, opened 2017-02-01
Add support for writing RGB color ppm files from 'rstest -d'

axxel:
Very helpfull when trying to debug issues with Canon mRaw/sRaw images.

### DIFF
--- a/src/utilities/rstest/rstest.cpp
+++ b/src/utilities/rstest/rstest.cpp
@@ -149,9 +149,12 @@ void writePPM(const RawImage& raw, const string& fn) {
 
   int width = raw->dim.x;
   int height = raw->dim.y;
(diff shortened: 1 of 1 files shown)


src/utilities/rstest/rstest.cpp
```diff
@@ -149,17 +149,20 @@ void writePPM(const RawImage& raw, const string& fn) {
 
   int width = raw->dim.x;
   int height = raw->dim.y;
+  string format = raw->getCpp() == 1 ? "P5" : "P6";
```
LebedevRI, line 152: cpp can be 1, or 3 or 4
(there can be other cases, but dt does not handle them anyway)
just passthrough-ing the cpp will just work?

also, the getDataType() can return float, and then you may want to write PFM

PR #67, opened 2017-02-02
ColorFilterArray: fix negative shift UB (coverty report)

axxel:
Also add test for negative y shift.

### DIFF
--- a/src/librawspeed/metadata/ColorFilterArrayTest.cpp
+++ b/src/librawspeed/metadata/ColorFilterArrayTest.cpp
@@ -135,6 +135,7 @@ TEST_P(ColorFilterArrayTest, DcrawFilterShift1) {
     ASSERT_EQ(ColorFilterArray::shiftDcrawFilter(rggb, 0, 1), gbrg);
     ASSERT_EQ(ColorFilterArray::shiftDcrawFilter(rggb, 1, 1), bggr);
     ASSERT_EQ(ColorFilterArray::shiftDcrawFilter(rggb, 0, 2), rggb);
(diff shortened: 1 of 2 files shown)


src/librawspeed/metadata/ColorFilterArrayTest.cpp
```diff
@@ -135,6 +135,7 @@ TEST_P(ColorFilterArrayTest, DcrawFilterShift1) {
     ASSERT_EQ(ColorFilterArray::shiftDcrawFilter(rggb, 0, 1), gbrg);
     ASSERT_EQ(ColorFilterArray::shiftDcrawFilter(rggb, 1, 1), bggr);
     ASSERT_EQ(ColorFilterArray::shiftDcrawFilter(rggb, 0, 2), rggb);
+    ASSERT_EQ(ColorFilterArray::shiftDcrawFilter(rggb, 0, -1), gbrg);
```
LebedevRI, line 138: JFYI, this test executes all the same shiftDcrawFilter checks 255 times.
But i'l rewrite that locally.

axxel: I thought so but did not have the courage to look into the details of the google test framework to fix that. Thanks for taking care of that.

LebedevRI:
Thank you for fixing `shiftLeft()` / `shiftDown()` to work with negative offsets :)

src/librawspeed/metadata/ColorFilterArrayTest.cpp
LebedevRI: There, now this executable will run 6744 tests :)

Issue #19, opened 2017-01-06
getCamera() with prefix-search issue


LebedevRI:
https://github.com/klauspost/rawspeed/issues/166

I added that api in https://github.com/klauspost/rawspeed/commit/a45724ec3c201b67f5ff732317921c0feca878c6 and https://github.com/klauspost/rawspeed/commit/e1b5ccae3515773755621cee305170fa5ea63808.
And now i have just discovered how it backfires.
With raw file from https://redmine.darktable.org/issues/11354 from `Canon EOS Rebel T6`, which currently has broken alias, it matches to `Canon EOS Rebel T6i`.
That happens because internally "make model mode" is a string, https://github.com/klauspost/rawspeed/blob/a45724ec3c201b67f5ff732317921c0feca878c6/RawSpeed/CameraMetaData.h#L37

tredlie:
This issue still seems to be around. What are the chances of getting this fixed as it brings darktable usage for a Leica M9 to a grinding halt, reporting this camera to be a Mamiya.
Or, maybe, this does not do any harm other than reporting the wrong Maker in the Image Information window in DT. The exported JPG looks OK.

LebedevRI:
I believe i have just fixed it locally. At least the unittest i just wrote no longer fails.

PR #68, opened 2017-02-03
CameraMetaData: fix search

LebedevRI:
See #19

### DIFF
--- a/src/librawspeed/metadata/CameraMetaData.h
+++ b/src/librawspeed/metadata/CameraMetaData.h
@@ -23,10 +23,22 @@
 #include "common/Common.h" // for uint32
 #include <map>             // for map
 #include <string>          // for string
+#include <tuple>           // for tuple
 
 namespace RawSpeed {
 
 class Camera;
+
+struct CameraId {
+  std::string make;
+  std::string model;
+  std::string mode;
+
+  bool operator<(const CameraId& rhs) const {
+    return std::tie(make, model, mode) <
+           std::tie(rhs.make, rhs.model, rhs.mode);
+  }
+};
 
 class CameraMetaData
 {
@@ -34,7 +46,7 @@ public:
   CameraMetaData();
   CameraMetaData(const char *docname);
   virtual ~CameraMetaData();
-  std::map<std::string,Camera*> cameras;
+  std::map<CameraId, Camera*> cameras;
   std::map<uint32,Camera*> chdkCameras;
 
   // searches for camera with given make + model + mode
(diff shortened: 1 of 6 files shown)


src/librawspeed/metadata/CameraMetaData.h
```diff
@@ -23,18 +23,21 @@
...
 #include <map>             // for map
 #include <string>          // for string
+#include <tuple>           // for tuple
 
 namespace RawSpeed {
 
 class Camera;
 
+using CameraId = std::tuple<std::string, std::string, std::string>;
+
```
axxel: This obviously duplicates/colides with the work I did last night (see TiffID). I suggest to use a trivial struct like the one I introduced instead of a tuple because it is _way_ less likely to mix up the meaning of the individual members if they have a name (make/model/mode) instead of just an index in a tuple.

Maybe you could merge the two efforts if you moved/renamed my TiffID struct and added a mode string there. It would unify the whole 'id' related information?

axxel: Also related to https://github.com/darktable-org/rawspeed/pull/61/commits/049d0ebe5cf20f9deabb9063fcf2e5061c0ff7eb

LebedevRI: Ah, i forgot about [`std::tie`](http://en.cppreference.com/w/cpp/utility/tuple/tie), then the `operator<` i was trying to avoid by using tuple is trivial, okay.

> Maybe you could merge the two efforts if you moved/renamed my TiffID struct and added a mode string there. It would unify the whole 'id' related information?

I did not look at that pr in *that detail*, so i have no clue what you are talking about. But since i'll need to pick this change to dt-2.2.x, this needs to be as non-intrusive as possible.

axxel: > this needs to be as non-intrusive as possible.

I understand. This change and my similar `TiffID` can certainly coexist. It just wanted to point out that in the end they should probably be merged. I was actually thinking about introducing a 'global' CameraID type to be used for both but also wanted to keep it 'minimal'.

LebedevRI:
Ugh, do not understand why that specific test fails, only on that platform+compiler

axxel:
I pulled the code and it runs fine on my lnux box. no surprises there. I guess the problem must be one of the two ASSERT_ANY_THROW cases. I only see the obvious explanation that there is bug in the compiler/gtest framework. Maybe just disable those two tests for now?

LebedevRI:
rstest is not happy about this, but all the changes are a definite fixes.

axxel:
> but all the changes are a definite fixes.

Hm, even that one?
```
-canonical_make: Ricoh
+canonical_make: RICOH IMAGING COMPANY, LTD.
```

Disclaimer: I have no proper understanding of what the canonical_make is used for exactly.

LebedevRI:
Does not look wrong, so i assume this one too.

axxel:
> Does not look wrong, so i assume this one too.

Ok. My limited understanding of the canonical_make thing was to 'merge' stuff like "NIKON" and "NIKON CORPORATION". That new string looked like the latter. But if there now is only one Richo in the form of "RICOH IMAGING COMPANY, LTD." then I guess all is good, indeed.

PR #71, opened 2017-02-04
A little bit of everything

LebedevRI:


src/librawspeed/common/AbstractException.h
```diff
@@ -18,27 +19,33 @@
...
-  static char buf[8192];
   vsnprintf(buf, 8192, fmt, val);
   va_end(val);
   writeLog(DEBUG_PRIO_EXTRA, "EXCEPTION: %s\n", buf);
-  throw FileIOException(buf);
+  throw T(buf);
 }
 
+class AbstractException : public std::runtime_error {
+public:
```
axxel: +1 for introducing a proper base class! But here the term "Abstract" seems really questionable. `std::exception` is also not called `std::abstract_exception`. Maybe just `Exception`? The namespace makes it sufficiently distinct. Or `RawSpeedException`. It would be the name that might show up in client code because someone want's to have it a little more specific than `std::exception`.

LebedevRI: Yeah, in *this* case, i think i'll go with some different name, probably `RawSpeedException`.
`Exception` is too similar to `std::exception` (also, hate that filenames are not lowercase)

src/librawspeed/decoders/Cr2Decoder.h
```diff
@@ -43,19 +43,11 @@ class Cr2Decoder final : public AbstractTiffDecoder
   void decodeMetaDataInternal(CameraMetaData *meta) override;
 
 protected:
-  int sraw_coeffs[3];
-
```
axxel, line 47: +1 :)

src/librawspeed/common/RawImage.h
```diff
@@ -112,24 +112,24 @@ class RawImageData
...
   iPoint2D getCropOffset();
   virtual void scaleBlackWhite() = 0;
   virtual void calculateBlackAreas() = 0;
   virtual void setWithLookUp(ushort16 value, uchar8* dst, uint32* random) = 0;
-  virtual void sixteenBitLookup();
-  virtual void transferBadPixelsToMap();
-  virtual void fixBadPixels();
+  virtual void sixteenBitLookup() final;
+  virtual void transferBadPixelsToMap() final;
+  virtual void fixBadPixels() final;
```
axxel: Please remove the `virtual` keyword here (and in similar places) instead of adding `final`. This is a bog standard method, see e.g. `getCropOffset()` above. Declaring something `virtual` and `final` in the base class makes no sense to me. (There have been way too many virtual keywords spread around the code).

cmake/Modules/thread-local.cmake
```diff
@@ -0,0 +1,25 @@
+include(CheckCXXSourceCompiles)
+
+CHECK_CXX_SOURCE_COMPILES("
+static thread_local int tls;
+int main(void)
+{
+  return 0;
+}" HAVE_THREAD_LOCAL)
```
axxel, line 8: `thread_local` is standard c++11 and since we require a conforming compiler, why not use that without any extra checks and alternatives?

LebedevRI: Because we live in the real world.
Even the latest(as used by travis-ci osx builds) XCode compiler does not support it.

axxel: Ahhh, Apple... always good for an unpleasant surprise. :-/

axxel:
I get this a bunch of similar errors:
```
CommonTest.cpp:45:76: error: converting to ‘powerOfTwoType {aka std::tuple<int, bool>}’ from initializer list would use explicit constructor ‘constexpr std::tuple<_T1, _T2>::tuple(_U1&&, _U2&&) [with _U1 = int; _U2 = bool; <template-parameter-2-3> = void; _T1 = int; _T2 = bool]’
     {6, false}, {7, false}, {8, true}, {9, false}, {10, false}, {11, false}};
                                                                            ^
```
Any ideas?

axxel:
Context: compiler = `c++ (Ubuntu 5.2.1-22ubuntu2) 5.2.1 20151010`, command line:
```
/usr/bin/c++   -DDEBUG -DHAVE_JPEG -DHAVE_POSIX_MEMALIGN -DHAVE_PUGIXML -DHAVE_THREAD_LOCAL -DHAVE_ZLIB
-std=c++11 -fno-strict-aliasing -Wall -Wformat=2 -Wshadow -Wvla -Wextra -Wno-unused-parameter -Wsuggest-attribute=noreturn
-Wsuggest-final-types -Wsuggest-final-methods -Wno-error=suggest-final-types -Wno-error=suggest-final-methods
-Wframe-larger-than=4096 -Wstack-usage=4096 -Wlarger-than=32768 -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls
-fno-common -U_FORTIFY_SOURCE -fsanitize=address -fstack-protector-strong -fPIE
-I/home/axel/contrib/rawspeed.build/src -I/home/axel/contrib/rawspeed.git/src/librawspeed
-isystem /home/axel/contrib/rawspeed.build/googletest/googletest-src/googletest/include
-isystem /home/axel/contrib/rawspeed.build/googletest/googletest-src/googlemock/include
-march=native -g3 -ggdb3 -Werror -o CMakeFiles/CommonTest.dir/__/common/CommonTest.cpp.o
-c /home/axel/contrib/rawspeed.git/src/librawspeed/common/CommonTest.cpp
```

LebedevRI:
I guess i will have to bring back gcc-4.8 build entry

src/librawspeed/common/Common.h
```diff
@@ -177,10 +180,9 @@ extern "C" { FILE __iob_func[3] = { *stdin,*stdout,*stderr }; }
 #endif
 
 // clampBits clamps the given int to the range 0 .. 2^n-1
-inline uint32 clampBits(int x, uint32 n)
-{
-  const int tmp = (1 << n) - 1;
-  x = x < 0 ? 0 : x;
+inline uint32 clampBits(int64 x, uint32 n) {
```
axxel, line 183: It seems CID 1402217 is finding it problematic that you upgraded x from 'int' to 'int64'. I was about to question that before, since the code has no use for that (it gets called with an int and n is always less than 17). It seemed to me you wanted to support n up to 32 and thought "why complain, it should not hurt". Just saying: the easy fix for the CID seems to go back to `int` and limit your tests to a max of `n == 31`. Maybe add an `assert(n < 32);` for good measure.

LebedevRI: With int, the test did not work on windows

PR #75, opened 2017-02-05
Complete the TiffIFD/TiffEntry interface cleanup

axxel:
* make getter function names consistent
* complete const-correctness of TiffIFD interface

### DIFF
--- a/src/librawspeed/decoders/AbstractTiffDecoder.cpp
+++ b/src/librawspeed/decoders/AbstractTiffDecoder.cpp
@@ -1,0 +1,50 @@
+/*
+    RawSpeed - RAW file decoder.
+
+    Copyright (C) 2017 Axel Waggershauser
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2 of the License, or (at your option) any later version.
+
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public
+    License along with this library; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+*/
+
+#include "decoders/AbstractTiffDecoder.h" // for AbstractTiffDecoder
+#include "tiff/TiffIFD.h"                 // for TiffRootIFD
+#include "tiff/TiffEntry.h"               // for TiffEntry
+#include "decoders/RawDecoderException.h" // for ThrowRDE
+
+namespace RawSpeed {
+
+const TiffIFD* AbstractTiffDecoder::getIFDWithLargestImage(TiffTag filter) const
+{
+  std::vector<TiffIFD*> ifds = mRootIFD->getIFDsWithTag(filter);
+
+  if (ifds.empty())
+    ThrowRDE("AbstractTiffDecoder: No suitable IFD with tag 0x%04x found.",
+             filter);
+
+  auto res = ifds[0];
+  uint32 width = res->getEntry(IMAGEWIDTH)->getU32();
+  for (auto ifd : ifds) {
+    TiffEntry* widthE = ifd->getEntry(IMAGEWIDTH);
+    // guard agains random maker note entries with the same tag
+    if (widthE->count == 1 && widthE->getU32() > width) {
+      res = ifd;
+      width = widthE->getU32();
+    }
+  }
+
+  return res;
+}
+
+} // namespace RawSpeed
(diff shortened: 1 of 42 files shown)


src/librawspeed/decoders/AbstractTiffDecoder.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "decoders/RawDecoderException.h" // for ThrowRDE
+
+namespace RawSpeed {
+
+const TiffIFD* AbstractTiffDecoder::getIFDWithLargestImage(TiffTag filter) const
+{
+  std::vector<const TiffIFD*> ifds = mRootIFD->getIFDsWithTag(filter);
+
+  if (ifds.empty())
+    ThrowRDE("AbstractTiffDecoder: No suitable IFD with IMAGEWIDTH tag found.");
```
LebedevRI: with `filter` you mean

axxel: absolutely right. see update.

axxel:
@LebedevRI Please merge (as early as convenient :)).

src/librawspeed/decoders/MosDecoder.cpp
```diff
@@ -124,17 +123,15 @@ RawImage MosDecoder::decodeRawInternal() {
...
+    raw = mRootIFD->getIFDWithTag(TILEOFFSETS);
     off = raw->getEntry(TILEOFFSETS)->getU32();
-  } else {
-    data = mRootIFD->getIFDsWithTag(CFAPATTERN);
-    if (!data.empty()) {
-      raw = data[0];
-      off = raw->getEntry(STRIPOFFSETS)->getU32();
-    } else
-      ThrowRDE("MOS Decoder: No image data found");
+  } catch(...) {
```
LebedevRI: using exceptions for normal codepath handling, wtf c++?

src/librawspeed/decoders/Rw2Decoder.cpp
```diff
@@ -255,19 +253,14 @@ void Rw2Decoder::decodeMetaDataInternal(CameraMetaData *meta) {
...
-  // bool isOldPanasonic = false;
+  const TiffIFD* raw;
 
-  if (data.empty()) {
-    if (!mRootIFD->hasEntryRecursive(STRIPOFFSETS))
-      ThrowRDE("RW2 Decoder: No image data found");
-    // isOldPanasonic = true;
-    data = mRootIFD->getIFDsWithTag(STRIPOFFSETS);
+  try {
+    raw = mRootIFD->getIFDWithTag(PANASONIC_STRIPOFFSET);
```
LebedevRI: i really don't like this pattern

LebedevRI (changes requested):
i *really* don't like that pattern of using exception handling as general-purpose codeflow mechaninsm that you are adding e.g. in `MosDecoder.cpp`.

axxel:
Well, there is _tons_ of 'prior art' in the code base for using exceptions to deal with 'non fatal' error conditions and it is not time critical either. So I don't see any reason why this would be an issue. But I see even less of a point in starting an argument over this detail. See appended patch.

axxel:
@LebedevRI you probably noticed, but just in case you did not and are waiting for this to be fixed: the compile issue seems to be caused by your recent gtest-option commit or something else build system related.

LebedevRI:
@axxel if the build breaks, i have all kinds of bells and whistles that start start signalling, so i do notice :)

axxel:
> so i do notice :)

:). Your usual approach to have a 'next' pr to get ci-feedback before merging seemed like a valid strategy to prevent breaking the 'default' build.

LebedevRI:
Just re-triggering the build does not seem to help, so please rebase.

LebedevRI:
> Well, there is tons of 'prior art' in the code base for using exceptions to deal with 'non fatal' error conditions and it is not time critical either. So I don't see any reason why this would be an issue. 

Since you want to touch everything immediately, you leave me no other option on how to change 'prior art' than to complain about it when i see it in the diff

src/librawspeed/decoders/Rw2Decoder.cpp
LebedevRI: this is still here

src/librawspeed/decoders/Rw2Decoder.cpp
```diff
@@ -255,24 +251,19 @@ void Rw2Decoder::decodeMetaDataInternal(CameraMetaData *meta) {
...
 
-  // bool isOldPanasonic = false;
-
-  if (data.empty()) {
-    if (!mRootIFD->hasEntryRecursive(STRIPOFFSETS))
-      ThrowRDE("RW2 Decoder: No image data found");
-    // isOldPanasonic = true;
-    data = mRootIFD->getIFDsWithTag(STRIPOFFSETS);
+  try {
+    raw = mRootIFD->getIFDWithTag(PANASONIC_STRIPOFFSET);
```
LebedevRI: this is still here

axxel: indeed

LebedevRI (approved):
Okay, seems to work :)
Will merge.

LebedevRI:
@axxel thank you.

Issue #76, opened 2017-02-05
Cleanup basic int type usage and type naming


axxel:
I have 3 issues with the custom types specified here: https://github.com/darktable-org/rawspeed/blob/develop/src/librawspeed/common/Common.h#L35-L41

1) Their **naming** contains the size info twice, e.g. `ushort16` and the names (strings) are not equally long.

I like to suggest two alternative variants:
 * Rust names (`u32, i16, etc.`)
 * `<cstdint>`: http://en.cppreference.com/w/cpp/types/integer

I'd prefer the short Rust names, especially because the std names are longer than what we have right now and the `_t` postfix looks very 'c-ish' to me. I'd also suggest to keep the word `byte` and even use that term as a type name for the `unsigned char` case. That name even almost made it into c++17. The idea is to distinguish between simple data bytes and characters (as in bits of a string).

2) The **usage** of the `uint32` for offets/sizes is 'non-standard'

I'd like to go over the interfaces that I touched and simply use `size_t` for that, even in the cases where I introduced the `size_type` name that can be found in the std:: containers. I always found that part a bit 'over engineered'.

3) Their definition is potentially wrong (depending on the data model, see http://en.cppreference.com/w/cpp/language/types). E.g. the tiff code breaks if 'unsigned int' would be 16bits wide. This should be non-controversial and is trivial to fix.

@LebedevRI What is your opinion on 1) and 2)?

LebedevRI:
I'm wiling to bet that they are not Rust names, but linux; [Assigning an Explicit Size to Data Items](https://static.lwn.net/images/pdf/LDD3/ch11.pdf), [linux/types.h](https://github.com/torvalds/linux/blob/master/include/linux/types.h)

axxel:
> I'm wiling to bet that they are not Rust names, but linux

Meaning what exactly in the context of my question? If you have an opinion about 1) and/or 2), please tell me about it.

LebedevRI:
No, at this point i'm unable to state my opinion/choice.
But, in either case the changes will be rather intrusive so it will be better to not rush changing it and creating a pr.

axxel:
> No, at this point i'm unable to state my opinion/choice.

Ok.

> But, in either case the changes will be rather intrusive so it will be better to not rush changing it and creating a pr.

[...]

PR #80, opened 2017-02-10
Reimplement dng opcodes

### LINKED ISSUE #79: Bad pixel correction for DNGs is broken
See https://redmine.darktable.org/issues/10078.

Example image: https://raw.pixls.us/getfile.php/683/raw/RAW_CANON_POWERSHOT_SX130IS_CHDK.DNG

I stumbled on this issues while cleaning up the DngOpcode classes. This has almost certainly never worked (I found 2 separate issues).

I have a complete rewrite of the DngOpcode classes, fixing this issue (among others). Just waiting for #75 to be merged.

axxel:
Fix #79 and a bunch of other issues (interface bloat, copy'n'paste errors, code duplication, leaks on exceptions). Details see commit messages.

### DIFF
--- a/src/librawspeed/common/DngOpcodes.cpp
+++ b/src/librawspeed/common/DngOpcodes.cpp
@@ -1,699 +1,425 @@
-
 /*
-RawSpeed - RAW file decoder.
+    RawSpeed - RAW file decoder.
 
-Copyright (C) 2012 Klaus Post
+    Copyright (C) 2009-2014 Klaus Post
+    Copyright (C) 2017 Axel Waggershauser
 
-This library is free software; you can redistribute it and/or
-modify it under the terms of the GNU Lesser General Public
-License as published by the Free Software Foundation; either
-version 2 of the License, or (at your option) any later version.
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2 of the License, or (at your option) any later version.
 
-This library is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-Lesser General Public License for more details.
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    Lesser General Public License for more details.
 
-You should have received a copy of the GNU Lesser General Public
-License along with this library; if not, write to the Free Software
-Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+    You should have received a copy of the GNU Lesser General Public
+    License along with this library; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 */
 
+#include "common/Common.h"                // for uint32...
+#include "common/Point.h"                 // for iRectangle2D
 #include "common/DngOpcodes.h"
 #include "common/RawImage.h"              // for RawImage, RawImageData
 #include "common/Threading.h"             // for pthread_mutex_lock, pthrea...
 #include "decoders/RawDecoderException.h" // for ThrowRDE
 #include "io/Endianness.h"
-#include "tiff/TiffEntry.h" // for TiffEntry
-#include <algorithm>        // for min
-#include <cmath>            // for lround, pow
-#include <cstddef>          // for NULL, size_t
+#include "tiff/TiffEntry.h"               // for TiffEntry
+#include <cmath>                          // for pow
 
 using namespace std;
 
 namespace RawSpeed {
 
+class DngOpcode
+{
+public:
+  virtual ~DngOpcode() = default;
+
+  // Will be called once before processing.
+  // Can be used for preparing pre-calculated values, etc.
+  virtual void setup(const RawImage& ri) {}
+
+  // Will be called for actual processing.
+  virtual void apply(RawImage& ri) = 0;
+};
+
+// ****************************************************************************
+
+class FixBadPixelsConstant : public DngOpcode
+{
+  uint32 value;
+
+public:
+  FixBadPixelsConstant(ByteStream& bs)
+  {
+    value = bs.getU32();
+    bs.getU32(); // Bayer Phase not used
+  }
+
+  void setup(const RawImage& ri) override
+  {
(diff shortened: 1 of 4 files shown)


src/librawspeed/common/Common.h
```diff
@@ -130,21 +130,39 @@ inline Endianness getHostEndianness() {
...
-  if (bswap) {
-    switch(sizeof(T)) {
-    case 1: break;
-    case 2: ret = BSWAP16(ret); break;
-    case 4: ret = BSWAP32(ret); break;
-    case 8: ret = BSWAP64(ret); break;
-    }
-  }
-  return ret;
+  return bswap ? getByteSwapped(ret) : ret;
```
LebedevRI: While there, can you add at some very least basic test for this area of code (`getByteSwapped()`) please?

axxel: Well, I'd consider those trivial functions not really worth writing elaborate tests for (I admire your patience in this regard looking at your latest pr :)). But I got this wrong the first time, so I kind of proved myself wrong. I'll add something to your new `CommonTests.cpp` once it is merged.

LebedevRI: Problem being, rstest is only run manually, while the tests are run by all the ci, and by OBS (with non-x86, and big-endian builds, hurray). So unless it is a test, problems are likely to stay unnoticed...

src/librawspeed/common/DngOpcodes.cpp
```diff
@@ -1,698 +1,424 @@
...
+  }
+};
 
+using ScalePerRow = ScalePerRowOrCol<SelectY>;
+using ScalePerCol = ScalePerRowOrCol<SelectX>;
 
-RawImage& OpcodeScalePerCol::createOutput( RawImage &in )
+// ****************************************************************************
+
+DngOpcodes::DngOpcodes(TiffEntry *entry)
```
LebedevRI, line 362: I don't get it. How is it so different that it now results in more than twice the amount of stack frame size it used to use
And moving of the function does not help with diff either

axxel: This OSX + stack usage situation is slightly annoying. You seem to have been struggling with it for a while already. This particular constructor hardly changed at all. The `sizeof(ByteStream)` is neglectable. Since there are only very select places where the mac-code differs at all: why not completely disable the error on mac and be happy as long as the linux builds don't blow a reasonable stack size limit?

LebedevRI: 2cd77a573316143b72d5ed0b2845b8002b50b807

axxel: works, thanks.

src/librawspeed/common/Common.h
LebedevRI: Looking at this in more detail, `common.h` is still kinda bloaty.
i'll pick the first commit, and then move this endianness stuff into new header

LebedevRI: @axxel ok, endianness stuff is in, with full tests.
please rebase this pr, and drop that commit from it
Hopefully will merge at least one of your pr's tomorrow...

axxel: Two points of constructive feedback:
 * If you had simply merged that pr as is and did your split of Common.h afterwards, it would have cost you the same amount of work but spared me about 5 minutes for the manual rebase.
 * Your full tests just increased the runtime of the tests by factor of nearly 10. This is out of proportion. If you checked that e.g. `0x0102 == swap(0x0201)` then there is no point at all in also testing `0x0103`. The whole set of swap functions can be tested with full coverage in a fraction of a second (instead of somewhere near 25s on my machine).

LebedevRI: > Your full tests just increased the runtime of the tests by factor of nearly 10. This is out of proportion.

I had to pick /some/ count of the samples, it happened to be the current values. Maybe there should be less samples. The actual ~problem~, is that i want more granulated output from `ctest`. Right now i switched back to less granulated output, and it should be **much** faster.

Also, you do know that you can do `ctest -j9`?

> If you had simply merged that pr as is and did your split of Common.h afterwards, it would have cost you the same amount of work but spared me about 5 minutes for the manual rebase.

Sorry, i agreed to assume maintaintership with thought of *me* being able to practice C++, and improve the library, with *everything* else being secondary. So yes, i guess this is me being somewhat discouraging from active contribution, but i also see no alternatives. Either it works for me, or it does not... Just putting all the cards on the table here.

TLDR: contributing back is important, even though the license does not require it. But *I* need to be motivated to review&merge things.

axxel: > and it should be much faster.

It is indeed. thanks.

> Also, you do know that you can do ctest -j9?

No I did not. It is not that I really cared about the 30 seconds test run. I mainly wanted to point out that your testing methodology could be improved. I think a white-box testing approach is the right tool for this kind of problem and the complete set of those byte swap functions has on the order of 10 code paths (did not count). What I am ultimately trying to say is: in my personal opinion you could have spent your time and effort in this particular case to produce something with more value for the project than those bash-scripted black-box test numbers.

> Just putting all the cards on the table here.

Thanks for that. Appreciated. Just my 2 cents, though: as a respectful maintainer, maybe the good of the project could use a little bump in your priority list ;).

> But I need to be motivated to review&merge things.

Agreed. I did this hacking marathon for fun and I don't expect others to do stuff they don't like. But reviewing PRs is part of the responsibility of the role _you gave yourself_. If you don't like that just do something else. If that means not looking into code from others anymore, so be it. But if your sense of 'ownership' for the code (most of which you did not write) means forking this project at this point in time for this reason would be just sad (IMHO).

Anyway, I said I was about done with my enthusiasm working on this code and that is still true. I simply wanted to leave something 'consistent' and therefore did the Tiff interface, FILEMAP_MARGIN and DngOpcodes pr. The `Hint` class was just a little thing that seemed so obvious, I could not resist. Same goes for the Coverity fixes. I was curious if they would make sense now that you changed the compiler and while going through them one by one, why not fix them right away...

The only thing left in my patch queue at the moment is the introduction of some 'size/offset' type into the BitStream and TiffEntry interfaces. I was still waiting on some final response from you on the naming question in #76 (which really is only about "which string is nicer to look at: 'u16' or 'uint16_t'?")

TLDR; I'll let you hack on in peace soon ;).

LebedevRI:
Is this just a rewrite, or does this fix any bugs?
In other words, how do i know that it actually works?

axxel:
> Is this just a rewrite, or does this fix any bugs?

Please read the PR summary.

> In other words, how do i know that it actually works?

In #79, I mentioned a dt issue and gave an example file showing the problem. That file shows black pixels when rendered before the fix. Those are gone with this fix. There were other issues in the old code (like the clampBits one) that were undiscovered probably because no one used them. Maybe more issues show up later when samples are available that actually use all the implemented opcodes.

PR #82, opened 2017-02-10
Introduce Hints class to simplify related use cases

axxel:
There are two often repeated use cases related to accessing camera hints parsed from the xml file:

* detect if a hint is present
* get the 'value' of an (optionally) present hint and assign it

This adds a simple class `Hints` to reduce code duplication and provide a convenient access to the camera hints.

### DIFF
--- a/src/librawspeed/decoders/SrwDecoder.cpp
+++ b/src/librawspeed/decoders/SrwDecoder.cpp
@@ -52,10 +52,7 @@ RawImage SrwDecoder::decodeRawInternal() {
 
   if (32769 == compression)
   {
-    bool bit_order = false;  // Default guess
-    auto msb_hint = hints.find("msb_override");
-    if (msb_hint != hints.end())
-      bit_order = ("true" == (msb_hint->second));
+    bool bit_order = hints.get("msb_override", false);
     this->decodeUncompressed(raw, bit_order ? BitOrder_Jpeg : BitOrder_Plain);
     return mRaw;
   }
@@ -63,10 +60,7 @@ RawImage SrwDecoder::decodeRawInternal() {
   if (32770 == compression)
   {
     if (!raw->hasEntry ((TiffTag)40976)) {
-      bool bit_order = (bits == 12);  // Default guess
-      auto msb_hint = hints.find("msb_override");
-      if (msb_hint != hints.end())
-        bit_order = ("true" == (msb_hint->second));
+      bool bit_order = hints.get("msb_override", bits == 12);
       this->decodeUncompressed(raw, bit_order ? BitOrder_Jpeg : BitOrder_Plain);
       return mRaw;
     }
--- a/src/librawspeed/decoders/ArwDecoder.cpp
+++ b/src/librawspeed/decoders/ArwDecoder.cpp
@@ -74,7 +74,7 @@ RawImage ArwDecoder::decodeRawInternal() {
       return mRaw;
     }
 
-    if (hints.find("srf_format") != hints.end()) {
+    if (hints.has("srf_format")) {
       raw = mRootIFD->getIFDWithTag(IMAGEWIDTH);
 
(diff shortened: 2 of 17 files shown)


src/librawspeed/decoders/SrwDecoder.cpp
```diff
@@ -52,21 +52,15 @@ RawImage SrwDecoder::decodeRawInternal() {
...
     return mRaw;
   }
 
   if (32770 == compression)
   {
     if (!raw->hasEntry ((TiffTag)40976)) {
-      bool bit_order = (bits == 12);  // Default guess
-      auto msb_hint = hints.find("msb_override");
-      if (msb_hint != hints.end())
-        bit_order = ("true" == (msb_hint->second));
```
LebedevRI, line 69: So what happens if the hint has value of `false`?

axxel: Then `bit_order` gets set to `false`. That is what `istringstream >> boolVar;` does.

LebedevRI: And if it's something other than "true"/"false"? Do note that it is not checked anywhere that xml only has one of these 2 values.

axxel: Right now there are 3 strings used in the xml: 'true', 'false' and '' (empty). They are mapped to the same values as with the old code. It could be that the `>>` operator also maps 'True' to `true` which the current code does not. I'd say that would be an improvement, if anything.

axxel: Well, turns out, you had a point there. I went back to see what happens if you put 'True' in there and found out that the code did in fact not work as I claimed, i.e. calling it with 'true' returned `false`. I changed one line of the patch to fix it: `iss >> std::boolalpha >> defaultValue;`. The `boolalpha` modifier was missing.

It turns out, there is not a single hint in the xml that depends on evaluating 'true' to `true`. There are only 5 'true' strings and all those hints are only checked for their presence. I think this should be improved by making it more obvious: remove the 'value' property from the xml of those hints that are true iff they are present.

LebedevRI: So, i checked, and there are three hints with boolean values:
```
$ grep -i "hint " cameras.xml | sort | uniq | grep "true\|false"
                        <Hint name="msb_override" value="false"/>
                        <Hint name="nikon_wb_adjustment" value="true"/>
                        <Hint name="wb_mangle" value="true"/>
```
And apparently the `msb_override` can also have a value of `` (empty). With old code that would mean false.
```
$ grep -i "hint " cameras.xml | sort | uniq | grep "msb_override\|nikon_wb_adjustment\|wb_mangle"
                        <Hint name="msb_override" value=""/>
                        <Hint name="msb_override" value="false"/>
                        <Hint name="nikon_wb_adjustment" value="true"/>
                        <Hint name="wb_mangle" value="true"/>
```
I guess to dispel my doubts i will just have to write a test ))

LebedevRI: And i was right, this changes behavior in case of `` (empty value).
Old code evaluates to false, new code evaluates to the provided defailt.

axxel:
> Old code evaluates to false, new code evaluates to the provided defailt.

indeed. Please consider the old code buggy. Or rather the fact that someone provided a hint in the xml to override the value and then not specify a value to be used for overriding it. If there is no value given, then using the default is the reasonable choice. And if the flag is only about being there or not, then removing the empty value property from the xml is the obvious improvement to the clarity of the xml+code combination.

LebedevRI:
The pr is stated to simplify usecases, yet it silently potentially breaks the config.
Even though i have already said it is not ok.

This really makes reviewing your pr's problematic for me. I do have a problem with that.

PR #84, opened 2017-02-12
clampBits() changes

axxel:
Fix most of the remaining core library issues (excluding Test and utilities).

The fix of the `fixBadPixel` copy'n'paste bug (last commit) results in a bunch of rstest hash mismatches.

### DIFF
--- a/src/librawspeed/common/Common.h
+++ b/src/librawspeed/common/Common.h
@@ -26,6 +26,7 @@
 #include <memory>           // for unique_ptr, allocator
 #include <string>           // for string
 #include <vector>           // for vector
+#include <cassert>          // for assert
 
 int rawspeed_get_number_of_processor_cores();
 
@@ -107,10 +108,11 @@ extern "C" { FILE __iob_func[3] = { *stdin,*stdout,*stderr }; }
 #endif
 #endif
 
-// clampBits clamps the given int to the range 0 .. 2^n-1
-inline uint32 clampBits(int64 x, uint32 n) {
-  const int64 tmp = (1ULL << n) - 1ULL;
-  x = x < 0LL ? 0LL : x;
+// clampBits clamps the given int to the range 0 .. 2^n-1, with n <= 16
+inline ushort16 clampBits(int x, uint32 n) {
+  assert(n <= 16);
+  const int tmp = (1 << n) - 1;
+  x = x < 0 ? 0 : x;
   x = x > tmp ? tmp : x;
   return x;
 }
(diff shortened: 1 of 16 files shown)


src/librawspeed/common/Common.h
```diff
@@ -26,10 +26,7 @@
 #include <memory>           // for unique_ptr, allocator
 #include <string>           // for string
 #include <vector>           // for vector
-
-#ifndef __BYTE_ORDER__
-#include <cassert> // for assert
```
LebedevRI: If you don't know why something is there, maybe not change it

src/librawspeed/decoders/CrwDecoder.cpp
```diff
@@ -193,8 +193,9 @@ void CrwDecoder::decodeMetaDataInternal(CameraMetaData *meta) {
       ushort16 wb_index = shot_info->getU16(7);
       CiffEntry *wb_data = mRootIFD->getEntryRecursive(CIFF_WHITEBALANCE);
       /* CANON EOS D60, CANON EOS 10D, CANON EOS 300D */
-      int wb_offset = (wb_index < 18) ? "0134567028"[wb_index]-'0' : 0;
-      wb_offset = 1+wb_offset*4;
+      if (wb_index > 9)
+        ThrowIOE("CrwDecoder: invalid white balance index");
```
LebedevRI, line 197: how is this an input/output issue?

src/librawspeed/common/Common.h
```diff
@@ -107,10 +108,11 @@ extern "C" { FILE __iob_func[3] = { *stdin,*stdout,*stderr }; }
 #endif
 #endif
 
-// clampBits clamps the given int to the range 0 .. 2^n-1
-inline uint32 clampBits(int64 x, uint32 n) {
-  const int64 tmp = (1ULL << n) - 1ULL;
-  x = x < 0LL ? 0LL : x;
+// clampBits clamps the given int to the range 0 .. 2^n-1, with n <= 16
+inline ushort16 clampBits(int x, uint32 n) {
+  assert(n <= 16);
```
LebedevRI, line 113: So what about dngopcodes where it is probably used with wrong argument order?

LebedevRI:
Picked all but `clampBits()` commit, thanks.
I don't agree with UNINIT_CTOR fixes, they only silence the warning, the real issues are still there.

src/librawspeed/common/Common.h
axxel: I fixed those in #80.

src/librawspeed/decoders/CrwDecoder.cpp
axxel: The type of the exception might be questionable (as are _all_ exception types). The fact is, if `wb_index > 9` then the file is broken.

LebedevRI: True

PR #87, opened 2017-02-13
Remove filemap margin

axxel:
Complete another open end from my basic infrastructure cleanup work. Details see below.

### DIFF
--- a/src/librawspeed/io/Buffer.h
+++ b/src/librawspeed/io/Buffer.h
@@ -22,15 +22,19 @@
 #pragma once
 
 #include "common/Common.h" // for uchar8, uint32, uint64
+#include "io/IOException.h"// for ThrowIOE
 #include "io/Endianness.h" // for getByteSwapped
 #include <algorithm>       // for swap
 
 namespace RawSpeed {
 
-// All file maps have this much space extra at the end. This is useful for
-// BitPump* that needs to have a bit of extra space at the end to be able to
-// do more efficient reads without crashing
-#define FILEMAP_MARGIN 16UL
+// This allows to specify the nuber of bytes that each Buffer needs to
+// allocate additionally to be able to remove one runtime bounds check
+// in BitSream::fill. There are two options:
+// 0 : allocate exactly as much data as required
+// 4 : add minimum number of bytes to keep all BitStream implementations happy
+//     this disables the bounds checking, saves about 1% on modern CPUs
+#define BUFFER_PADDING 0
 
 /*************************************************************************
  * This is the buffer abstaction.
@@ -72,7 +76,12 @@ public:
   }
 
   // get pointer to memory at 'offset', make sure at least 'count' bytes are accessable
-  const uchar8* getData(size_type offset, size_type count) const;
+  const uchar8* getData(size_type offset, size_type count) const {
+    if (!isValid(offset, count))
+      ThrowIOE("Buffer overflow: image file may be truncated");
+
+    return &data[offset];
+  }
 
   // convenience getter for single bytes
   uchar8 operator[](size_type offset) const {
@@ -96,12 +105,8 @@ public:
     return size;
   }
 
-  inline bool isValid(size_type offset) const {
-    return offset < size;
-  }
-
-  inline bool isValid(size_type offset, size_type count) const {
-    return (uint64)offset + count - 1 < size;
+  inline bool isValid(size_type offset, size_type count = 1) const {
+    return (uint64)offset + count - 1 < (uint64)size + BUFFER_PADDING;
   }
 
 //  Buffer* clone();
--- a/src/librawspeed/decoders/AriDecoder.cpp
+++ b/src/librawspeed/decoders/AriDecoder.cpp
@@ -98,7 +98,6 @@ void AriDecoder::decodeThreaded(RawDecoderThread * t) {
       uint32 b = bits.getBits(12);
       dest[x*2] = b;
       dest[x*2+1] = a;
-      bits.checkPos();
     }
   }
 }
--- a/src/librawspeed/decoders/ArwDecoder.cpp
+++ b/src/librawspeed/decoders/ArwDecoder.cpp
@@ -245,7 +245,6 @@ void ArwDecoder::DecodeARW(ByteStream &input, uint32 w, uint32 h) {
   int sum = 0;
   for (int64 x = w - 1; x >= 0; x--) {
     for (uint32 y = 0; y < h + 1; y += 2) {
-      bits.checkPos();
       bits.fill();
       if (y == h) y = 1;
       uint32 len = 4 - bits.getBitsNoFill(2);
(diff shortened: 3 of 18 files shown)


src/librawspeed/io/Buffer.h
```diff
@@ -22,15 +22,19 @@
...
-// BitPump* that needs to have a bit of extra space at the end to be able to
-// do more efficient reads without crashing
-#define FILEMAP_MARGIN 16UL
+// This allows to specify the nuber of bytes that each Buffer needs to
+// allocate additionally to be able to remove one runtime bounds check
+// in BitSream::fill. There are two options:
+// 0 : allocate exactly as much data as required
+// 4 : add minimum number of bytes to keep all BitStream implementations happy
+//     this disables the bounds checking, saves about 1% on modern CPUs
+#define BUFFER_PADDING 0
```
LebedevRI, line 37: So if it is not needed, is there any reason to not completely remove it?

LebedevRI: Also, pretty sure it should be `0UL`

axxel: I explained it both in the commit message and the comment right there why I left it it, so people (like me) could use it if they'd like to achieve a performance gain. It is clearly documented and does no harm to let it be as is. As for `UL`: you may or may not add that, it has exactly no influence on the behavior of the resulting binary.

LebedevRI: Then at least do it properly please.
I.e. if `explicit Buffer(const uchar8 *data_, size_type size_)` is used, the padding is **guaranteed** to always be zero...

axxel: That is not true. The user setting that value to `4` might have allocated `4` extra bytes in `data_` just as required or he might have allocated `size_/2` bytes or any other number. The API could be changed (e.g. by having two different Buffer types) so that the chance of someone doing something wrong is decreased. I don't think that would be worth it. I fixed the issue we used to have and merely wanted to not let the knowledge that was present in the code get lost.

LebedevRI: When `buffer` is created from passed `pointer` and `size`, the `size` is the **size** of usable area of the buffer. The pointer may be pointing to bigger buffer, but only the `size` is deemed usable. Assuming anything else, i.e. that there **may** be more size that told, is asking for trouble.

axxel: This really is a non-issue: The user has to use the library in way very likely no one uses it right now (allocating the memory themself instead of using `FileReader`) _and_ has to modify the source code of the library to create a potential problem. And if he modifies the code he can introduce all kinds of issues if does not know what he is doing. This define is about as safe as some comment that explains how the code could be changed.

LebedevRI: > allocating the memory themself instead of using `FileReader`

Aha, so we *are* both talking about the same thing, good 

> ...

Other than these 3 open prs, do you have other changes to be upstreamed? yes/no

axxel: No

LibRaw: Is there any chance that RawSpeed will work with mmap-ed file? No extra bytes at end, sure.

LebedevRI: @LibRaw that is why i'm so pissed. it *MUST* work properly with any passed buffer+size pair, without making any assumptions about the actual-or-so-i-think size of the buffer behind everyone's back. i will be fixing that, and can finally start looking into fuzzing :)

@axxel OK, good. i will just merge everything then

LebedevRI (changes requested):
The code does not work, even if i simply check-out your branch.
`-DCMAKE_BUILD_TYPE=ASAN`
It starts processing `./Canon/EOS 10D/RAW_CANON_10D.CRW`, and never ends.

src/librawspeed/io/Buffer.h
axxel: > Is there any chance that RawSpeed will work with mmap-ed file? No extra bytes at end, sure.

Yes, after this is merged, that will be possible. Actually, it would have been almost certainly been possible even without this due to block/page sizes involved.

Note though: there is (disabled) code inside the `FileReader` that uses memory mapping and it was disabled (presumably) because it actually makes the code slower. That is also what my personal experience showed in similar situations.

> i will be fixing that

Is that so? This is _exactly_ what _this_ pr fixes. (My pr sets the padding to `0`).

axxel:
> The code does not work, even if i simply check-out your branch.

I'll have a look. It certainly worked the time I pushed the pr.

src/librawspeed/io/Buffer.h
LebedevRI: > Is that so? This is exactly what this pr fixes. (My pr sets the padding to 0).

The code still does assume that the buffer has `PADDING` padding size at the end, regardless of how it said buffer was acquired, yes?

axxel:
I can not reproduce your problem. It worked just fine. I rebased the pr to current develop to see if I find any issue there. I did not. I tested ASAN/RELEASE with and without pthreads.

src/librawspeed/io/Buffer.h
axxel: > The code still does assume that the buffer has PADDING padding size at the end, regardless of how it said buffer was acquired, yes?

I'm not sure I understand you question. I'll try answering it anyway: The code does assume the buffer has 0 bytes padding at the end (i.e. _no_ padding), regardless of how that buffer was acquired.

LebedevRI: > the buffer has 0 bytes padding at the end 

Where 0 is BUFFER_PADDING, correct?

LebedevRI:
It works if i do any of the following:
1. it works if i do *not* do this: `export ASAN_OPTIONS="detect_stack_use_after_return=1"`
2. it works if i change padding to `#define BUFFER_PADDING 16UL`

So the problem is very likely in the `#if BUFFER_PADDING==0 ... #endif` block.

LebedevRI:
Hmm, though it does not seem to happen with clang. I'm leaning towards it being a gcc bug.
But i'll wait a bit just in case...

axxel:
I could reproduce it now with the `detect_stack_use_after_return=1` setting. It makes sense that this check kicks in for the `uchar8 tmp[4]` variable. But there is nothing wrong with that variable, meaning it is not used after the stack frame is left.

LebedevRI:
> I could reproduce it now with the detect_stack_use_after_return=1 setting.

Good

> It makes sense that this check kicks in for the uchar8 tmp[4] variable. But there is nothing wrong with that variable, meaning it is not used after the stack frame is left.

I have to agree. And i only observe this with gcc, not clang (which had this option first)
Also, i never see the asan output, it just keeps looping forever, and bt isn't too helpful.

I guess i'll merge as-is.

PR #90, opened 2017-02-15
Fix complete failure on NO_PTHREAD library vs executable mismatch

axxel:
Unfortunately the library user has to know if the library was built with the internal 'flag' `-DNO_PTHREAD` or not and has to compile its executable with the same define. If he does not, the ABI changes and the result is complete garbage (as in rstest fails with all files). Segfaults might also be around the corner.

The problem is that the public facing `RawImageData` class changes its internal layout depending on the define. That must not happen, hence the `pthread_mutex_t` members must be removed. This pr first reduces the problem surface my going from 3 mutexes to 1 and then fixes it by using the pimpl idiom for the mutext member.

It also removes the other ABI difference between the two build options: remove the need to define the `rawspeed_get_number_of_processor_cores()` function in the client code.


LebedevRI:
I have considered my options, and decided to not handle more than one pr a week.
So i guess it will be one of these repos with hundreds of stale prs.

As for this changes, i will not be merging it.
1. i have already explicitly stated everything there is to state about c++11 threads
2. this changes api, the client code `rawspeed_get_number_of_processor_cores()` can return whatever number it wants, it is not guaranteed nor expected to be the same value you get from `std::thread::hardware_concurrency()`
3. as for abi, it's a copylib, so no abi

axxel:
> decided to not handle more than one pr a week

Good to know. Thanks for telling.

ad 1): If you see a problem even with the simple platform independent function call `hardware_concurrency()` you might as well use the OpenMP or windows specific pthread extension that was used before to get the default value.

ad 2): please note `setThreadCount()` which allows client code to achieve the exact same think as before, just without forcing them to define a 'random' external symbol in their code.

ad 3): it kind of stopped being that the moment you added a comprehensive cmake based build system (in the sense that now someone could use that build system instead of compiling the code by incorporating it in their existing project). Besides, listing that as one reason why will not merge this, you seem to argue that being able to accidentally break your build is a 'feature' of the library that should be kept?

LebedevRI:
Please start using github issues to report problems, not pull requests

axxel:
> Please start using github issues to report problems, not pull requests

Excuse me for stating the obvious: this pr does not 'report a problem', it is providing a fix for a problem.

LebedevRI:
> Unfortunately the library user has to know if the library was built with the internal 'flag' -DNO_PTHREAD or not and has to compile its executable with the same define. 

What are exact steps to reproduce the problem?

axxel:
> What are exact steps to reproduce the problem?

You can relatively easy 'simulate' the issue by setting the cmake build option to use pthreads but then add a `#define NO_PTHREAD` at the top of `rstest.cpp`. Or globally disable it and then `#undef NO_PTHREAD` in `rstest.cpp`.

I noticed when I tried to remove the code from my production build environment and use your cmake system instead.

LebedevRI:
Feel free to open new *issue* once there is an reproducer.

axxel:
> Feel free to open new issue once there is an reproducer.

You mean I should find someone else that has this problem, too? Because me having this issue is not 'good enough' a reason to fix it?

TurboGit:
No, a reproducer is a simple step by step procedure to show the problem so that anyone can see the issue, test it and check against a possible fix.

LebedevRI:
> You can relatively easy 'simulate' the issue by setting the cmake build option to use pthreads but then add a #define NO_PTHREAD at the top of rstest.cpp. Or globally disable it and then #undef NO_PTHREAD in rstest.cpp.

*That* is not a reproducer; similarly, chopping-off someone's feet to show that he can walk upside down on hands is not a solution either.

axxel:
> No, a reproducer is a simple step by step procedure to show the problem so that anyone can see the issue, test it and check against a possible fix.

I see, thanks for clarifying that. It did not occur to me that @LebedevRI meant this, since I provided a "simple step by step procedure" in the comment https://github.com/darktable-org/rawspeed/pull/90#issuecomment-280009588 just 10 minutes earlier.

> That is not a reproducer; similarly, chopping someone's feet to show that he can walk upside down on hands is not a solution either.

This analogy seems a bit far fetched. I'm sure anyone with a basic understanding of the involved technicalities could follow the reasoning of my argument and acknowledge that a user building the library with disabled pthreads and then not defining `NO_PTHREADS` in his client build will run into this issue. Since he does not even know about the existence of this define, that is actually quite likely.

axxel:
@TurboGit Since you chimed in on the matter: to give a bit of background that might help with understanding this particular 'discussion', please see https://github.com/darktable-org/rawspeed/commit/78307bcb8726a897631ce13d599365d15282e817#commitcomment-20894267.

LebedevRI:
FWIW, if rs is built with `-DWITH_PTHREADS=OFF`, rstest works just fine.
So i suspect the larger problem here is *how* you are linking to the built library.
Do you just link to the object, or `add_directory(rawspeed)` and then `target_link_libraries(binary rawspeed_static)` ? In first case, it's clearly not a bug.
TLDR: without reproducer, there is just nothing to talk about.

axxel:
> FWIW, if rs is built with -DWITH_PTHREADS=OFF, rstest works just fine.

You seemed to have missed the point. I'll try to explain it again (trying to be as brief as I can):

 1. setup your build with `-DWITH_PTHREADS=OFF` (as you did), this results in a compiler command line that includes `-DNO_PTHREAD`.
 2. add the line `#undef NO_PTHREAD` at the top of `rstest.cpp` to 'simulate' how a 'normal' client would be compiled
 3. compile and run `rstest`

LebedevRI:
Assuming you properly link to `rawspeed_static` target (or otherwise tell cmake to use `$<TARGET_PROPERTY:rawspeed_static,COMPILE_DEFINITIONS>`) when building the client code, this should now be fixed.

axxel:
Your fix to the problem requires someone to 'hook into' your cmake build definitions to compile his own software, which is about as likely as him manually defining the proper `NO_PTHREAD`. Maybe the user has no interest in pthreads support and uses some other build system for his own code (like @LibRaw).

So: no, your counter proposal is no fix for the issue that this pr already fixes.

LebedevRI:
> Maybe the user has no interest in pthreads support and uses some other build system for his own code

Both are perfectly valid use-cases.
As you said, `I noticed when I tried to remove the code from my production build environment and use your cmake system instead.`, so it worked perfectly with your build system, without the rawspeed cmake build system, and with it too, i assume.

This build system is the reference implementation. One can either use it, and then everything is guaranteed to work. Or one can NOT use it, and use something else, and then it's his problem to make sure that everything works.

I'm sorry, but i don't have to agree with each and every pullrequest made by everyone. Similarly, i don't have to unconditionally agree that everyone's pullrequest is the right and the proper fix. And lastly, i don't have to explain my decision-making logic.

Please use github issues to report issue.

axxel:
> and with it too, i assume.

No. To brake it down even more, here is the use case I'm referring to:
 1. download rawspeed
 1. configure it without pthreads and build it with the supplied cmake files
 1. `#include "RawSpeed-API.h"` in your own code in your own build system
 1. be surprised why you get garbage

> I'm sorry, but i don't have to agree with each and every pullrequest made by everyone.

Ack.

> Similarly, i don't have to unconditionally agree that everyone's pullrequest is the right and the proper fix.

Ack.

>And lastly, i don't have to explain my decision-making logic.

Depends on what kind of collaboration environment you are trying to create. In the type of project I would like to contribute, you would have to.

LebedevRI:
> No. To brake it down even more, here is the use case I'm referring to:
>  1. download rawspeed
>  1. configure it without pthreads and build it with the supplied cmake files
>  1. `#include "RawSpeed-API.h"` in your own code in your own build system
>  1. be surprised why you get garbage

Since you did not link to the rawspeed, how do you expect it to work?
More seriously, *H-O-W* do you link to rawspeed?

Yes, you need to use the same defines both when compiling and when using (read: your own code) the compiled library. Failure to do so is intentional bug in user's build system.

PR #91, opened 2017-02-21
Panasonic DMC TZ61: add basic camera support

schenlap:
see darktable #11508

also TZ60, TZ61 and ZS40 seems to be the same camera there
a differences in black and white points for TZ60 (already
in cameras.xml) and TZ61
http://www.camera-usermanual.com/manuals/panasonic/Panasonic_LUMIX_DMC-TZ60.pdf


LebedevRI:
@schenlap thank you.

PR #101, opened 2017-05-20
Panasonic DC-FZ82: add basic camera support

schenlap:
see darktable #11602

original patch is corrected (black level and aliases) and checked


LebedevRI:
@schenlap thank you and the original author.

PR #107, opened 2017-06-23
Fujifilm compressed RAF support.

muessel:
Hello,

this pull request implements support for Fujifilm compressed RAF Files for the X-Pro2, X-T2 and X-T20. The code is based on LibRaw / rawtherapee.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -8598,6 +8598,19 @@
 		<Crop x="0" y="0" width="-128" height="0"/>
 		<Sensor black="1024" white="16383"/>
 	</Camera>
+	<Camera make="FUJIFILM" model="X-Pro2" mode="compressed">
+		<ID make="Fujifilm" model="X-Pro2">Fujifilm X-Pro2</ID>
+		<CFA2 width="6" height="6">
+			<ColorRow y="0">GGRGGB</ColorRow>
+			<ColorRow y="1">GGBGGR</ColorRow>
+			<ColorRow y="2">BRGRBG</ColorRow>
+			<ColorRow y="3">GGBGGR</ColorRow>
+			<ColorRow y="4">GGRGGB</ColorRow>
+			<ColorRow y="5">RBGBRG</ColorRow>
+		</CFA2>
+		<Crop x="0" y="6" width="-16" height="0"/>
+		<Sensor black="1024" white="16383"/>
+	</Camera>
 	<Camera make="FUJIFILM" model="X70">
 		<ID make="Fujifilm" model="X70">Fujifilm X70</ID>
 		<CFA2 width="6" height="6">
@@ -8827,6 +8840,19 @@
 		<Crop x="0" y="0" width="-126" height="0"/>
 		<Sensor black="1024" white="16383"/>
 	</Camera>
+	<Camera make="FUJIFILM" model="X-T2" mode="compressed">
+		<ID make="Fujifilm" model="X-T2">Fujifilm X-T2</ID>
+		<CFA2 width="6" height="6">
+			<ColorRow y="0">GGRGGB</ColorRow>
+			<ColorRow y="1">GGBGGR</ColorRow>
+			<ColorRow y="2">BRGRBG</ColorRow>
+			<ColorRow y="3">GGBGGR</ColorRow>
+			<ColorRow y="4">GGRGGB</ColorRow>
+			<ColorRow y="5">RBGBRG</ColorRow>
+		</CFA2>
+		<Crop x="0" y="6" width="-16" height="0"/>
+		<Sensor black="1024" white="16383"/>
+	</Camera>
 	<Camera make="FUJIFILM" model="X-T10">
 		<ID make="Fujifilm" model="X-T10">Fujifilm X-T10</ID>
 		<CFA2 width="6" height="6">
@@ -8851,6 +8877,19 @@
 			<ColorRow y="5">GGRGGB</ColorRow>
 		</CFA2>
 		<Crop x="0" y="0" width="-126" height="0"/>
+		<Sensor black="1024" white="16383"/>
+	</Camera>
+	<Camera make="FUJIFILM" model="X-T20" mode="compressed">
+		<ID make="Fujifilm" model="X-T20">Fujifilm X-T20</ID>
+		<CFA2 width="6" height="6">
+			<ColorRow y="0">GGRGGB</ColorRow>
+			<ColorRow y="1">GGBGGR</ColorRow>
+			<ColorRow y="2">BRGRBG</ColorRow>
+			<ColorRow y="3">GGBGGR</ColorRow>
+			<ColorRow y="4">GGRGGB</ColorRow>
+			<ColorRow y="5">RBGBRG</ColorRow>
+		</CFA2>
+		<Crop x="0" y="6" width="-16" height="0"/>
 		<Sensor black="1024" white="16383"/>
 	</Camera>
 	<Camera make="KONICA MINOLTA" model="DYNAX 5D">
--- a/src/librawspeed/decoders/RafDecoder.cpp
+++ b/src/librawspeed/decoders/RafDecoder.cpp
@@ -172,7 +172,7 @@ void RafDecoder::checkSupportInternal(const CameraMetaData* meta) {
     auto id = mRootIFD->getID();
     const Camera* cam = meta->getCamera(id.make, id.model, mRaw->metadata.mode);
     if (!cam)
-      ThrowRDE("Couldn't find camera %s %s", id.make, id.model);
+      ThrowRDE("Couldn't find camera %s %s", id.make.c_str(), id.model.c_str());
 
     mRaw->cfa = cam->cfa;
   }
--- a/src/librawspeed/decoders/RafDecoder.cpp
+++ b/src/librawspeed/decoders/RafDecoder.cpp
@@ -24,6 +24,7 @@
 #include "common/Point.h"                           // for iPoint2D, iRecta...
 #include "decoders/RawDecoderException.h"           // for RawDecoderExcept...
 #include "decompressors/UncompressedDecompressor.h" // for UncompressedDeco...
+#include "decompressors/FujiDecompressor.h"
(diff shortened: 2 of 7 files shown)


LebedevRI:
Hi.
Nice to finally see that.
Are there *any* changes other than adapting to some of rs's infrastructure?

src/librawspeed/decompressors/FujiDecompressor.cpp
```diff
@@ -0,0 +1,1097 @@
...
+    FujiDecompressor - Decompress Fujifilm compressed RAF.
+
+    Copyright (C) 2016 Alexey Danilchenko
+    Copyright (C) 2016 Alex Tutubalin
+    Copyright (C) 2017 Uwe Müssel
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2 of the License, or (at your option) any later version.
```
LebedevRI:
```diff
+    License as published by the Free Software Foundation; either
+    version 2 of the License, or (at your option) any later version.
```
Hmm, where did you get that?
https://github.com/LibRaw/LibRaw/commit/aeab4d8ea58f85050d1253bdc41f51baa79ffe65#diff-03b018f1deeb8fc38d2c927f09465635R11 says:
```diff
+LibRaw is free software; you can redistribute it and/or modify
+it under the terms of the one of three licenses as you choose:
+
+1. GNU LESSER GENERAL PUBLIC LICENSE version 2.1
+   (See file LICENSE.LGPL provided in LibRaw distribution archive for details).
+
+2. COMMON DEVELOPMENT AND DISTRIBUTION LICENSE (CDDL) Version 1.0
+   (See file LICENSE.CDDL provided in LibRaw distribution archive for details).
+
```
(2 vs 2.1)

data/cameras.xml
```diff
@@ -8598,6 +8598,19 @@
...
+		<ID make="Fujifilm" model="X-Pro2">Fujifilm X-Pro2</ID>
+		<CFA2 width="6" height="6">
+			<ColorRow y="0">GGRGGB</ColorRow>
+			<ColorRow y="1">GGBGGR</ColorRow>
+			<ColorRow y="2">BRGRBG</ColorRow>
+			<ColorRow y="3">GGBGGR</ColorRow>
+			<ColorRow y="4">GGRGGB</ColorRow>
+			<ColorRow y="5">RBGBRG</ColorRow>
+		</CFA2>
+		<Crop x="0" y="6" width="-16" height="0"/>
```
LebedevRI, line 8611: Ugh, so they are even writing different raw image WxH in compressed vs. uncompressed formats..

src/librawspeed/decompressors/FujiDecompressor.cpp
```diff
@@ -0,0 +1,1097 @@
...
+  }
+
+  *data <<= bits_left;
+  bits_left_in_byte -= bits_left;
+  *data |= ((1 << bits_left) - 1) &
+           ((unsigned)info->cur_buf[info->cur_pos] >> bits_left_in_byte);
+  info->cur_bit = (8 - (bits_left_in_byte & 7)) & 7;
+}
+
+int FujiDecompressor::bitDiff(int value1, int value2) {
```
LebedevRI:
```
int __attribute__(const) FujiDecompressor::bitDiff(int value1, int value2) {
```

src/librawspeed/decoders/RafDecoder.cpp
```diff
@@ -155,6 +165,17 @@ RawImage RafDecoder::decodeRawInternal() {
...
   if (!this->checkCameraSupported(meta, mRootIFD->getID(), ""))
     ThrowRDE("Unknown camera. Will not guess.");
+
+  if (isCompressed()) {
+    mRaw->metadata.mode = "compressed";
+
+    auto id = mRootIFD->getID();
+    const Camera* cam = meta->getCamera(id.make, id.model, mRaw->metadata.mode);
+    if (!cam)
+      ThrowRDE("Couldn't find camera %s %s", id.make, id.model);
```
LebedevRI:
```
ThrowRDE("Couldn't find camera %s %s", id.make.c_str(), id.model.c_str());
```

src/librawspeed/decompressors/FujiDecompressor.h
```diff
@@ -0,0 +1,145 @@
...
+#include "LJpegDecompressor.h"
+
+namespace rawspeed {
+
+class FujiDecompressor //: public LJpegDecompressor
+{
+public:
+  FujiDecompressor(Buffer input, RawImage img, int offset);
+  ~FujiDecompressor(void);
+
```
LebedevRI, line 34: i guess you want to add
```
using ushort = ushort16;
```
right here for now.

LebedevRI (changes requested):
Please make CI happy :)
The code is far from ideal, which is quite understandable, but it will be easier for me to fix it after merging as-is...

src/librawspeed/decompressors/FujiDecompressor.cpp
muessel: Is it enough to change the version from 2 to 2.1 or do I have to copy the whole LibRaw license header?

data/cameras.xml
muessel: Yes, uncomressed and comressed files are totally different, eg.:

exiftool Fujifilm\ -\ X-T20\ -\ compressed\ \(3-2\).RAF
...
Raw Image Width                 : 6032
Raw Image Height                : 4032
Raw Image Full Width            : 6048
Raw Image Full Height           : 4038

exiftool Fujifilm\ -\ X-T20\ -\ Uncompressed\ \(3-2\).RAF
...
Raw Image Width                 : 6032
Raw Image Height                : 4032
Raw Image Full Width            : 6160
Raw Image Full Height           : 4032

LebedevRI: Yes, that is what i was talking about.

src/librawspeed/decompressors/FujiDecompressor.cpp
LebedevRI: I think it is enough just to change the version.

LebedevRI:
Ok, i see.
Ignore those windows clang-4.0 warnings, i'll handle them manually.

muessel:
Travis CI has problems with mac builds with std::abs. Do I have to change them to abs and then include "using std::abs;" ?

LebedevRI:
Will fix locally, thank you!

data/cameras.xml
LebedevRI: @muessel 
An entry for FUJIFILM X100F is missing.
Is there some problem with that camera, or it is a simple oversight?

muessel:
Hello Roman,

I just forgot the the X100F also supports the compressed format.

I've seen you also implemented support for the GFX 50S. In the method
FujiDecompressor::copy_line_to_bayer
a static CFA is used.  Maybe this should be converted into something
configurable like with the X-Trans sensors.

Regards
   Uwe

2017-06-24 13:22 GMT+02:00 Roman Lebedev <notifications@github.com>:

LebedevRI:
> Hello Roman,
> I just forgot the the X100F also supports the compressed format.

Yeah, i checked, and X100F compressed works too, fixed in branch.
Also, while there, FUJIFILM GFX 50S uncompressed+compressed support added - the only fuji camera with bayer sensor and with this compresson scheme, i believe.

What i'm currently missing is some compressed .RAF with 12-bits data, so this branch is taken:
https://github.com/darktable-org/rawspeed/blob/fae7f4014024f7aeb34fcab0c2d7967377764ca6/src/librawspeed/decompressors/FujiDecompressor.cpp#L106-L110
I think, it is produced in high-speed continuous shooting mode, 14/12/8/whaever fps.

LebedevRI:
> I've seen you also implemented support for the GFX 50S. In the method
FujiDecompressor::copy_line_to_bayer
a static CFA is used.  Maybe this should be converted into something
configurable like with the X-Trans sensors.

All that is very strange. If that is the actual code used in current libraw, then it may mean two things (both can be valid at the same time):
1. All cameras which support compressed raw, have the exact same CFA filter pattern
2. The CFA used in compression is static/the same for all cameras/does not necessarily match the actual camera's CFA pattern.

muessel:
>
> What i'm currently missing is some compressed .RAF with 12-bits data, so
> this branch is taken:
> https://github.com/darktable-org/rawspeed/blob/
> fae7f4014024f7aeb34fcab0c2d7967377764ca6/src/librawspeed/decompressors/
> FujiDecompressor.cpp#L106-L110
> I think, it is produced in high-speed continuous shooting mode,
> 14/12/8/whaever fps.
>
I've tested it on my X-T2 with 8fps mechanical shutter and 11/14 fps
electronical shutter. The .RAFs are all 14bit according to the code
mentioned above.

LebedevRI:
https://discuss.pixls.us/t/compressed-raf-help-needed/4527

Floessie:
@muessel 
> Travis CI has problems with mac builds with std::abs. Do I have to change them to abs and then include "using std::abs;" ?

Uwe, just my 2¢ (I know Roman will fix or has already fixed it): Most often the `abs()` problem isn't about the namespace, but a change from the way this was defined for C++03 vs. C++11. There's an integer version in [`<cstdlib>`](http://en.cppreference.com/w/cpp/numeric/math/abs) and a float version in [`<cmath>`](http://en.cppreference.com/w/cpp/numeric/math/fabs). As the latter link notes:

> Between C++11 and C++14, the standard erroneously required std::abs to have overloads for integer types returning double. This requirement was removed in C++17 by defect report 2735.

That causes trouble again and again, and that's most probably the case here also. The best solution usually is [to use `std::fabs()` for floats and to make sure, you don't feed `std::abs()` with an `unsigned`]( https://github.com/Beep6581/RawTherapee/issues/3339#issuecomment-268975964).

HTH,
Flössie

LebedevRI:
Update: still no sign of that elusive 12bit sample, and no reply from @LibRaw.
I'll wait a few more days and proceed without it...

PR #108, opened 2017-07-18
Better path definition for Pugixml.

mcourteaux:
This allows rawspeed to be used as git submodule, in combination with the Download pugixml option.

### DIFF
--- a/cmake/Modules/Pugixml.cmake
+++ b/cmake/Modules/Pugixml.cmake
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.0)
 project(pugixml NONE)
 
 # Download and unpack pugixml at configure time
-configure_file(${CMAKE_SOURCE_DIR}/cmake/Modules/Pugixml.cmake.in ${CMAKE_BINARY_DIR}/pugixml/CMakeLists.txt @ONLY)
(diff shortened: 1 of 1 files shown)


LebedevRI (changes requested):
Hmm, never seen `CMAKE_CURRENT_LIST_DIR` before.
Will `CMAKE_CURRENT_SOURCE_DIR` work for you?

mcourteaux:
No, that looks for `rawspeed/src/Pugixml.cmake.in` (thus, `$CMAKE_CURRENT_SOURCE_DIR` seems to be `rawspeed/src` at that time, where `rawspeed` is the root of the repo).
https://cmake.org/cmake/help/v3.0/variable/CMAKE_CURRENT_LIST_DIR.html

As the file we are looking for is next to the current CMake file, my proposal makes sense.

LebedevRI (approved):
Hmm, after actually checking with darktable, i guess it indeed makes sense.

mcourteaux:
It probably made sense in the context of darktable indeed. But this is simply more portable. :) Thanks for your help!

LebedevRI:
Most of these cmake knobs are there for me for standalone rawspeed development, not for usage as part of some other build, thus they clearly break when used like that.
It is certainly better to have all the deps properly installed, but hey, if it works...

Thank you for your contribution!

LebedevRI:
@mcourteaux hopefully it still works after the follow-up build fixes..

mcourteaux:
Yep still works!

PR #117, opened 2017-11-29
[RFC] Skeleton of Phase One Sensor Corrections

bieber:
I'd like to get dcraw's IIQ sensor corrections into RawSpeed, so I started working just on reading the data today.  I don't expect this to be accepted, I just wanted to show what I'm thinking of doing before I get too far and see if there's anything objectionable about the approach I'm taking or the file formatting.  I've never touched RawSpeed before, so apologies if I've made a bunch of mistakes

Commit Message Below:
----------------------------------------------
dcraw has a function called phase_one_correct, which applies some
corrections to the raw data based on metadata embedded in the RAW
file.  I would like to implement any of these that I can find RAW
files serving as examples for, as they can make a significant
difference on the output (the quadrant multipliers, in particular, can
cause visible seams to appear in the rendered image if not correctly
applied).  This commit just adds some code to read the extended
metadata, with the skeleton of a function for one particular type of
correction and room to add more.

### DIFF
--- a/src/librawspeed/decoders/IiqDecoder.cpp
+++ b/src/librawspeed/decoders/IiqDecoder.cpp
@@ -127,10 +127,13 @@ RawImage IiqDecoder::decodeRawInternal() {
 
   uint32 width = 0;
   uint32 height = 0;
+  uint32 split_row = 0;
+  uint32 split_col = 0;
 
   Buffer raw_data;
   ByteStream block_offsets;
   ByteStream wb;
+  ByteStream correction_meta_data;
 
   for (uint32 entry = 0; entry < entries_count; entry++) {
     const uint32 tag = es.getU32();
@@ -151,12 +154,21 @@ RawImage IiqDecoder::decodeRawInternal() {
     case 0x10f:
       raw_data = bs.getSubView(data, len);
       break;
+    case 0x110:
+      correction_meta_data = bs.getSubStream(data);
+      break;
     case 0x21c:
       // they are not guaranteed to be sequential!
       block_offsets = bs.getSubStream(data, len);
       break;
     case 0x21d:
       black_level = data >> 2;
+      break;
+    case 0x222:
+      split_col = data;
+      break;
+    case 0x224:
+      split_row = data;
       break;
     default:
       // FIXME: is there a "block_sizes" entry?
@@ -187,6 +199,9 @@ RawImage IiqDecoder::decodeRawInternal() {
   mRaw->createData();
 
   DecodePhaseOneC(strips, width, height);
+  if (correction_meta_data.getSize() != 0) {
+    CorrectPhaseOneC(correction_meta_data, split_row, split_col);
+  }
 
   for (int i = 0; i < 3; i++)
     mRaw->metadata.wbCoeffs[i] = wb.getFloat();
@@ -240,6 +255,47 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
                                  uint32 width, uint32 height) {
   for (const auto& strip : strips)
     DecodeStrip(strip, width, height);
+}
+
+void IiqDecoder::CorrectPhaseOneC(ByteStream meta_data, uint32 split_row,
+                                  uint32 split_col) {
+  ByteStream meta_data_base(meta_data.getSubStream(0));
+
+  meta_data.skipBytes(8);
+  const uint32 bytes_to_entries = meta_data.getU32();
+  meta_data.skipBytes(bytes_to_entries - 12);
+  const uint32 entries_count = meta_data.getU32();
+  meta_data.skipBytes(4);
+
+  // this is how much is to be read for all the entries
+  ByteStream entries(meta_data.getStream(16 * entries_count));
+
+  bool quadrant_multiplier_applied = false;
+
+  for (uint32 entry = 0; entry < entries_count; entry++) {
+    const uint32 tag = entries.getU32();
+    const uint32 len = entries.getU32();
+    const uint32 data = entries.getU32();
+
+    switch (tag) {
+    case 0x431:
+      if (!quadrant_multiplier_applied) {
+        CorrectQuadrantMultipliersCombined(
+            meta_data_base.getSubStream(data, len), split_row, split_col);
+      }
+      quadrant_multiplier_applied = true;
+      break;
+    default:
+      break;
+    }
+  }
+}
+
+void IiqDecoder::CorrectQuadrantMultipliersCombined(ByteStream data,
+                                                    uint32 split_row,
+                                                    uint32 split_col) {
+  // TODO: Implementation
+
 }
 
 void IiqDecoder::checkSupportInternal(const CameraMetaData* meta) {
--- a/src/librawspeed/decoders/IiqDecoder.h
+++ b/src/librawspeed/decoders/IiqDecoder.h
@@ -73,6 +73,10 @@ protected:
   uint32 black_level = 0;
   void DecodePhaseOneC(const std::vector<IiqStrip>& strips, uint32 width,
                        uint32 height);
+  void CorrectPhaseOneC(ByteStream meta_data, uint32 split_row,
+                        uint32 split_col);
+  void CorrectQuadrantMultipliersCombined(ByteStream data, uint32 split_row,
+                                          uint32 split_col);
 };
 
 } // namespace rawspeed


src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +257,47 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+  ByteStream meta_data_base(meta_data.getSubStream(0));
+
+  meta_data.skipBytes(8);
+  const uint32 bytes_to_entries = meta_data.getU32();
+  meta_data.skipBytes(bytes_to_entries - 12);
+  const uint32 entries_count = meta_data.getU32();
+  meta_data.skipBytes(4);
+
+  // this is how much is to be read for all the entries
+  ByteStream entries(meta_data.getStream(16 * entries_count));
```
LebedevRI, line 271:
```diff
- ByteStream entries(meta_data.getStream(16 * entries_count));
+ ByteStream entries(meta_data.getStream(entries_count, 16));
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +257,47 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+
+  bool quadrant_multiplier_applied = false;
+
```
LebedevRI, line 274: You don't actually use `meta_data` below, so
```diff
+ meta_data.setPos(0);
```
```
s/meta_data_base/meta_data/
```
```diff
- ByteStream meta_data_base(meta_data.getSubStream(0));
- 
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +257,47 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
     DecodeStrip(strip, width, height);
 }
 
+void IiqDecoder::CorrectPhaseOneC(ByteStream meta_data, uint32 split_row,
+                                  uint32 split_col) {
+  ByteStream meta_data_base(meta_data.getSubStream(0));
+
+  meta_data.skipBytes(8);
+  const uint32 bytes_to_entries = meta_data.getU32();
+  meta_data.skipBytes(bytes_to_entries - 12);
```
LebedevRI, line 266: And if `bytes_to_entries` is less than `12`, you've got an unsigned overflow.
`12` is `8 + (32/8)`, i.e. `meta_data.getPos()`, so just `meta_data.setPos(bytes_to_entries);`

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +257,47 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+
+  // this is how much is to be read for all the entries
+  ByteStream entries(meta_data.getStream(16 * entries_count));
+
+  bool quadrant_multiplier_applied = false;
+
+  for (uint32 entry = 0; entry < entries_count; entry++) {
+    const uint32 tag = entries.getU32();
+    const uint32 len = entries.getU32();
+    const uint32 data = entries.getU32();
```
LebedevRI, line 278: `s/data/offset/`

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -187,6 +199,9 @@ RawImage IiqDecoder::decodeRawInternal() {
   mRaw->createData();
```
LebedevRI, line 200:
```diff
// FIXME: could be wrong. max "active pixels" in "Sensor+" mode - "101 MP"
if (width == 0 || height == 0 || width > 11608 || height > 8708)
  ThrowRDE("Unexpected image dimensions found: (%u; %u)", width, height);

+if (split_col > width || split_row > height)
+  ThrowRDE("Invalid sensor quadrants: (%u; %u)", split_col, split_row);
```
or something like that

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -151,13 +154,22 @@ RawImage IiqDecoder::decodeRawInternal() {
     case 0x10f:
       raw_data = bs.getSubView(data, len);
       break;
+    case 0x110:
+      correction_meta_data = bs.getSubStream(data);
```
LebedevRI, line 158: i *strongly* suspect
```diff
- correction_meta_data = bs.getSubStream(data);
+ correction_meta_data = bs.getSubStream(data, len);
```

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: Also, what happens if
* `correction_meta_data` is there, but not `split_col` and not `split_row`?
  Does that mean it is all one quadrant?
* `correction_meta_data` is not there, but `split_col` and/or `split_row` is there?

I.e. needs more sanitization.

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -187,6 +199,9 @@ RawImage IiqDecoder::decodeRawInternal() {
   mRaw->createData();
 
   DecodePhaseOneC(strips, width, height);
+  if (correction_meta_data.getSize() != 0) {
+    CorrectPhaseOneC(correction_meta_data, split_row, split_col);
```
LebedevRI, line 203: Elide braces, at least now they are not needed.

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: You read only 3x `U32` per entry there, why `16`?

LebedevRI (commented):
Thanks for looking into this! It should be pretty self-contained, so i think i will be ok with it.

> file formatting

https://github.com/darktable-org/darktable/blob/master/CONTRIBUTING.md#coding-style
Just install `clang-format` pre-commit hook, re-commit this patch (i.e. reformat the diff) and forget about it. :)
(fineprint: if it is not formatted with clang-format, it is no-go.)

> I can find RAW files serving as examples for

If you still have the `Credo 40`, please contribute the full set: (camera should be stable for the two shots, and the scene needs to be static)
```
7.6 File Format
A Leaf Credo back can save files in two different Raw formats;
IIQ L and IIQ S. IIQ stands for Intelligent Image Quality Raw.
IIQ L is the default setting and it is a lossless capture format.
IIQ S is a smaller file that is not a totally lossless format. 
```

That being said, we have far too few IIQ samples on RPU. And if there are no samples that trigger *all* the code that will be added by this PR, i will be *very* reluctant to merge it.

LibRaw:
Small note:

have not read patch source yet.... but please make this additional correction step optional (by setting some flag).

LibRaw may use RawSpeed as decoder (only), but this phase_one_correct..() step is performed later in LibRaw, using own code (derived from dcraw).

We still use old (Klaus's) RawSpeed, but hope to switch to this new-beautiful-one in foreseeable future

LebedevRI:
While not talking about modularity/configurability

> LibRaw may use RawSpeed as decoder (only), but this phase_one_correct..() step is performed later in LibRaw, using own code (derived from dcraw).

Of course it is, because that code is not currently implemented in rawspeed.
Some context/underlying reasoning is missing/elided in your comment.

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: Also, either https://github.com/darktable-org/rawspeed/pull/117/files#r153709175, or this is missing `entries.skipBytes(4);`

LibRaw:
reasoning is simple: LibRaw::unpack() provides unaltered data (no black subtration, no cropping, no aspect ratio stretch, no phase-one correction, no zero pixels wipe out for panasonics, etc) and it is critical  to keep this semantics even if PhaseOne data was unpacked using RawSpeed(/new) decoder.

If raw data was altered by some additional code (e.g. PhaseOne flat field corrector), all software that uses LibRaw as decoder only (own raw postprocessing) should be altered (to not do same correction again)

src/librawspeed/decoders/IiqDecoder.cpp
bieber: I got errors when I tried that, but that might have been another issue so I'll take another look at it.  I know the length of the new metadata section is encoded in the section itself though

src/librawspeed/decoders/IiqDecoder.cpp
bieber: I would imagine it's possible for the one to be present without the other, because there's a range of sensor corrections and only some of them rely on a split column/row.  Some of them could even apply to images with more than just four segments, in which case split_col and split_row would be nonsensical, and the camera might include them just for informational purposes even if there's no corrections needed (unfortunately I don't have any cameras without corrections to test this), so I don't think there's anything I can really infer from the presence/absence of one or the other

src/librawspeed/decoders/IiqDecoder.cpp
bieber: Copied from the previous metadata reading, this should be 12

src/librawspeed/decoders/IiqDecoder.cpp
bieber: Yeah, I goofed the size calculation

bieber:
> Just install clang-format pre-commit hook, re-commit this patch (i.e. reformat the diff) and forget about it. :)

Will do

> If you still have the Credo 40, please contribute the full set: (camera should be stable for the two shots, and the scene needs to be static)

Oh yeah, I forgot about small-RAW.  I'll submit both in the next day or two.  I've got this thing permanently now, hence the interest in getting this quadrant correction right ;)

> And if there are no samples that trigger all the code that will be added by this PR, i will be very reluctant to merge it.

That makes sense.  Right now I'm only really looking at the decisive correction (the quadrant multipliers) that will make the camera basically usable in DT.  From there I'll probably progress to the other corrections present in my own RAW files (I've got some bad pixel corrections and flat fields in there), and then ones I can find in other IIQ files in the sample set.  I'm not looking to implement anything I can't actually test as I go along, though.

> reasoning is simple: LibRaw::unpack() provides unaltered data (no black subtration, no cropping, no aspect ratio stretch, no phase-one correction, no zero pixels wipe out for panasonics, etc) and it is critical to keep this semantics even if PhaseOne data was unpacked using RawSpeed(/new) decoder.

> If raw data was altered by some additional code (e.g. PhaseOne flat field corrector), all software that uses LibRaw as decoder only (own raw postprocessing) should be altered (to not do same correction again)

I'll defer to @LebedevRI on this, but my own opinion is that this belongs in the decoding library.  I'd argue that this is no more post-processing than decompression is--the post-correction bytes might not be the ones that are literally written in the file, but they are nonetheless the "correct" ones to feed into demosaicing and whatever else lives downstream.  It would seem odd to me to implement RAW decoding by default such that it's impossible for any downstream software to get a correct image without doing special, vendor-specific correction to the data it gets from the decoder.

But I'm not super familiar with this, so I'll just follow what the two of you have to say.  In the event that we do end up making this optional, do you have any example of an optional feature flag in other decoders I could pattern it after?

LebedevRI:
> Will do

Great :)

> Oh yeah, I forgot about small-RAW. I'll submit both in the next day or two.

Awesome!

> I've got this thing permanently now, hence the interest in getting this quadrant correction right ;)

Ok then :)

> I've got some bad pixel corrections

Ok, that is still in rawspeed's scope, should be also easy-ish.

> and flat fields in there

Not in the scope i think. I guess you'll want to come with 
a) either converter of the parameters to something that is understood by lensfun
or 
b) some generic flatfield module for dt.

LibRaw:
> but my own opinion is that this belongs in the decoding library. I'd argue that this is no more post-processing than decompression is--the post-correction bytes might not be the ones that are literally written in the file, but they are nonetheless the "correct" ones to feed into demosaicing and whatever else lives downstream. It would seem odd to me to implement RAW decoding by default such that it's impossible for any downstream software to get a correct image without doing special, vendor-specific correction to the data it gets from the decoder.

Yes, LibRaw already has this call (derived from dcraw.c), so any software that uses LibRaw may just call this function. Also, it is called in LibRaw if LibRaw's postprocessing is used.

But this processing is very long (comparable to decoding itself), so if one need fastest possible processing (e.g. for generating very low-res thumbnails, so P1 'squares' are not visible) it should be possible to skip this stage.

Also, life not ends with end-user raw processing packages. Our RawDigger has special option ('turn off phase one flat field') which is very useful for some (rare) cases.

LebedevRI:
> But this processing is very long (comparable to decoding itself), so if one need fastest possible processing (e.g. for generating very low-res thumbnails, so P1 'squares' are not visible) it should be possible to skip this stage.

I did not look at IIQ decoder in details, but i'm 99.99% sure the normal decoding can be parallelized with zero effort, and this post-processing *should* also be parallelizable, if it is truly just a per-quadrant "scaling" for each pixel.

> Also, life not ends with end-user raw processing packages. Our RawDigger has special option ('turn off phase one flat field') which is very useful for some (rare) cases.

Finally, you have spilled the beans.
Just start with the real ugly truth and not general marketing blurb next time, k? :)

LibRaw:
Thumbnail generation is parallelizable by nature (if we have enough files), but lot of processing steps can be skipped. Including P1 flat field.

> Finally, you have spilled the beans.
Just start with the real ugly truth and not general marketing blurb next time, k? :)

For RawDigger solution is simple: just turn off RawSpeed if this option checked. We already do that for 'Linear raw curve'.

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: I guess https://github.com/darktable-org/rawspeed/pull/117#discussion_r153820014 will resolve those issues.

bieber:
Closing this one now that I have a real PR up

LebedevRI:
You are aware that you don't actually have to open the new pr?
You could just [force]push the commits to the old branch.

bieber:
Sorry, my bad.  I forgot to put these changes on their own branch in my repo

PR #118, opened 2017-12-02
Phase one correction

bieber:
This is a complete implementation of the combined quadrant correction for phase one RAW files.  Below I'm attaching before/after 100% crops of sky rendering before and after this patch.  I *think* this is ready to merge, but so far I've only tested it on Credo 40 files.  I'm out of time for tonight so I want to get this up for comments, but please don't merge it until I've gotten a chance to test it on all the available IIQ samples (I'll also add an IIQ-S sample for the Credo 40 tomorrow).

comparison

### DIFF
--- a/src/librawspeed/common/CMakeLists.txt
+++ b/src/librawspeed/common/CMakeLists.txt
@@ -17,6 +17,7 @@ FILE(GLOB SOURCES
   "RawImageDataFloat.cpp"
   "RawImageDataU16.cpp"
   "RawspeedException.h"
+  "Spline.h"
   "TableLookUp.cpp"
   "TableLookUp.h"
   "Threading.h"
--- a/src/librawspeed/common/Spline.h
+++ b/src/librawspeed/common/Spline.h
@@ -1,0 +1,95 @@
+/*
+    RawSpeed - RAW file decoder.
+
+    Copyright (C) 2017 Robert Bieber
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2 of the License, or (at your option) any later version.
+
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public
+    License along with this library; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+*/
+
+#pragma once
+
+#include "common/Common.h" // for ushort16
+#include "common/Point.h"  // for iPoint2D
+#include <vector>          // for vector
+
+namespace rawspeed {
+
+class Spline final {
+public:
+  static std::vector<ushort16>
+  calculateCurve(const std::vector<iPoint2D>& control_points) {
+    const int num_coords = control_points.size();
+    const int num_segments = num_coords - 1;
+
+    // These are the constant factors for each segment of the curve.
+    // Each segment i will have the formula:
+    // f(x) = a[i] + b[i]*(x - x[i]) + c[i]*(x - x[i])^2 + d[i]*(x - x[i])^3
+    std::vector<double> a(num_coords);
+    std::vector<double> b(num_segments);
+    std::vector<double> c(num_coords);
+    std::vector<double> d(num_segments);
+
+    // Extra values used during computation
+    std::vector<double> h(num_segments);
+    std::vector<double> alpha(num_segments);
+    std::vector<double> l(num_coords);
+    std::vector<double> mu(num_coords);
+    std::vector<double> z(num_coords);
+
+    for (int i = 0; i < num_coords; i++)
+      a[i] = control_points[i].y;
+
+    for (int i = 0; i < num_segments; i++)
+      h[i] = control_points[i + 1].x - control_points[i].x;
+
+    for (int i = 1; i < num_segments; i++)
+      alpha[i] =
+          (3. / h[i]) * (a[i + 1] - a[i]) - (3. / h[i - 1]) * (a[i] - a[i - 1]);
+
+    l[0] = mu[0] = z[0] = 0;
+
+    for (int i = 1; i < num_segments; i++) {
+      l[i] = 2 * (control_points[i + 1].x - control_points[i - 1].x) -
+             (h[i - 1] * mu[i - 1]);
+      mu[i] = h[i] / l[i];
+      z[i] = (alpha[i] - h[i - 1] * z[i - 1]) / l[i];
+    }
+
+    l[num_segments] = 1;
+    z[num_segments] = c[num_segments] = 0;
+
+    for (int i = num_segments - 1; i >= 0; i--) {
+      c[i] = z[i] - mu[i] * c[i + 1];
+      b[i] = (a[i + 1] - a[i]) / h[i] - h[i] * (c[i + 1] + 2 * c[i]) / 3.;
+      d[i] = (c[i + 1] - c[i]) / (3. * h[i]);
+    }
+
+    std::vector<ushort16> curve(65536);
+
+    for (int i = 0; i < num_segments; i++) {
+      for (int x = control_points[i].x; x <= control_points[i + 1].x; x++) {
+        double diff = x - control_points[i].x;
+        double diff_2 = diff * diff;
+        double diff_3 = diff * diff * diff;
+
+        curve[x] = a[i] + b[i] * diff + c[i] * diff_2 + d[i] * diff_3;
+      }
+    }
+
+    return curve;
+  }
+};
+
+} // namespace rawspeed
--- a/src/librawspeed/decoders/IiqDecoder.cpp
+++ b/src/librawspeed/decoders/IiqDecoder.cpp
@@ -23,6 +23,7 @@
 #include "decoders/IiqDecoder.h"
 #include "common/Common.h"                // for uint32, int32, ushort16
 #include "common/Point.h"                 // for iPoint2D
+#include "common/Spline.h"                // for calculateCurve
 #include "decoders/RawDecoderException.h" // for ThrowRDE
 #include "io/BitPumpMSB32.h"              // for BitPumpMSB32
 #include "io/Buffer.h"                    // for Buffer, DataBuffer
@@ -288,11 +289,71 @@ void IiqDecoder::CorrectPhaseOneC(ByteStream meta_data, uint32 split_row,
   }
 }
 
+// This method defines a correction that compensates for the fact that
+// IIQ files may come from a camera with multiple (four, in this case)
+// sensors combined into a single "sensor."  Because the different
+// sensors may have slightly different responses, we need to multiply
+// the pixels in each by a correction factor to ensure that they blend
+// together smoothly.  The correction factor is not a single
+// multiplier, but a curve defined by seven control points.  Each
+// curve's control points share the same seven X-coordinates.
 void IiqDecoder::CorrectQuadrantMultipliersCombined(ByteStream data,
                                                     uint32 split_row,
                                                     uint32 split_col) {
-  // TODO: Implementation
 
+  // The curves should all include (0, 0) and (65535, 65535), so the
+  // first and last points are predefined and the middle seven are
+  // read from the file
+  std::array<uint32, 9> shared_x_coords;
+  shared_x_coords.fill(0);
+  shared_x_coords.back() = 65535;
+  std::generate_n(std::next(shared_x_coords.begin()), 7,
+                  [&data] { return data.getU32(); });
+
+  std::array<std::array<std::vector<iPoint2D>, 2>, 2> control_points;
+  for (auto& quadRow : control_points) {
+    for (auto& quadrant : quadRow) {
+      quadrant.emplace_back(0, 0);
+
+      for (int i = 1; i < 8; i++) {
+        // These multipliers are expressed in ten-thousandths in the
+        // file
+        const int y_coord = (data.getU32() * shared_x_coords[i]) / 10000;
+        quadrant.emplace_back(shared_x_coords[i], y_coord);
+      }
+
+      quadrant.emplace_back(65535, 65535);
+      assert(quadrant.size() == 9);
+    }
+  }
+
+  for (int quadRow = 0; quadRow < 2; quadRow++) {
+    for (int quadCol = 0; quadCol < 2; quadCol++) {
+      const std::vector<ushort16> curve =
+          Spline::calculateCurve(control_points[quadRow][quadCol]);
+
+      int row_start = quadRow == 0 ? 0 : split_row;
+      int row_end = quadRow == 0 ? split_row : mRaw->dim.y;
+      int col_start = quadCol == 0 ? 0 : split_col;
+      int col_end = quadCol == 0 ? split_col : mRaw->dim.x;
+
+      for (int row = row_start; row < row_end; row++) {
+        ushort16* pixel =
+            reinterpret_cast<ushort16*>(mRaw->getData(col_start, row));
+        for (int col = col_start; col < col_end; col++, pixel++) {
+          // This adjustment is expected to be made with the
+          // black-level already subtracted from the pixel values.
+          // Because this is kept as metadata and not subtracted at
+          // this point, to make the correction work we subtract the
+          // appropriate amount before indexing into the curve and
+          // then add it back so that subtracting the black level
+          // later will work as expected
+          const ushort16 diff = *pixel < black_level ? *pixel : black_level;
+          *pixel = curve[*pixel - diff] + diff;
+        }
+      }
+    }
+  }
 }
 
 void IiqDecoder::checkSupportInternal(const CameraMetaData* meta) {
--- a/src/librawspeed/decoders/IiqDecoder.cpp
+++ b/src/librawspeed/decoders/IiqDecoder.cpp
@@ -127,10 +127,13 @@ RawImage IiqDecoder::decodeRawInternal() {
 
   uint32 width = 0;
   uint32 height = 0;
+  uint32 split_row = 0;
+  uint32 split_col = 0;
 
   Buffer raw_data;
   ByteStream block_offsets;
   ByteStream wb;
+  ByteStream correction_meta_data;
 
   for (uint32 entry = 0; entry < entries_count; entry++) {
     const uint32 tag = es.getU32();
@@ -151,12 +154,21 @@ RawImage IiqDecoder::decodeRawInternal() {
     case 0x10f:
       raw_data = bs.getSubView(data, len);
       break;
+    case 0x110:
+      correction_meta_data = bs.getSubStream(data);
+      break;
     case 0x21c:
       // they are not guaranteed to be sequential!
       block_offsets = bs.getSubStream(data, len);
       break;
     case 0x21d:
       black_level = data >> 2;
+      break;
+    case 0x222:
+      split_col = data;
+      break;
+    case 0x224:
+      split_row = data;
       break;
     default:
       // FIXME: is there a "block_sizes" entry?
@@ -167,6 +179,10 @@ RawImage IiqDecoder::decodeRawInternal() {
   // FIXME: could be wrong. max "active pixels" in "Sensor+" mode - "101 MP"
   if (width == 0 || height == 0 || width > 11608 || height > 8708)
     ThrowRDE("Unexpected image dimensions found: (%u; %u)", width, height);
+
+  if (split_col > width || split_row > height)
+    ThrowRDE("Invalid sensor quadrant split values (%u, %u)", split_row,
+             split_col);
 
   block_offsets = block_offsets.getStream(height, sizeof(uint32));
 
@@ -187,6 +203,8 @@ RawImage IiqDecoder::decodeRawInternal() {
   mRaw->createData();
 
   DecodePhaseOneC(strips, width, height);
+  if (correction_meta_data.getSize() != 0)
+    CorrectPhaseOneC(correction_meta_data, split_row, split_col);
 
   for (int i = 0; i < 3; i++)
     mRaw->metadata.wbCoeffs[i] = wb.getFloat();
@@ -240,6 +258,41 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
                                  uint32 width, uint32 height) {
   for (const auto& strip : strips)
     DecodeStrip(strip, width, height);
+}
+
+void IiqDecoder::CorrectPhaseOneC(ByteStream meta_data, uint32 split_row,
+                                  uint32 split_col) {
+  meta_data.skipBytes(8);
+  const uint32 bytes_to_entries = meta_data.getU32();
+  meta_data.setPosition(bytes_to_entries);
+  const uint32 entries_count = meta_data.getU32();
+  meta_data.skipBytes(4);
+
+  // this is how much is to be read for all the entries
+  ByteStream entries(meta_data.getStream(entries_count, 12));
+  meta_data.setPosition(0);
+
+  for (uint32 entry = 0; entry < entries_count; entry++) {
+    const uint32 tag = entries.getU32();
+    const uint32 len = entries.getU32();
+    const uint32 offset = entries.getU32();
+
+    switch (tag) {
+    case 0x431:
+      CorrectQuadrantMultipliersCombined(meta_data.getSubStream(offset, len),
+                                         split_row, split_col);
+      return;
+    default:
+      break;
+    }
+  }
+}
+
+void IiqDecoder::CorrectQuadrantMultipliersCombined(ByteStream data,
+                                                    uint32 split_row,
+                                                    uint32 split_col) {
+  // TODO: Implementation
+
 }
 
 void IiqDecoder::checkSupportInternal(const CameraMetaData* meta) {
--- a/src/librawspeed/decoders/IiqDecoder.h
+++ b/src/librawspeed/decoders/IiqDecoder.h
@@ -73,6 +73,10 @@ protected:
   uint32 black_level = 0;
   void DecodePhaseOneC(const std::vector<IiqStrip>& strips, uint32 width,
                        uint32 height);
+  void CorrectPhaseOneC(ByteStream meta_data, uint32 split_row,
+                        uint32 split_col);
+  void CorrectQuadrantMultipliersCombined(ByteStream data, uint32 split_row,
+                                          uint32 split_col);
 };
 
 } // namespace rawspeed


src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -151,13 +154,22 @@ RawImage IiqDecoder::decodeRawInternal() {
     case 0x10f:
       raw_data = bs.getSubView(data, len);
       break;
+    case 0x110:
+      correction_meta_data = bs.getSubStream(data);
```
LebedevRI, line 159: So now i have to re-review this all :/

Pretty sure this is
```c++
correction_meta_data = bs.getSubStream(data, len));
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -168,6 +180,10 @@ RawImage IiqDecoder::decodeRawInternal() {
   if (width == 0 || height == 0 || width > 11608 || height > 8708)
     ThrowRDE("Unexpected image dimensions found: (%u; %u)", width, height);
 
+  if (split_col < 0 || split_col > width || split_row < 0 || split_row > height)
```
LebedevRI: `split_col`/`split_row` are unsigned, you can't compare them `<0`.

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -187,6 +203,8 @@ RawImage IiqDecoder::decodeRawInternal() {
   mRaw->createData();
 
   DecodePhaseOneC(strips, width, height);
+  if (correction_meta_data.getSize() != 0)
```
LebedevRI, line 207: So if `0x110` is there, but neither of the `0x222`/`0x224` is there, we are simply assuming the 4'th quadrant (the bottom-right one)?

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+  bool quadrant_multiplier_applied = false;
+
+  for (uint32 entry = 0; entry < entries_count; entry++) {
+    const uint32 tag = entries.getU32();
+    const uint32 len = entries.getU32();
+    const uint32 offset = entries.getU32();
+
+    switch (tag) {
+    case 0x431:
+      if (!quadrant_multiplier_applied) {
```
LebedevRI: Are there cases (raws) where there more than one `0x431` tag?

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+      }
+      quadrant_multiplier_applied = true;
+      break;
+    default:
+      break;
+    }
+  }
+}
+
+// This function defines a correction that compensates for the fact
```
LebedevRI: This is C++, not C :)

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+// single multiplier, but a curve defined by seven control points.
+// Each curve's control points share the same seven X-coordinates.
+void IiqDecoder::CorrectQuadrantMultipliersCombined(ByteStream data,
+                                                    uint32 split_row,
+                                                    uint32 split_col) {
+
+  // The curves should all include (0, 0) and (65535, 65535), so the
+  // first and last points are predefined and the middle seven are
+  // read from the file
+  uint32 shared_x_coords[9] = {0, 0, 0, 0, 0, 0, 0, 0, 65535};
```
LebedevRI:
```c++
using Coords = std::array<uint32, 9>;

Coords shared_x_coords;
shared_x_coords.fill(0);
shared_x_coords.back() = 65535;
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+  for (int i = 0; i < 7; i++)
+    shared_x_coords[i + 1] = data.getU32();
+
+  uint32 curve_control_points_y[2][2][9];
```
LebedevRI:
```
std::array<std::array<Coords, 2>, 2> curve_control_points_y;
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+  for (int quadRow = 0; quadRow < 2; quadRow++) {
```
LebedevRI: Not sure it compiles, but something more like
```c++
for(auto& row : curve_control_points_y) {
  for(auto& col : row) {
    col[0] = 0;
    col[8] = 65535;
    // These multipliers are expressed in ten-thousandths in the
    // file
    for (int i = 0; i < 7; i++)
      col[i + 1] = (data.getU32() * shared_x_coords[i + 1]) / 10000;
  }
}
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+void IiqDecoder::CorrectQuadrantMultipliersCombined(ByteStream data,
+                                                    uint32 split_row,
+                                                    uint32 split_col) {
+
+  // The curves should all include (0, 0) and (65535, 65535), so the
+  // first and last points are predefined and the middle seven are
+  // read from the file
+  uint32 shared_x_coords[9] = {0, 0, 0, 0, 0, 0, 0, 0, 65535};
+  for (int i = 0; i < 7; i++)
+    shared_x_coords[i + 1] = data.getU32();
```
LebedevRI:
```c++
std::generate_n(std::next(shared_x_coords.begin(), 1), 7,
                [&data]() { return data.getU32(); });
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+      curve_control_points_y[quadCol][quadRow][8] = 65535;
+      for (int i = 0; i < 7; i++)
+        // These multipliers are expressed in ten-thousandths in the
+        // file
+        curve_control_points_y[quadRow][quadCol][i + 1] =
+            (data.getU32() * shared_x_coords[i + 1]) / 10000;
+    }
+  }
+
+  ushort16 curve[65536];
```
LebedevRI: `std::vector`

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+                     9, curve);
+
+      int row_start = quadRow == 0 ? 0 : split_row;
+      int row_end = quadRow == 0 ? split_row : mRaw->dim.y;
+      int col_start = quadCol == 0 ? 0 : split_col;
+      int col_end = quadCol == 0 ? split_col : mRaw->dim.x;
+
+      for (int row = row_start; row < row_end; row++) {
+        for (int col = col_start; col < col_end; col++) {
+          ushort16* pixel =
```
LebedevRI: And that is where all the performance have gone out of the window :)

LebedevRI:
```c++
for (int row = row_start; row < row_end; row++) {
  auto* pixel = reinterpret_cast<ushort16*>(mRaw->getData(col_start, row));
  for (int col = col_start; col < col_end; col++, pixel++)
    *pixel = curve[*pixel];
}
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+          *pixel = curve[*pixel];
+        }
+      }
+    }
+  }
+}
+
+// This is an implementation of the cubic spline algorithm found on
+// Wikipedia at https://en.wikipedia.org/wiki/Spline_(mathematics)
+void IiqDecoder::calculateCurve(const uint32* x_coords, const uint32* y_coords,
```
LebedevRI: Let's put it into `src/librawspeed/common/Spline.h` i guess.

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +260,153 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+                                const int num_coords, ushort16* curve) {
```
LebedevRI: It should *return* `std::vector<ushort16>` curve.

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: Also *DONT* take pointers like this.

LebedevRI (changes requested):
Nice!
Looks better than dcraw's code :)

I'm guessing you built rawspeed from within dt?
That would explain why all that compiled for you :)

src/librawspeed/decoders/IiqDecoder.cpp
bieber: For whatever reason the `len` value in this tag is garbage.  I just have to read the number of entries from the data

src/librawspeed/decoders/IiqDecoder.cpp
bieber: Oh, duh

src/librawspeed/decoders/IiqDecoder.cpp
bieber: It wouldn't necessarily have to be any kind of a quadrant correction.  It could also be bad pixels, flat field, or I think a couple of other things that don't depend on the split row/column

src/librawspeed/decoders/IiqDecoder.cpp
bieber: Oddly enough, there seem to be multiple of them in all my RAWs.  I'm not sure how the manufacturer's software deals with this, but dcraw just applies whatever is the first quadrant correction tag (of any of the three types it handles) in the file

src/librawspeed/decoders/IiqDecoder.cpp
bieber: Damn, I guess I can't skip learning the new C++ ;)

src/librawspeed/decoders/IiqDecoder.cpp
bieber: Okay, so I'm guessing I should just pass the coords vectors around as values since they're small-is?

src/librawspeed/decoders/IiqDecoder.cpp
bieber: Not a reference (or auto_ptr or something)?  Seems like a big-ish thing to copy, but I guess four times an image it shouldn't be too bad

bieber:
I did build it all from within DT, should I pull and compile this separately?  What am I missing because of that?

LebedevRI:
> What am I missing because of that?

E.g. all the warnings. dt builds it with `-w`.

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: There won't be any copy.

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: > dcraw just applies whatever is the first quadrant correction tag

So just return after the `CorrectQuadrantMultipliersCombined()` call?

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: :(

src/librawspeed/decoders/IiqDecoder.cpp
bieber: I could do that for now if you want, it just won't make sense once there's more corrections in here that can apply to the same image

LebedevRI: > once there's more corrections in here that can apply to the same image

I have already replied to that in https://github.com/darktable-org/rawspeed/pull/117#issuecomment-347901618
The decision of *what* is rawspeed's job is case-by-case based.

bieber: Will you still want the bad pixel corrections in RS?  Those appear alongside the quadrant multiplier

bieber:
Ahh, that explains a lot.  I'll have to try building RawSpeed on its own...I'm guessing that's also the reason I'm failing all these CI builds?

src/librawspeed/decoders/IiqDecoder.cpp
LebedevRI: Yes, bad pixel are fair game :)
But that is for future pr's.

LebedevRI:
> I'm guessing that's also the reason I'm failing all these CI builds?

That would be my guess

LebedevRI:
> I'm guessing that's also the reason I'm failing all these CI builds?

That would be my guess
Alternatively, provide the sample set, and i could try to take it from there.

Not too sure yet whether to merge into `git develop` or other branch, we are kinda in the feature freeze.

bieber:
Here's my first attempt at C++-ifying it.  Sorry if I've done anything ghastly, just let me know what needs to change

src/librawspeed/common/Spline.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "common/Common.h" // for ushort16
+#include "common/Point.h"  // for iPoint2D
+#include <vector>          // for vector
+
+namespace rawspeed {
+
+// This is an implementation of the cubic spline algorithm found on
+// Wikipedia at https://en.wikipedia.org/wiki/Spline_(mathematics)
+std::vector<ushort16>
+Spline::calculateCurve(std::vector<iPoint2D> control_points) {
```
LebedevRI: Just so that i have asked: the dcraw's `cubic_spline()` is indeed the cubic spline, just like this code, not counting bugs?
I.e. not counting bugs, they are supposed to produce the same curve?

bieber:
Whoops, one more change.  I forgot to build RawSpeed on its own and missed some warnings, they should be fixed now

src/librawspeed/common/Spline.cpp
bieber: Mathematically I believe they should be doing the same thing, but as far as I can tell (I haven't attempted to fully comprehend dcraw's implementation) we're using different algorithms to accomplish it.  So it's entirely possible that we'll end up infinitesimally different coefficients due to rounding

LebedevRI: ok, so at most rounding issues.

bieber: I believe so, but if it will make you feel better I could try comparing the curve values between my and dcraw's implementation for the same image to see if there's any difference

LebedevRI: As long as both the implementations are of the *same* algorithm, and the IIQ de-quadrant-ization is actually meant to be corrected via that algorithm, i suppose it is ok. We don't want to mimic some other implementation. What it needs is tests against trustworthy spline implementation in some mathematical software, but that is for the future.

bieber: Okay, we should be good then.  It's definitely doing the same fundamental thing

bieber:
Any thoughts on the rest of the new version, @LebedevRI?  Should I switch the PR to a new branch if you're keeping new features out of develop for now?

src/librawspeed/common/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ FILE(GLOB SOURCES
   "RawImageDataFloat.cpp"
   "RawImageDataU16.cpp"
   "RawspeedException.h"
+  "Spline.h"
+  "Spline.cpp"
```
LebedevRI: Yes, i *really* did mean `Spline.h`, **not** `Spline.cpp`

src/librawspeed/common/Spline.cpp
LebedevRI:
```diff
-      std::vector<iPoint2D>  control_points
+const std::vector<iPoint2D>& control_points
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +261,91 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+  // first and last points are predefined and the middle seven are
+  // read from the file
+  std::array<uint32, 9> shared_x_coords;
+  shared_x_coords.fill(0);
+  shared_x_coords.back() = 65535;
+  std::generate_n(std::next(shared_x_coords.begin()), 7,
+                  [&data] { return data.getU32(); });
+
+  std::array<std::array<std::vector<iPoint2D>, 2>, 2> control_points;
+  for (int quadRow = 0; quadRow < 2; quadRow++) {
```
LebedevRI: Something like this *should* be better:
```c++
for (const auto& row : control_points) {
  for (auto& col : row) {
    col.reserve(9);
    col.emplace_back(0, 0);

    // These multipliers are expressed in ten-thousandths in the file
    for (int i = 1; i < 8; i++) {
      const int y_coord = (data.getU32() * shared_x_coords[i]) / 10000;
      col.emplace_back(shared_x_coords[i], y_coord);
    }
    col.emplace_back(65535, 65535);
    assert(col.size() == 9);
  }
}
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -242,6 +261,91 @@ void IiqDecoder::DecodePhaseOneC(const std::vector<IiqStrip>& strips,
...
+        control_points[quadRow][quadCol].push_back(iPoint2D(
+            shared_x_coords[i], (data.getU32() * shared_x_coords[i]) / 10000));
+
+      control_points[quadRow][quadCol].push_back(iPoint2D(65535, 65535));
+    }
+  }
+
+  for (int quadRow = 0; quadRow < 2; quadRow++) {
+    for (int quadCol = 0; quadCol < 2; quadCol++) {
+      std::vector<ushort16> curve =
```
LebedevRI:
```diff
-      std::vector<ushort16> curve = Spline::calculateCurve(control_points[quadRow][quadCol]);
+const std::vector<ushort16> curve = Spline::calculateCurve(control_points[quadRow][quadCol]);
```

LebedevRI (changes requested):
Looks better! :)

Re PR: i'm not sure :/ It is not a bugfix, but a (mid-sized) feature.

It would be best to polish it in the PR, and merge once 2.4.0 is out.
That way i don't have to think about one more branch,
about pull-requests with some new features for that branch, etc.

Since for whatever reasons 2.4.x series will keep using the `git develop`, it's not like not merging it here and now will loose much :)

src/librawspeed/common/CMakeLists.txt
bieber: Oh, do I not need to include the .cpp file in the list?

LebedevRI: I.e. there should not be `Spline.cpp`, just `Spline.h`.

src/librawspeed/common/Spline.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "common/Spline.h"
+#include "common/Common.h" // for ushort16
+#include "common/Point.h"  // for iPoint2D
+#include <vector>          // for vector
+
+namespace rawspeed {
+
+// This is an implementation of the cubic spline algorithm found on
+// Wikipedia at https://en.wikipedia.org/wiki/Spline_(mathematics)
+std::vector<ushort16>
```
LebedevRI: Just move this code into the `Spline.h`

src/librawspeed/common/Spline.h
```diff
@@ -0,0 +1,35 @@
...
+#include "common/Common.h" // for ushort16
+#include "common/Point.h"  // for iPoint2D
+#include <vector>          // for vector
+
+namespace rawspeed {
+
+class Spline final {
+public:
+  static std::vector<ushort16>
+  calculateCurve(const std::vector<iPoint2D>& control_points);
```
LebedevRI: I.e.
```c++
class Spline final {
public:
  static std::vector<ushort16>
  calculateCurve(const std::vector<iPoint2D>& control_points) {
    const int num_coords = control_points.size();
    const int num_segments = num_coords - 1;
    // These are the constant factors for each segment of the curve.
...
```

src/librawspeed/decoders/IiqDecoder.h
```diff
@@ -73,6 +73,14 @@ class IiqDecoder final : public AbstractTiffDecoder {
   uint32 black_level = 0;
   void DecodePhaseOneC(const std::vector<IiqStrip>& strips, uint32 width,
                        uint32 height);
+  void CorrectPhaseOneC(ByteStream meta_data, uint32 split_row,
+                        uint32 split_col);
+  void CorrectQuadrantMultipliersCombined(ByteStream data, uint32 split_row,
+                                          uint32 split_col);
+
+private:
+  void calculateCurve(const uint32* x_coords, const uint32* y_coords,
```
LebedevRI: This is dead code.

src/librawspeed/common/Spline.h
bieber: Ah, I see.  What's the guideline for putting implementations in .h files instead of a separate .cpp?

LebedevRI: Dunno, nothing specific.

bieber:
@LebedevRI: I've run into another rendering issue that I could use some help with, if you're interested in taking a look into it.  I'm afraid I've hit the limits of my abilities here.  I just uploaded a RAW file to raw.pixls.us that showcases it (it's a new Leaf Credo 40 image).

It seems that while applying the quadrant curve correction fixes mismatches in the highlight areas, it simultaneously *introduces* mismatches in the shadows of some images.  For instance, in the RAW file of a squirrel I uploaded to raw.pixls.us you can see a vertical seam above the squirrel's head, and a horizontal one to the right of it.

I've been trying to figure out what's wrong with the curve correction all night, but as best I can figure out the implementation here is fine.  It produces very slightly different results than dcraw's algorithm at some points, but I've found that neither copy/pasting dcraw's curve algorithm nor even directly loading the curve as calculated by dcraw from a file fixes this issue.

I've also turned off all the other types of phase one correction in my hacked version of dcraw, but somehow it's still rendering the file perfectly without any visible seams, while the seams persist in darktable.

Presumably this means I'm missing something somewhere else, but I have no idea where it could be.  Is it possible that on of DT's steps after the RAW decoding phase could be introducing this artifact?  Any suggestions on how I could hunt down the source?

LebedevRI:
I do confirm that i see the vertical seam if i push all the sharpen/contrast-like modules to the max.
And it is **not** caused by [haze removal module](https://redmine.darktable.org/issues/11862).

> I've been trying to figure out what's wrong with the curve correction all night, but as best I can figure out the implementation here is fine. It produces very slightly different results than dcraw's algorithm at some points, but I've found that neither copy/pasting dcraw's curve algorithm nor even directly loading the curve as calculated by dcraw from a file fixes this issue.

and

> I've also turned off all the other types of phase one correction in my hacked version of dcraw, but somehow it's still rendering the file perfectly without any visible seams, while the seams persist in darktable.

lead me to believe that the curve is meant to be used with different data.
`phase_one_load_raw_c()` looks way more sophisticated than `IiqDecoder::DecodeStrip()`

Of particular interest is:

```c
for (i=0; i < 256; i++)
  curve[i] = i*i / 3.969 + 0.5;
...
if (ph1.format == 5 && pixel[col] < 256)
  pixel[col] = curve[pixel[col]];
```
which looks like de-linearization of dark pixels (i'm almost betting it's this one)

and
```c
fseek (ifp, ph1.black_col, SEEK_SET);
if (ph1.black_col)
  read_shorts ((ushort *) cblack[0], raw_height*2);
rblack = cblack + raw_height;
fseek (ifp, ph1.black_row, SEEK_SET);
if (ph1.black_row)
  read_shorts ((ushort *) rblack[0], raw_width*2);

for (col=0; col < raw_width; col++) {
  i = (pixel[col] << 2*(ph1.format != 8)) - ph1.black
    + cblack[row][col >= ph1.split_col]
    + rblack[col][row >= ph1.split_row];
  if (i > 0) RAW(row,col) = i;
}
```
Which is again looks like some kind of adjustment.

So i guess try disabling/disecting more of dcraw and see when it starts to produce the same result.

> Is it possible that on of DT's steps after the RAW decoding phase could be introducing this artifact?

I doubt it.

bieber:
Thanks @LebedevRI!  Hacking around a little more, it looks like the second bit you identified, the black level subtraction, is what did it.  I see that in DecodeStrip there's no black level subtraction going on at all, however, it's just written to a field on the raw image class.  I'm guessing that means that black level is subtracted somewhere later on in the process?

That might make things tricky here, because it seems like the corrections being made here are dependent on there already having been both the general black level and a quadrant-specific black level applied to the pixels beforehand.  I guess to make this work I might need to  do the black level adjustment before making the correction, then add the general black level back in to all the pixels after the correction is done?

bieber:
Or I guess I could just have the black levels on hand in the phase one corrections function, and then do something like `curve[*pixel - blackLevel + colBlackLevel + rowBlackLevel] + blackLevel` in place of the current curve lookup

bieber:
Yep, subtracting the black level before indexing into the table and then adding the black level back in eliminates the seams perfectly.  I don't even need the per-quadrant black levels, they're all 0 on my camera so I'll have to find a sample RAW file with them to be able to test that feature anyways.  I'll update with the black level correction tonight and that should resolve this issue

bieber:
We're past 2.4.0 now, how would you feel about merging this @LebedevRI?

LebedevRI:
Indeed.
Generally, that iiq decoder looks quite partial as compared to what dcraw does, but for that same reason, i suppose this will have to do for now.

bieber:
Yeah, it's a start ;)  I'll try to add some others over time as well, but this seems like the big one for the new Credo backs, at least.  I'll have to check and see which corrections are actually present in the other Leaf/Phase One RAWs we have on raw.pixls.us later, but this should be good to go for now, right?  Just let me know if anything else needs to change

LebedevRI:
>  this should be good to go for now

I think so, but i wasted an entire day chasing The weirdest performance regression in 2d8507d1909829e9151fac73ffc0d20f691f558c, which didn't even get me anywhere, so i didn't have had the final look yet.

src/librawspeed/common/Spline.h
```diff
@@ -0,0 +1,95 @@
...
+
+    for (int i = num_segments - 1; i >= 0; i--) {
+      c[i] = z[i] - mu[i] * c[i + 1];
+      b[i] = (a[i + 1] - a[i]) / h[i] - h[i] * (c[i + 1] + 2 * c[i]) / 3.;
+      d[i] = (c[i + 1] - c[i]) / (3. * h[i]);
+    }
+
+    std::vector<ushort16> curve(65536);
+
+    for (int i = 0; i < num_segments; i++) {
```
LebedevRI, line 81: Uhm, why there are two loops?
Either the outer one is not needed, or the inner one is incorrect.

LebedevRI: Ah, or maybe this is correct.

bieber: The outer loop iterates through the segments, inner loop iterates through the range of coordinates in each segment.  You could do everything with one loop, but I think it reads a lot cleaner this way

src/librawspeed/common/Spline.h
```diff
@@ -0,0 +1,95 @@
...
+      a[i] = control_points[i].y;
+
+    for (int i = 0; i < num_segments; i++)
+      h[i] = control_points[i + 1].x - control_points[i].x;
+
+    for (int i = 1; i < num_segments; i++)
+      alpha[i] =
+          (3. / h[i]) * (a[i + 1] - a[i]) - (3. / h[i - 1]) * (a[i] - a[i - 1]);
+
+    l[0] = mu[0] = z[0] = 0;
```
LebedevRI, line 61: @bieber but the https://en.wikipedia.org/wiki/Spline_(mathematics) says `l_0 = 1`?

bieber: Ooh, yeah, I think I missed the 1 while reading through.  I'm not sure how consequential it is, but that's definitely a mistake

LebedevRI: I'll add some reference curves from [scipy.interpolate.CubicSpline](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.interpolate.CubicSpline.htm), and then see how this changes things.

bieber: Actually, looking at the rest of the function I don't think l[0] is ever read (the next loop starts at 1 and references l[i] a bunch of times)

cytrinox:
@bieber is the test file still around? I've no other sample file with these artefacts to test my own implementation. Thanks!

bieber:
Hi @cytrinox, here's a copy: https://drive.google.com/file/d/1zxEqyx9nUn9YnYeBJq2F58cFidsKHSgK/view?usp=sharing

PR #120, opened 2018-01-11
Added install steps to the librawspeed CMakeLists.txt

guysherman:
Copies the .h files to <prefix>/include/librawspeed
Copies the lib to <prefix>/lib
Copies the rawspeedconfig.h and ThreadSafetyAnalysis.h files to
<prefix>/lib/librawspeed
Generates a pkgconfig file and copies to <prefix>/lib/pkgconfig

I've put a version string into the .pc file, but I don't really know what it should be, so very open to feedback on that. It might be worth putting a version string into the top-level CMakeLists.txt file that can be passed down where it is needed.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -15,3 +15,16 @@ add_subdirectory(librawspeed)
 if(BUILD_TOOLS)
   add_subdirectory(utilities)
 endif()
+
+set(PRIVATE_LIBS "-lpthread -ljpeg -lz -lrawspeed")
+set(DEST_DIR "${CMAKE_INSTALL_PREFIX}")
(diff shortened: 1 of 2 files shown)


LebedevRI:
https://github.com/darktable-org/rawspeed/pull/119#issuecomment-356519287
```
@guysherman Hi.
RawSpeed is a copylib, a particular version (as chosen by the app developers, maybe with patches) is supposed to be built as part of the application it is going to be used in.
```

LebedevRI:
Having it install this stuff will allow/simplify distros to unbundle the particular version of rawspeed used in some program, which is a) just was never supposed to happen, and is pretty much guaranteed to break stuff  and b) will give an impression that it is *not* a copylib.
Neither of these things are acceptable.

As for version string, since there is no installation, there is no need for that.

I'm sorry for wasting your time :(

PR #131, opened 2018-05-20
Add k-1 II support.

mazhe:
Hi, I juste received my Pentax K-1 II and am working on darktable support.

Since the sensor is strictly the same as the K-1 mark I, I don't think there should be any change in values? I've sent a first sample to raw.pixls.us

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -7519,6 +7519,17 @@
 		<Crop x="4" y="18" width="-10" height="-2"/>
 		<Sensor black="64" white="16316"/>
 	</Camera>
+	<Camera make="RICOH IMAGING COMPANY, LTD." model="PENTAX K-1 Mark II">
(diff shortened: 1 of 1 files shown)


LebedevRI:
Yes, i have seen the pr.
Sorry, did not to this yet..

mazhe:
Hello, any progress regarding this request? Then I can open one for the rest of the support in darktable...

Thanks!

LebedevRI:
Uhm, right, sorry.
With some tuning, merged.

Please contribute the full sample set for that camera - i was so sure there was coverage already - to the RPU - just two samples - DNG and PEF.
No need for pixel shift, or crops.

mazhe:
Thanks, I already tried to upload some samples (as I did with the first k-1), but they never showed up... Is there some size limit or something like that? I'll try again this weekend.

LebedevRI:
There is a pre-review. I don't think i remember seeing any dng+pef samples for `k-1 II`.

PR #135, opened 2018-05-21
Improve comprehensibility of RW2 decoder/decompressor interaction

shoffmeister:
Rename load_lags to section_split_offset to clarify what really happens - after all, load flags is not a flag, but used only as an offset.

This does not introduce functional change. This is only meant to improve general maintainability.


LebedevRI:
This is pretty clearly not `clang-format`ted
https://github.com/darktable-org/darktable/blob/master/CONTRIBUTING.md#coding-style

shoffmeister:
Correct, I did not send this through a pretty-printer; I kept all formatting exactly as it was, which has benefits of its own, I believe.

Anyway, I have added the git pre-commit clang-format-6-0 check to my rawspeed working copy. (FWIW, in Ubuntu 18.04 all it takes is "apt install clang-format" and then creating a soft link "clang-format" pointing to clang-format-6-0)

PR #146, opened 2018-09-25
Add support for GoPro's .GPR file format

stloeffler:
Initial, partial implementation of a VC-5 decompressor based on
https://github.com/gopro/gpr to decompress GoPro's .GPR file format.
Only tags that are needed are handled at the moment.


src/librawspeed/common/table17.inc
```diff
@@ -0,0 +1,285 @@
...
+	Codebook automatically generated by Huffman program
+	Created: Wed May 11 2005 21:23:08 UTC
+
+	Modified by hand to simplify for use with the reference decoder.
+	
+	Note that the codebook entries with a run length of zero are
+	special codewords that are used to mark significant locations
+	in the encoded bitstream.
+
+	(c) 2013 Society of Motion Picture & Television Engineers LLC and Woodman Labs, Inc.
```
LebedevRI: I don't think this will fly

stloeffler: IANAL, but as the first comment line states, this file was taken as-is from the gopro/gpr repository, which is dual-licensed under Apache2 or MIT so it should be compatible. As I don't have access to the original specs, there is not much else I can do about this.

supertobi: I opened a issue:
https://github.com/gopro/gpr/issues/8

supertobi: Fixed now:
https://github.com/gopro/cineform-sdk/blob/master/Common/table17.inc

supertobi: Sorry, wrong link:
https://github.com/gopro/gpr/blob/d2d78d1b236a278a13ed5abee6870af2b490c89f/source/lib/vc5_common/table17.inc

stloeffler: Thanks @supertobi , I updated the PR.

LebedevRI: Thanks, i suppose that is better / as good as it will get.
I guess i now need to read up about using "dual-licensed under Apache2 or MIT" in LGPL library.

LebedevRI:
Since there does not _seem_ to be a licensing issue, i suppose, i will be, as usual,
able to close my eyes, should i decide to merge this, and then cleanup afterwards.

So what is the status of the implementation?
As in, how correctly is this all implemented?
If adobe dng converter can process those "DNGs", do the hashes match?

LebedevRI:
Oh damn, i should have looked into static analysis build failing to propagate the build status earlier :/

stloeffler:
The implementation produces a "correctly looking" image, but the hashes don't match (yet). I will look into this when I have time.

LebedevRI:
> The implementation produces a "correctly looking" image, but the hashes don't match (yet).

I see.
So i guess i should forget about this for a bit before going in with a hammer.

stloeffler:
It seems the problem was "just" a typo and some rounding errors. Should be fixed now. At least for the 5 GoPro (Hero5 + Hero6) samples on https://raw.pixls.us/, the hashes for the .gpr file and for the .dng files created with the Adobe DNG converter now match.

LebedevRI:
Sweet!

supertobi:
Btw. the gpr project has a GoPro Fusion and a Hero7 Sample file here:
https://github.com/gopro/gpr/tree/master/data/samples

LebedevRI:
They are absolutely welcomed to contribute to [RPU](https://raw.pixls.us/).

stloeffler:
Thanks for the link to the samples, @supertobi. They allowed me to track down and fix another clamping issue.
At least for the linked samples (RPU and gopro/data), the hashes now match with the Adobe-converted dngs.

data/cameras.xml
```diff
@@ -2066,6 +2066,46 @@
 			<Hint name="wb_offset" value="142"/>
 		</Hints>
 	</Camera>
+	<Camera make="GoPro" model="FUSION">
```
LebedevRI: Why is this even here?
It's DNG.

src/librawspeed/decompressors/VC5Decompressor.cpp
```diff
@@ -0,0 +1,614 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public
+    License along with this library; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+*/
+
+/*
+ * The code in this file is based on the Apache2/MIT dual-licensed code of the
```
LebedevRI: Define `based`?
Is there copy-pasted code?
I really would want to avoid having non-LGPL code.

data/cameras.xml
stloeffler: It is based on DNG, but uses a special encoding/compression.
Anyway, the <camera> declarations can be removed if they are not required elsewhere in rawspeed.

LebedevRI: This should be reduced down to
```xml
	<Camera make="GoPro" model="FUSION" mode="dng">
		<ID make="GoPro" model="FUSION">GoPro FUSION</ID>
	</Camera>
...
```

src/librawspeed/decompressors/VC5Decompressor.cpp
stloeffler: No, there is no directly copy-pasted code. Only the table17.inc file and the definitions needed by it (_rlv, RLVTABLE) are copied verbatim and marked as such.
However, the algorithm is logic is the same; as I stated above, I did not have access to the original VC5 specs, so the code is a reimplementation of the gopro code and I would therefore consider it a derivative work.
If it helps, I could remove the explicit mentioning of the licensing, but I think the original code should still be credited.

src/librawspeed/decompressors/VC5Decompressor.h
```diff
@@ -0,0 +1,125 @@
...
+
+class ByteStream;
+class RawImage;
+
+// Decompresses VC-5 as used by GoPro
+
+#define VC5_LOG_TABLE_SIZE 4096
+
+class VC5Decompressor final : public AbstractDecompressor
+{
```
LebedevRI: Is it _so_ hard to simply install clang-format git hook?

data/cameras.xml
stloeffler: Is the <ID> tag actually necessary? From the doctype spec at the beginning of the file it seems that <ID> is optional, presumably if make and model are the same as in the <Camera> tag and the display string is just the concatenation of the two?
I take it the same applies to the other GoPro tags I added (Hero5 - Hero7)?

src/librawspeed/decompressors/VC5Decompressor.h
stloeffler: Sorry, I don't normally use clang, and don't have that tool installed by default (and did not know I should have, either). I will look into it if required.

src/librawspeed/decompressors/VC5Decompressor.cpp
LebedevRI: > No, there is no directly copy-pasted code. 

Great. Then i would prefer something much more vague.

"This is a decompressor for VC-5 raw compression algo, as used in GoPro raws.
This implementation is similar to that one of the official reference implementation in https://github.com/gopro/gpr project, and is producing bitwise-identical output
as compared with the Adobe DNG Converter implementation."

Or something, don't know.
But actually talking about the license, i would say, does not give any positive benefit for us.

data/cameras.xml
LebedevRI: It will (well, should, unless they really messed with their 'DNG') work without all these `data/` changes.
I don't understand that `<ID>` tag, but that is how it is structured.
Having this minimal snipped i posted will allow for that camera to be listed in the "cameras known/supported as per the rawspeed/data/cameras.xml" lists, as seen on e.g. https://rawspeed.org/CameraSupport.html, https://www.darktable.org/resources/camera-support/, https://raw.pixls.us/?missing#repo

src/librawspeed/decompressors/VC5Decompressor.h
LebedevRI: `clang-format` is standalone, so you wouldn't even need to use clang.
Really, everyone should be using it this late in this decade.

It may be best for you to format the new files (git blame reasons :))

LebedevRI:
(ignore the "AppVeyor build failed", i think they are broken as usual)

adeelabbas:
I believe a new Hero7 sample image has just been uploaded to [https://raw.pixls.us/](https://raw.pixls.us/)

LebedevRI:
> Thanks for the link to the samples, @supertobi. They allowed me to track down and fix another clamping issue.

Hopefully just the GOPR9231.GPR (HERO7) sample is sufficient to see that issue?

stloeffler:
FWIW, 1407898f produces the wrong hash for the Hero7 sample whereas 92315b96 produces the correct one.

LebedevRI:
I'll take that as a "yes" :)

stloeffler:
I think I tackled all the issues so far. Is there anything else I should do / look into?

LebedevRI:
> I think I tackled all the issues so far. 

Thank you!

> Is there anything else I should do / look into?

I don't know. Probably the rest is for me, but i have not looked yet..

LebedevRI:
Aright, hopefully that was enough of a headroom.

Ugh, they use DNG, and don't use tiling?!
https://docs.google.com/spreadsheets/d/1sdTd6PQkvUDSxjgqgjbEU1fYcc0aA6n8YSxbxfjtKoU/edit?usp=sharing

stloeffler:
Yes, it's based on DNG, but uses wavelets instead of tiling. I.e., the image is split into a low frequency (low-resolution) quarter and three high-frequency quarters. This procedure is applied repeatedly for the low-frequency part (see https://github.com/gopro/gpr#wavelet-transforms). For reconstruction, the process has to be reversed. The advantage is that if only a low-resolution "preview" is sufficient, the decoding can stop early. But if (a portion of) the final image is required, the decoding process must run to the end. With quite a bit of math, it might be possible to reduce the number of pixels that have to be decoded in some cases (e.g. if only the top-left pixel is required), but this will fail in other cases (e.g. if a central pixel is required).

It would probably be possible to parallelize the algorithm, e.g., on a row-by-row basis, but I don't know enough about how rawspeed handles multi-threading to tackle this. Also, such an implementation will likely depend on the memory architecture, will have multiple synchronization points, etc., so it will make things quite a bit more complicated.

LebedevRI:
(I have merged this into my local branch, so any further changes here will not be reflected.)

LebedevRI:
Awesome, hashes for **all** the _existing*_ RPU samples indeed match:
```
raw.pixls.us/GoPro$ grep -ri md5sum | sort
HERO5 Black/GOPR0020.GPR.hash:md5sum of per-line md5sums: a51a89236dbda9772fcbe0a0689cd1d0
HERO5 Black/GOPR0020.dng.hash:md5sum of per-line md5sums: a51a89236dbda9772fcbe0a0689cd1d0
HERO5 Black/GOPR2435.GPR.hash:md5sum of per-line md5sums: 7357254b07915e8c30a494dfeca10c5f
HERO5 Black/GOPR2435.dng.hash:md5sum of per-line md5sums: 7357254b07915e8c30a494dfeca10c5f
HERO6 Black/GOPR8921.GPR.hash:md5sum of per-line md5sums: 0385660735e740db31c766fdbcb0b0c8
HERO6 Black/GOPR8921.dng.hash:md5sum of per-line md5sums: 0385660735e740db31c766fdbcb0b0c8
HERO6 Black/GOPR9172.GPR.hash:md5sum of per-line md5sums: 5b6b2991ee341361b188f5f85775e214
HERO6 Black/GOPR9172.dng.hash:md5sum of per-line md5sums: 5b6b2991ee341361b188f5f85775e214
HERO6 Black/GOPR9231.GPR.hash:md5sum of per-line md5sums: 1d8ee456a1e2b60ea9d98f3ce8994763
HERO6 Black/GOPR9231.dng.hash:md5sum of per-line md5sums: 1d8ee456a1e2b60ea9d98f3ce8994763
HERO7 Black/GOPR9231.GPR.hash:md5sum of per-line md5sums: 15d1f9823eff1bfea02eddb25344325b
HERO7 Black/GOPR9231.dng.hash:md5sum of per-line md5sums: 15d1f9823eff1bfea02eddb25344325b
```
* No Fusion sample[s]

src/librawspeed/decompressors/VC5Decompressor.cpp
```diff
@@ -0,0 +1,689 @@
...
+  // Decode final wavelet into image
+  Array2D<uint16_t> out(reinterpret_cast<uint16_t*>(mImg->getData()),
+                        static_cast<unsigned int>(mImg->dim.x),
+                        static_cast<unsigned int>(mImg->dim.y),
+                        mImg->pitch / sizeof(uint16_t));
+
+  unsigned int width = 2 * mTransforms[0].wavelet[0].width;
+  unsigned int height = 2 * mTransforms[0].wavelet[0].height;
+
+  Array2D<int16_t> channels[4];
```
LebedevRI, line 633: Hello, leak, long time no see.

src/librawspeed/decompressors/VC5Decompressor.cpp
```diff
@@ -0,0 +1,689 @@
...
+
+// static
+template <class T>
+VC5Decompressor::Array2D<T>
+VC5Decompressor::Array2D<T>::create(const unsigned int width,
+                                    const unsigned int height) {
+  return Array2D<T>(new T[width * height], width, height);
+}
+
+template <class T> void VC5Decompressor::Array2D<T>::destroy() {
```
LebedevRI, line 95: Nice, this required **deep** thought to understand.

LebedevRI:

> It would probably be possible to parallelize the algorithm

See https://github.com/darktable-org/rawspeed/commit/163aaee9c86e2fdef327bb7ae2d9bbf2642297d4 for the *actual* perf numbers (or, the commit log for per-commit perf changes),
but it was pretty easy and **rather** effective after some refactoring.

There is _some_ room for further improvement, but the stabilization/fuzzing is up next...

photopea:
Hi guys, I see, that GPR files have a regular DNG / TIFF structure, but they use a wavelet compression. Did you implement the wavelet decompressor yourself, or did you used an existing implementation (e.g. CineForm)?

Could you share some document, which describes this compression method? The GPR Github page contans just some general info.

LebedevRI:
I won't comment for @stloeffler, but i referred to the open-source code published at https://github.com/gopro/gpr

> Could you share some document, which describes this compression method? The GPR Github page contans just some general info.

I'm not aware of any such documents.

IIRC you are writing all this code in webassembly? How mature is the language nowadays?
I wonder if you could just compile (e.g. this) existing library and just use it.

photopea:
@LebedevRI Thanks. I managed to find a document, which describes the VC-5 format (used in GPR compression).

www.Photopea.com can open GPR files now. I write my code in Javascript.

I think I could compile your library to webassembly, but it would be probably too huge. As Photopea is opened by millions of people, it makes no sense to make it 50 kB larger because of a feature, that is needed by one person in 10000. 

Also, I am not very good with compilers (I have been writing only javascript in the last 7 years).

My VC-5 implementation is only 7.5 kB.  Out of it, 4 kB are Huffman codes, which are the same for all compressed files (and specified in VC-5 standard).

stloeffler:
I did implement the decompressor myself, based heavily on https://github.com/gopro/gpr. The documentation there indeed is more of a broad overview (which can be very good when trying to figure out which part of the image goes where), but if you dig into the code you'll find that it's very detailed and broadly applicable.

photopea:
Yes, I also looked at their implementation, as VC-5 for bayer arrays is a bit different than for regular images or videos. 

E.g. nobody mentioned, that the red, blue and the second green channel are stored as differences from the first green channel. I had to find it in their code.

PR #150, opened 2018-12-06
CMake: correctly detect OpenMP on macOS with Xcode 10 compiler

parafin:
It's now possible to use OpenMP with Xcode without the need to use gcc as compiler. One need to install libomp, pass -Xclang -fopenmp flags to compiler and link against libomp. This commit changes cmake openmp target name from OpenMP::OpenMP to OpenMP::OpenMP_CXX so that it matches upstream name from FindOpenMP.cmake. This way correct settings from that module are used. I also copied more correct version of setting up OpenMP target from upstream cmake module for older cmake versions. Original version uses separate_arguments function with NATIVE_COMMAND option, which we can't use because it's unavailable in cmake-3.4. separate_arguments is indeed required for macOS OpenMP support, because without it -Xclang -fopenmp flags are treated as one and passed to compiler quoted: "-Xclang -fopenmp". Since this commit makes sure we use values set in FindOpenMP.cmake module it doesn't matter much that we're missing separate_arguments call. Or we can add it with UNIX_COMMAND option which is available in cmake-3.4.

### DIFF
--- a/cmake/src-dependencies.cmake
+++ b/cmake/src-dependencies.cmake
@@ -46,13 +46,17 @@ if(WITH_OPENMP)
     message(STATUS "Looking for OpenMP - found")
     set(HAVE_OPENMP 1)
 
-    if(NOT TARGET OpenMP::OpenMP)
-      add_library(OpenMP::OpenMP INTERFACE IMPORTED)
-      set_property(TARGET OpenMP::OpenMP PROPERTY INTERFACE_COMPILE_OPTIONS "${OpenMP_CXX_FLAGS}")
-      set_property(TARGET OpenMP::OpenMP PROPERTY INTERFACE_LINK_LIBRARIES "${OpenMP_CXX_FLAGS}")
+    if(NOT TARGET OpenMP::OpenMP_CXX)
+      add_library(OpenMP::OpenMP_CXX INTERFACE IMPORTED)
+      if(OpenMP_CXX_FLAGS)
+        set_property(TARGET OpenMP::OpenMP_CXX PROPERTY INTERFACE_COMPILE_OPTIONS ${OpenMP_CXX_FLAGS})
+      endif()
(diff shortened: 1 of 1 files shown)


LebedevRI:
Yay! let's see..

LebedevRI (approved):
Yep, looks good. Thank you for finally looking into this!
Does not break openmp for clang or gcc on linux at least.

LebedevRI:
I'm not sure when @TurboGit is planning to do the next rc, but i'll see if i will be able to update the submodule in dt tomorrow.

parafin:
Great, thanx, I was about to ask you to do it;)

LebedevRI:
Hmm, some of these changes broke some builds.

parafin:
What are the errors?

LebedevRI:
I'm currently trying to assess whether the problem is in rawspeed, or on darktable.
https://build.opensuse.org/build/graphics:darktable:master/Debian_9.0/x86_64/darktable/_log : [log.txt](https://github.com/darktable-org/rawspeed/files/2658209/log.txt)

LebedevRI:
Okay, the problem is here too.
https://build.opensuse.org/package/live_build_log/graphics:darktable:master/rawspeed/openSUSE_Leap_42.3/x86_64
[log.txt](https://github.com/darktable-org/rawspeed/files/2658246/log.txt)

parafin:
Okay, the issue is that in older cmake there's no OpenMP_CXX_LIBRARIES just OpenMP_CXX_FLAGS

LebedevRI:
Ah, that would make sense.

parafin:
I guess we can just use concatenation of both variables for INTERFACE_LINK_LIBRARIES. Or check if OpenMP_CXX_LIBRARIES set and fallback to OpenMP_CXX_FLAGS if not.

PR #152, opened 2018-12-07
Cmake: fix OpenMP for older cmake versions (<3.9)

parafin:
Actually I checked and in cmake 3.8 there's no OpenMP_CXX_LIBRARIES and no OpenMP target, while in cmake 3.9 there's both. So either we use pre-defined OpenMP target or OpenMP_CXX_FLAGS


.travis.yml
```diff
@@ -88,13 +88,13 @@ jobs:
     sudo: required
     services:
       - docker
-    env: CC=clang-7 CXX=clang++-7 TARGET=STATICANALYSIS ECO="-DUSE_CLANG_TIDY=ON -DWITH_OPENMP=OFF"
```
LebedevRI, line 91: I'm very sure this should stay as-is :)
There is no libomp in docker image https://hub.docker.com/r/darktable/rawspeed/builds/bkscbkihgtmqags6bakwylb/

cmake/src-dependencies.cmake
```diff
@@ -52,9 +52,7 @@ if(WITH_OPENMP)
       add_library(OpenMP::OpenMP_CXX INTERFACE IMPORTED)
```
LebedevRI, line 52: Can you please add a comment here:

```
// FIXME: OpenMP::OpenMP_CXX target, and ${OpenMP_CXX_LIBRARIES} were both added in cmake-3.9
if(NOT TARGET OpenMP::OpenMP_CXX)
```

LebedevRI:
@parafin thank you!
I'll wait for https://build.opensuse.org/project/monitor/graphics:darktable:master to finish, and update dt's version.

.travis.yml
parafin: It's an error if OpenMP is not detected in the system?

LebedevRI: > It's an error if OpenMP is not detected in the system?

It is since 2a3b9dc913d2cb95965b7e545044cd23a4e04293

> I will remove this commit.

Did that when merging.

cmake/src-dependencies.cmake
parafin: Not sure it's a FIXME, more of TOREMOVE after cmake requirement is updated.

LebedevRI:
Great, https://build.opensuse.org/project/monitor/graphics:darktable:master is all green again.
Hopefully OSX is still good, too.

parafin:
It is. Everything is ready for RC build.

TurboGit:
That's great news!

PR #153, opened 2018-12-17
use travis homebrew addon for reliability

stephengroat:


LebedevRI:
Not sure that is an improvement.
Previously all these were in one place, now they are not.
Also, to be noted, those Travis OSX builds are currently disabled anyway.

stephengroat:
just saw that. no worries, just a suggestion

PR #154, opened 2018-12-25
Add Sony RX100M6

bobobo1618:


LebedevRI:
(yes, i see this pr, will take a look soon)

LebedevRI:
_Please_ contribute/replace the raw sample for this camera on https://raw.pixls.us
The existing sample is unusably noisy, and has rather extreme vignetting.

bobobo1618:
I've uploaded a new one (though the old one wasn't mine). It was taken outdoors but it's quite overcast so I had to turn ISO/noise up.

Now that I look at it again, I have a feeling my white levels are off and should be set to 15360.

LebedevRI:
> I've uploaded a new one (though the old one wasn't mine). It was taken outdoors but it's quite overcast so I had to turn ISO/noise up.

Thank you, but sadly that sample won't do, it has people in it..

> Now that I look at it again, I have a feeling my white levels are off and should be set to 15360.

`15360` is indeed what EXIF metadata says for *that* shot.
https://github.com/darktable-org/rawspeed/blob/develop/src/librawspeed/decoders/ArwDecoder.cpp does not seem to currently read white level from makernotes, thus yes, it will be ever so slightly off every so often. (possibly black levels, too)

bobobo1618:
> Thank you, but sadly that sample won't do, it has people in it..

What exactly do you need from the photo? Low noise, no people and the site says it should be in focus but does the whole image need to be in focus or does there just have to be a sharp section? I use the camera for birds a lot so I can upload one of those if it works.

> 15360 is indeed what EXIF metadata says for that shot.

I've checked quite a number of photos taken with the camera and I'm yet to see one that shows anything different.

LebedevRI:
> > Thank you, but sadly that sample won't do, it has people in it..
> 
> What exactly do you need from the photo? Low noise, no people and the site says it should be in focus but does the whole image need to be in focus or does there just have to be a sharp section? I use the camera for birds a lot so I can upload one of those if it works.

The only rules are specified on the front page of https://raw.pixls.us
A bird should do, as long as it's not that noisy :)

> > 15360 is indeed what EXIF metadata says for that shot.
> 
> I've checked quite a number of photos taken with the camera and I'm yet to see one that shows anything different.

Hmm, and the original sample on the RPU also has `15360` white level.
And after conversion with ADC the white level (in `raw prepare` module) is `16360`.
Also, if i enable `auto` mode in `exposure` module, for both the DNG and the ARW, the computed exposure shift is identical.
So i think the white level may actually be correct-_ish_.

bobobo1618:
Aha, I found a less noisy image without people. ISO 100 of a monument with no people in sight.

And okie dokie! I think if the white level is wrong, it should probably be wrong for the RX100M5A as well since they share the same sensor and internals (I think the only difference is the lens?).

LebedevRI:
> And okie dokie! I think if the white level is wrong, it should probably be wrong for the RX100M5A as well since they share the same sensor and internals (I think the only difference is the lens?).

To be noted, since the white level **is** specified in the metadata, and it is **not** being read,
then the white level (as reported by the library) is _probably_ wrong,
even though there is no evidence (samples) of that yet.
This hardcoding is not really the best choice, it is just what mostly worked so far.

PR #163, opened 2019-01-08
Fix early return from IiqDecoder::CorrectPhaseOneC

RobbieAB:
Alter behaviour of IiqDecoder::CorrectPhaseOneC when encountering tag
0x431, to no longer return from the function allowing for checking other
tags.

### DIFF
--- a/src/librawspeed/decoders/IiqDecoder.cpp
+++ b/src/librawspeed/decoders/IiqDecoder.cpp
@@ -238,12 +238,10 @@ void IiqDecoder::CorrectPhaseOneC(ByteStream meta_data, uint32 split_row,
 
     switch (tag) {
     case 0x431:
-      if (!iiq.quadrantMultipliers)
-        return;
-
-      CorrectQuadrantMultipliersCombined(meta_data.getSubStream(offset, len),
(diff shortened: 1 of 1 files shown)


LebedevRI (commented):
The question here being, do we want to keep looking for other tags, or do we want to keep looking for other tags AND fix tags more than once, should it be present more than once?

RobbieAB:
We can't fix bad columns if we don't look for the tags for them. Some tags may also appear more than once with different values, like bad columns.

RobbieAB:
Ok, thinking about it, bad columns is a sub-tag of sensor defects, so I suppose sensor defects possibly should not appear more than once.

I think dealing with bad raw files due to overlapping tags may be a "hard problem" anyway, how do we know the first tag we encounter is the correct one? I suspect in practice we should never encounter a tag twice.

LebedevRI:
> We can't fix bad columns if we don't look for the tags for them.

That's what the pr subject says, yes.

> Some tags may also appear more than once with different values, like bad columns.

And that was precisely the question i asked :D

LebedevRI:
Thank you.

PR #164, opened 2019-01-27
Gh157 phase one bad column

RobbieAB:


LebedevRI:
Looks like the ci is finally happy.
But i want to get the develop branch back to shape, and propagate it to darktable's master/26x branches.
So this will have to wait for a bit (a few days?)

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -390,7 +391,7 @@ void IiqDecoder::correctBadColumn(const ushort16 col) {
         if (dev[max] < dev[i])
           max = i;
       }
-      *mRaw->getData(col, row) = (sum - val[max]) / 3.0 + 0.5;
+      *mRaw->getData(col, row) = std::lround((sum - val[max]) / 3.0);
```
LebedevRI: For future reference, it's best (and possible!) to avoid floating point here completely, without sacrificing the answer:
```c++
#include <iostream>
#include <string>
#include <cmath>
#include <cassert>

int xx(int x) {
  return std::lround(x / 3.0);
}

int yy(int x) {
  return (((10 * x) / 3) + 5) / 10;
}


int main()
{
  for(int i = 0; i < 214748364; i++)  // real x will always be much smaller, 196605 at max (3*UINT16_MAX)
    assert(xx(i) == yy(i));
  assert(false && "done");
  return 0;
}
```
```
$ clang++ -fsanitize=address,undefined,integer test.cpp 
$ ./a.out 
a.out: test.cpp:20: int main(): Assertion `false && "done"' failed.
Aborted
```
Somehow i don't think that will be slower than `lround()` call (https://godbolt.org/z/8DxzSi)

PR #165, opened 2019-02-05
Travis-CI - older clang

LebedevRI:


LebedevRI:
@libraw this should address the "older clang version" question.

There is no CI coverage for OSX though, i haven't checked if they
increased throughput of OSX builds since last time i checked.

It won't stay this way forever, it's just too ugly to live.
I suspect i might be bumping to C++17 (gcc7, clang4.0, xcode9.0) maybe in +1 year (a.k.a 2020)

PR #171, opened 2019-03-14
various fixes to enhance portability

rabauke:
The applied fixes and modifications allow to compile rawspeed not only by using gcc and the clang compiler but also by the Intel C++ compiler version 19.0.2.187.

### DIFF
--- a/src/librawspeed/common/Memory.h
+++ b/src/librawspeed/common/Memory.h
@@ -35,8 +35,12 @@ namespace rawspeed {
 
 // coverity[+alloc]
 void* alignedMalloc(size_t size, size_t alignment)
+#if defined(__ICC) || defined(__ICL) 
+    __attribute__((malloc, warn_unused_result, alloc_size(1)));
+#else
     __attribute__((malloc, warn_unused_result, alloc_size(1), alloc_align(2),
                    deprecated("use alignedMalloc<alignment>(size)")));
+#endif
 
 template <typename T, size_t alignment>
 // coverity[+alloc]
--- a/src/librawspeed/common/Spline.h
+++ b/src/librawspeed/common/Spline.h
@@ -147,6 +147,20 @@ public:
 
   std::vector<Segment> getSegments() const { return segments; }
 
+private:
+  template<typename U>
+  typename std::enable_if<std::is_floating_point<U>::value, double>::type clamp(double x) const {
+    return x;
+  }
+
+  template<typename U>
+  typename std::enable_if<not std::is_floating_point<U>::value, double>::type clamp(double x) const {
+     x = std::max(x, double(std::numeric_limits<value_type>::min()));
+     x = std::min(x, double(std::numeric_limits<value_type>::max()));
+    return x;
+  }
+
+public:
   std::vector<value_type> calculateCurve() const {
     std::vector<value_type> curve(65536);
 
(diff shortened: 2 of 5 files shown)


LebedevRI:
Why, what is the motivation?
What's the practical use-case?

rabauke:
The Intel compiler tends to be even more strict than GCC and Clang and produces often faster machine code. Most of my changes just improve conformance with the C++ language, particularly in `PanasonicDecompressorV5.*`. Future versions of GCC or Clang may complain about these issues too.

Travis indicates an issue with GCC-5. I tested the patch with Clang-7, GCC-7 and ICC 19 only. I will fix this soon and update my pull request.

rabauke:
Currently no idea how to fix the GCC-5 issue. Think the compiler is just wrong here. One problem that I try to address in my pull request is that `PanasonicDecompressorV5.h` (development branch) declares 
```
static const PacketDsc TwelveBitPacket;
```
and `PanasonicDecompressorV5.cpp` implements
```
constexpr PanasonicDecompressorV5::PacketDsc
    PanasonicDecompressorV5::TwelveBitPacket =
        PanasonicDecompressorV5::PacketDsc(/*bps=*/12);
```
which is incompatible with the previous declaration and therefore an error. Have not yet found a way to fix this in a way that is compatible with GCC-5 without breaking other things. :-(

rabauke:
What is precisely required to get this compiled via your contiguous integration tools?

LebedevRI:
What i'm saying is, there can be any number of compilers out there,
with varying levels of how broken they are, how strictly they implement any particular standard.

Right now, only clang (+clang-cl, although needs messing with cmake flags), xcode and gcc
are supported.

Any (every new) entry in the support matrix increases the testing time (even just the CI time),
since the bugs of all the compilers aren't identical, further restricts the code, as it can be seen here,
and have been observed previously.

With currently supported compilers, that makes sense - gcc&clang (xcode is clang) is the compiler in,
dunno, 85% of cases. They also are trivially available for CI, since they are free and packaged.
That isn't so for ICC. I won't be able to add to CI, i can't even use it locally.
Which means, even if you get to to build for you with ICC, it won't stay that way, it *will* regress,
, it won't be caught automatically, and i won't be able to fix it. And you are back to square one.

So i'm asking: **why** do i/we care for ICC? **Why** that price should be paid? It's not one-time cost.
Is there a project/software that wants to use the library, one that is specifically built with ICC? etc

rabauke:
Let me explain it in this way: The changes in my pull request are (up to one exception) not specific at all to the Intel compiler. They address code segments where the rawspeed source does not conform to the rules of the C++ programming language or has other issues (dead code, numerical underflow). Thus these changes will improve the general code quality. As a side effect rawspeed also compiles even with compilers which are not officially supported. At no time I did propose that the Intel compiler should become an officially supported compiler.

According to my experience, issues with standard conformance that pop up with one compiler usually become an issue with compilers from other vendors as well (sooner or later) when an updated compiler is released.

LebedevRI:
> Let me explain it in this way: The changes in my pull request are (up to one exception) not specific at all to the Intel compiler. They address code segments where the rawspeed source does not conform to the rules of the C++ programming language

Ok, so your claim is that the issues being addressed by this PR are non-standard conformant code
of the library that is currently being accepted by the existing compilers, correct?
Then please, _after_ the CI passes:
1. Squash commits into one
2. Partition it into one commit per ICC warning instance
    (i.e. if a warning X fired twice, that should be two commits),
    and do include that warning into each commit message.
3. For each of these commits, apply creduce, reduce the minimal source that still exhibits
   the issue (accepted by clang and/or gcc, not accepted by icc),
   and file a bug against ICC **and** clang **and** gcc,
   include links to these 3 issues into commit messages.

That is what i have been doing with compiler bugs i have been encountering.

LebedevRI:
Thank you for looking into this!
It looks like i should close this for now, sorry.
if you (or anyone else) wishes to try this again (for any compiler, not specifically ICC),
please do this in smaller steps, following https://github.com/darktable-org/rawspeed/pull/171#issuecomment-473663049 
Thanks!

PR #174, opened 2019-05-05
SAMSUNG NX500 sensor crop fix

pmjdebruijn:
The SAMSUNG NX500 sensor has 16 extra pixels on the bottom and right of
the sensor that do have valid image data.

The camera generated JPEGs however do not use this area, and therefore
our RAW processed images are slightly misaligned (off center), while
subtle, it's still best avoided.

Another side effect is that the new 6480x4320 crop, has a perfect 3:2
aspect ratio, as where the old 6496x4336 crop, was slightly off.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -8150,7 +8150,7 @@
 			<Color x="0" y="1">BLUE</Color>
 			<Color x="1" y="1">GREEN</Color>
 		</CFA>
-		<Crop x="0" y="0" width="0" height="0"/>
+		<Crop x="0" y="0" width="-16" height="-16"/>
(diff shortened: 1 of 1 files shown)


LebedevRI:
@pmjdebruijn friendly reminder that raw sample set from that camera is still wanted by RPU :)

The sample situation with Samsung cameras is extremely worrying,
i'm still "unable" to really work on the `SamsungV{0,1,2}Decompressor` because of this.

LebedevRI:
@pmjdebruijn bump :)

LebedevRI:
Disregard, samples finally acquired via @emanuele @ https://github.com/Beep6581/RawTherapee/issues/3000#issuecomment-570769010

PR #187, opened 2019-06-27
Adds support for the Fuji XF10

bastibe:
The entry is copied from the X-A5, which uses the same sensor. It seems to do the trick on my machine.


LebedevRI (approved):
Yep, thank you.
In this case this appears to be correct.

PR #198, opened 2019-07-06
Fujifilm RAW Exposure Bias

denzp:
The change implements reading of a Fujifilm's proprietary tag with an exposure difference between RAF and camera-produced JPEGs.

References:
- https://redmine.darktable.org/issues/11370
- https://sno.phy.queensu.ca/~phil/exiftool/TagNames/FujiFilm.html


LebedevRI:
I'm not confident this belongs in this library.
It's just EXIF tag reading, and this tag isn't essential for the raw decoding, only for it's ideal display.

denzp:
Oh, I see. It's not a usual EXIF tag, though. The required data is stored in an RAF Metadata, and at first, I thought it's out of scope for other libraries such as `exiv2`.

After some investigation, I found that there is an [ongoing work](https://github.com/Exiv2/exiv2/pull/810) to parse the metadata, so it can be later extended to get the value of the **RAW Exposure Bias** tag.

Anyway, thanks for reviewing!

> It's just EXIF tag reading, and this tag isn't essential for the raw decoding, only for it's ideal display.

And to reduce the amount of routine work correcting exposure for hundreds of Fuji images with high ISO or "dynamic range" in Darktable :wink:

PR #199, opened 2019-07-25
Support for DefaultUserCrop in DNG files.

jenshannoschwalm:
Some cameras offer cropped images by using the DefaultUserCrop tag (DNG spec 1.4).
The file contains the whole sensor data but the used area of the sensor defined
by ActiveArea and DefaultCropOrigin/Size is further restricted.

This is implemented at the rawspeed side by:
a) adding the tiff tag as specified -- TiffTag.h
b) adding a float[4] array to class ImageMetaData -- RawImage.h
c) ImageMetaData::ImageMetaData sets default as whole image -- RawImage.cpp
d) DngDecoder::handleMetadata checks for this tag, does some security checks
   and possible sets the UserCrops std::array

I know that rawspeed does not want to use metadata, i hope you accept this pr as an
enhancement and don't say no because of a violation of that policy.

PRs for dt include already fully working support for the crop module and next will be adding a search
for these pics in collections. exiv2 does not support this tag yet, also we would have to blow up the database further.


LebedevRI:
I think i'm not understanding the bigger picture here.

darktable reads that metadata in `exif.cc`, e.g.:
https://github.com/darktable-org/darktable/blob/8ffe190d1225fad4172c61546fa1219b6e7b1c6e/src/common/exif.cc#L602-L606
Where `FIND_EXIF_TAG` is a wrapper for `dt_exif_read_exif_tag()`:
https://github.com/darktable-org/darktable/blob/8ffe190d1225fad4172c61546fa1219b6e7b1c6e/src/common/exif.cc#L475-L488
That calls `Exiv2::ExifData::findKey()` with `Exiv2::ExifKey`.
But as per https://www.exiv2.org/doc/classExiv2_1_1ExifKey.html unsurprisingly has a lot of constructors, not just the one to construct from string, but e.g.
```
ExifKey (uint16_t tag, const std::string &groupName)
--
Constructor to create an Exif key from the tag number and group name. More...
```
So it does not matter whether exiv2 knows the name of the tag or not, you still can get it.
Have you tried using that one?

jenshannoschwalm:
Thanks a lot for your support and discussion.

Indeed i was not aware about other ways exiv2 can do a search. I will learn about that.

About the "bigger picture" or maybe "user scenario"

Some cameras offer "cropped modes" like offering 1:1 of 16:9 ratios, normally both raw and jpeg files will only contain the used sensor data for that crop.

For DNGs this is/might be different, one example is the Leica Q2 with a fixed 28mm lens. There we have full images and images shot in "cropped to 35(or 50/75)mm, the jpegs are cropped, the DNGs are not cropped - they have the full sensor data - but have the DefaultUserCrop" information.

The default then should be to crop/clip the image to the parameters told when using the clipping module but you will be able to change the area. (As i read from descriptions thats what lightroom does). Anyway thats a nice feature.

So the point is - how do we get & keep that information available for the clipping module.
How can that be done?
a) Sure - we can test in exif.cc, (only done when importing) - then we would have to keep that information in the database. (Something i wanted to avoid but maybe that would be the correct way? At least it would mean lots of code in the database housekeeping.)
b) We could test in exif.cc and keep a flag in the database (there are some free bits), so when opening a picture we can test for the flag and look for the exact cropping data if necessary.
c) Or we can ask for rawspeed "assistance" with a neglectable overhead (but of course this is a feature that you might not think to be a problem for rawspeed)

I hope this make my ideas more understandable. Please let me know what you think...

BTW so far i have no idea why the travis check failed ...

LebedevRI:
> Thanks a lot for your support and discussion.
> 
> Indeed i was not aware about other ways exiv2 can do a search. I will learn about that.

Yay

> About the "bigger picture" or maybe "user scenario"
> 
> Some cameras offer "cropped modes" like offering 1:1 of 16:9 ratios, normally both raw and jpeg files will only contain the used sensor data for that crop.
> 
> For DNGs this is/might be different, one example is the Leica Q2 with a fixed 28mm lens. There we have full images and images shot in "cropped to 35(or 50/75)mm, the jpegs are cropped, the DNGs are not cropped - they have the full sensor data - but have the DefaultUserCrop" information.
> 
> The default then should be to crop/clip the image to the parameters told when using the clipping module but you will be able to change the area. (As i read from descriptions thats what lightroom does). Anyway thats a nice feature.

Thank you for the explanation.
I do understand the user side of the issue.
The question was specifically about the implementation approach.

> So the point is - how do we get & keep that information available for the clipping module.
> How can that be done?
> a) Sure - we can test in exif.cc, (only done when importing) - then we would have to keep that information in the database. (Something i wanted to avoid but maybe that would be the correct way? At least it would mean lots of code in the database housekeeping.)
> b) We could test in exif.cc and keep a flag in the database (there are some free bits), so when opening a picture we can test for the flag and look for the exact cropping data if necessary.

It's been a while since i last touched dt's code.
Back then even all these things that you get from rawspeed were cached in database.
Things may or may not have changed since then.
I'm going to guess they didn't break _that_ yet, so i'd guess that is still the case.
I.e. even if you get this info from rawspeed, you will still need to bloat the database.

So yeah, the question i'm asking is, given that it will end up in database regardless of
how it is acquired, why is this must not be acquired trough dt's `exif.cc`, but through rawspeed?

> c) Or we can ask for rawspeed "assistance" with a neglectable overhead (but of course this is a feature that you might not think to be a problem for rawspeed)
> 
> I hope this make my ideas more understandable. Please let me know what you think...

> BTW so far i have no idea why the travis check failed ...

There is no such ternary comparison in C/C++.
While it isn't an invalid code, it doesn't do what you think it does.
If you go to specific build in travis, it says as much:
https://travis-ci.org/darktable-org/rawspeed/jobs/563634931#L815

jenshannoschwalm:
> t's been a while since i last touched dt's code.
Back then even all these things that you get from rawspeed were cached in database.

NO, this is not correct. dt does *not* have this in the database, neither in images nor in legacy_presets. "Crop" in "images" does not refer to this of course. And if you think of the "crop" data (saved to the sidecar file) used by the clipping module, that is the "clipping" range selected by the user when he/she changed something.

What we want is the "default clip area" for that module. See `void reload_defaults(dt_iop_module_t *self)` in dt's iop/clipping.c
That so far sets the default clipping rect to 0,0,1,1
After this PR it can be
` = (dt_iop_clipping_params_t){ 0.0f, img->user_crop[0], img->user_crop[1], img->user_crop[2], img->user_crop[3], 0.0f, 0.0f,  0.2f, 0.2f, 0.8f, 0.2f, 0.8f, 0.8f, 0.2f, 0.8f, 0,    0,    FALSE, TRUE, -1,   -1 };`

> I.e. even if you get this info from rawspeed, you will still need to bloat the database.

Again it's a NO.
`dt_imageio_retval_t dt_imageio_open_rawspeed` always makes sure the data in `dt_image_t *img` is fully available. So we just have to copy the new `td::array<float, 4> UserCrops;` to the  `dt_image_t` struct. Voila!

About bloating the database.
Yes i indeed will bloat it, exactly by one bit per picture in the flags field. Why?
This will allow to add a 'collect all usercropped images' feature.

About my two NOs. I will be in moscow late october for some days, we meet and you have free drinks if you prove me wrong :-)

LebedevRI:
> > t's been a while since i last touched dt's code.
> > Back then even all these things that you get from rawspeed were cached in database.
> 
> NO, this is not correct. dt does _not_ have this in the database, neither in images nor in legacy_presets. "Crop" in "images" does not refer to this of course. And if you think of the "crop" data (saved to the sidecar file) used by the clipping module, that is the "clipping" range selected by the user when he/she changed something.

Please refer to [`dt_image_cache_allocate()`](https://github.com/darktable-org/darktable/blob/f15bc99f9ce948ed693c48ec3dd4040489de211f/src/common/image_cache.c#L29), [`dt_image_cache_write_release()`](https://github.com/darktable-org/darktable/blob/f15bc99f9ce948ed693c48ec3dd4040489de211f/src/common/image_cache.c#L215).
Are you telling me that does not read/write e.g. `crop` from/to `exif_crop` field?

jenshannoschwalm:
No i do not!
Using exif_crop looks indeed promising. BUT

1. DefaultUserCrop specifies all 4 borders individually. Ok, this will often be around the centre and so could be replaced with one parameter but we can't be sure.
2. exif_crop is calculated by FocalLength and FocalLengthIn35mmFilm. I checked my dt database with my and many testfotos from raw.pixls.us, some examples: for Pana LX3 this will be ~4, Nikon3400 gives 1.5 ... So we can't simply use that or the crop in the database
3. the crop_x and friends in `dt_image_t` are used otherwise.

LebedevRI:
That field was only an example; i would recommend to explore dt-centric solution via exif.cc & db.
So far i have not read any concise reasoning as to why that would work / is not the right solution.

jenshannoschwalm:
Let me try it once more.

It works, it you want to test you could use https://github.com/jenshannoschwalm/darktable/tree/usercrop. I use this without flaws.

The DefaultUserCrop tag is something not implemented at all in dt.
So far we have the sensor size and the cropped region defined by 'magic' via the cameras.xml information or the data supplied via rawspeeds DNGDecoder ACTIVEAREA, DEFAULTCROPORIGIN and DEFAULTCROPSIZE. So we know the complete usable area within the supplied data.

This pr also looks for DefaultUserCrop.
It
- would not hurt
- could be easily done without bloating the database
- would offer the user to crop the raw image exactly as the corresponding jpeg
- could be used by other decoders if we know the usable area and what should be offered for further clipping.

I could do a dt-centric solution via exif.cc & db but that would mean either a db update for all users - something i strictly would like to avoid - or fiddling with exif data in the clipping module.

LebedevRI:
The darktable database questions are questions for darktable.
So far all the reasoning provided here in justification can be applied to
many other fields (lens name, camera orientation, ...), and that will 
effectively result in bloating this library with functionality that belongs to Exiv2.

https://github.com/darktable-org/rawspeed/blob/develop/README.rst
> RawSpeed does NOT…
> * read metadata information, beside whitebalance information.

Other than desire to not touch darktable's database,
what are the technical reasons to not do this in darktable?
Does this somehow depend on the image being decoded?

Once again, i would recommend to try to implement this via `exif.cc` approach,
that will either just work, or will provide the justification as to why this needs to be done here.

jenshannoschwalm:
Ok, i will have to try exiv2 magic...
keep going, all the best to you

PR #205, opened 2019-11-09
Fix bounds check.

mcourteaux:
This seems more logical.


LebedevRI:
Thanks.
I've been slowly attempting to kill usages of `RawImageData::getData()`,
but this does seem logical, yes.

Issue #215, opened 2020-01-05
DNG BlackLevel handling when cpp != 1


LebedevRI:
@LebedevRI @alexvanderberkel  I got darktable decoding dual pixel dng files.

There is one more thing in rawspeed that "requires relaxing". `bool DngDecoder::decodeBlackLevels(const TiffIFD* raw) {` in DngDecoder.cpp should accept cpp==2 too.

So i suggest it to be
``` 
 if ((mRaw->getCpp() < 1) || (mRaw->getCpp() > 2))
    return false;

```
Can you agree here? And could you modify this or do you want me to do a pr?
BTW it would be nice to push current rawspeed to dt :-)

_Originally posted by @jenshannoschwalm in https://github.com/darktable-org/rawspeed/issues/214#issuecomment-570887227_

More specifically, as per https://www.adobe.com/content/dam/acom/en/products/photoshop/pdfs/dng_spec_1.4.0.0.pdf, all that blacklevel info is not at all specific to cpp=1/cpp=2.
See `BLACKLEVELREPEATDIM`, `BLACKLEVEL`, `BLACKLEVELDELTAV`, `BLACKLEVELDELTAH`.
Which only highlights how broken the current 4-channel `mRaw->blackLevelSeparate` is.

jenshannoschwalm:
Ok, i agree on missing proper blacklevel support.

You know http://wwwimages.adobe.com/www.adobe.com/content/dam/acom/en/products/photoshop/pdfs/dng_spec_1.5.0.0.pdf ?
There is even more to be worked on.

Do you have priority list of more "things-to-be-fixed" regarding dng?

LebedevRI:
> Ok, i agree on missing proper blacklevel support.
> 
> You know http://wwwimages.adobe.com/www.adobe.com/content/dam/acom/en/products/photoshop/pdfs/dng_spec_1.5.0.0.pdf ?

Oh, nice, i should update my bookmark then.

> There is even more to be worked on.
> 
> Do you have priority list of more "things-to-be-fixed" regarding dng?

I don't think i have any list regarding DNG. (well, other than the general issue of 'multi-image raw support')

madwyn:
Thanks for pointing me to the right direction. I'm wondering will this be fixed in the `develop` branch?

RGB DNG files have `cpp=3` and they also can have black levels label `50714 Black Level`.

It's confusing that black level handling needs to consider `cpp`, I can understand `<1` or `>3` looks "bad".

[...]

PR #216, opened 2020-01-11
DngDecoder should return blacklevels also for 2-component images. (#215)

jenshannoschwalm:
These are observed in DNG's converted from dual pixels CR3 raws.
Wrong colors result from non-applying the blacklevels.

Second part to fix issue darktable-org#214
See also: https://github.com/darktable-org/rawspeed/issues/215

### DIFF
--- a/src/librawspeed/decoders/DngDecoder.cpp
+++ b/src/librawspeed/decoders/DngDecoder.cpp
@@ -689,7 +689,7 @@ bool DngDecoder::decodeBlackLevels(const TiffIFD* raw) {
   if (!raw->hasEntry(BLACKLEVEL))
     return true;
 
-  if (mRaw->getCpp() != 1)
+  if ((mRaw->getCpp() < 1) || (mRaw->getCpp() > 2))
     return false;
 
   TiffEntry* black_entry = raw->getEntry(BLACKLEVEL);


LebedevRI:
As i've noted in https://github.com/darktable-org/rawspeed/issues/215, i'm not confident that this is as simple as that,
that we aren't simply exchanging lack of blacklevel info with incorrect blacklevel info.

jenshannoschwalm:
What we miss is proper BlackLevel for both images. But when reading the 1.5 specs it says on page 26
"The values are stored in row-column-sample scan order." Doesn't this imply, that at least for the first image the read-in values are correct?

LebedevRI (changes requested):
I'm not very sure that instead of not producing any info,
producing some likely known-broken info is good/better.

All that code so far clearly was designed for cpp=1,
i don't really buy that it is all magically correct for the cpp!=1 case.

In particular, `BLACKLEVEL` tag is defined in https://www.adobe.com/content/dam/acom/en/products/photoshop/pdfs/dng_spec_1.4.0.0.pdf as 
> Count BlackLevelRepeatRows * BlackLevelRepeatCols * SamplesPerPixel

Now SamplesPerPixel (cpp) is 2.
How are we accounting for that?
How //should// we be accounting for that?

jenshannoschwalm:
You cited just one part of the specs

> Count BlackLevelRepeatRows * BlackLevelRepeatCols * SamplesPerPixel

Refers to the count and here you are are right. We //should// in fact have not only 4 values but more as defined above. Right. We can **not** surely assume that blacklevel correction is the same for image2. Right again.

But the refs also says

> This tag specifies the zero light (a.k.a. thermal black or black current) encoding level, as a repeating pattern. The origin of this pattern is the top-left corner of the ActiveArea rectangle. The values are stored in row-column-sample scan order.

This means that black levels read-in as in current rawspeed code are absolutely correct for image1. So if i stay with image1 everything is fine. Of course there could be dual-raw where the blacklevels are different so for those image2 would not be correctly covered by this pr.

jenshannoschwalm:
@LebedevRI No chance to get this merged?

What is the problem? We have the number of blacklevel values defined by BlackLevelRepeatDim, this code is not touched. The (default) first 4 blacklevels are absolutely the same for cpp=1 or cpp=2 and are valid for the first sample.

See the 1.5 dng specs page 25 `The values are stored in row-column-sample scan order`

We would not get correct values for the second sample but we don't want to access that at all.

What would be required ...

LebedevRI:
Now that there's a non-obscure working (affected) example (#225),
i think progress can be made here.
I think there may be some preparatory problems to handle before that:
1. The fact that we currently represent blacklevels as some abstract fixed-size 4-channel array is just broken.
2. Should we really store them as `float`s? Perhaps it should be rational.

jenshannoschwalm:
I confess i don't understand how this could help. Of course we have rationals that can't be expressed as a float or double but that can't be what you are hinting at.

BlackLevel is always SHORT, LONG or RATIONAL for dng.

Could you enlighten me?

LebedevRI:
I'm not saying it _would_ help, i'm just wondering if it would be more correct to not use `float` there.

PR #217, opened 2020-01-11
Fixing FixBadPixelsList for DNG RawImageDataFloat images

jenshannoschwalm:
There have been discussions about HDRmerged images showing strange artefacts, that are difficult to explain just by applying a wrong FixBadPixelsList.

See darktable-org/darktable#3510 and  jcelaya/hdrmerge#197

Instead of my first assumption this seems to originate from a wrong RawImageDataFloat::fixBadPixel

@heckflosse , @KristijanZic

### DIFF
--- a/src/librawspeed/common/RawImageDataFloat.cpp
+++ b/src/librawspeed/common/RawImageDataFloat.cpp
@@ -296,48 +296,50 @@ void RawImageDataFloat::fixBadPixel(uint32_t x, uint32_t y, int component) {
   std::array<float, 4> weight;
 
   uint8_t* bad_line = &mBadPixelMap[y * mBadPixelMapPitch];
+  // We can have cfa or no-cfa for RawImageDataFloat
+  int step = isCFA ? 2 : 1;
 
   // Find pixel to the left
-  int x_find = static_cast<int>(x) - 2;
+  int x_find = static_cast<int>(x) - step;
   int curr = 0;
   while (x_find >= 0 && values[curr] < 0) {
     if (0 == ((bad_line[x_find>>3] >> (x_find&7)) & 1)) {
-      values[curr] = (reinterpret_cast<float*>(getData(x_find, y)))[component];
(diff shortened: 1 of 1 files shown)


LebedevRI:
I agree that this seems to be the proper fix, but then it matches the `uint16_t` implementation,
which highlights why these bugs are here - code duplication.
https://github.com/darktable-org/rawspeed/blob/198952f61c28f7bf4496e5bde2ffddcbf2363a6d/src/librawspeed/common/RawImageDataU16.cpp#L376-L471

Can you rebase this to not include the dng commit? I'd merge this then.

jenshannoschwalm:
I haven't rebased anything yet, could you explain in short?

heckflosse:
About badpixel interpolation (on raw data) maybe have a look at the rt code

for bayer:
https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/badpixels.cc#L66

for xtrans:
https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/badpixels.cc#L288

Both are also well documented in code (which I usually do not)

LebedevRI:
Licenses are incompatible though

fanckush:
@jenshannoschwalm So the issue is `RawImageDataFloat::fixBadPixel` and not HDRMerge? or was it simply both together? I guess what I'm trying to say: is HDRMerge still supposed to fix anything regarding BadPixels or is this sufficient?

jenshannoschwalm:
@fanckush in fact it's two problems. One was the one fixed here resulting in those strange artefacts. **But** of course HDRMerge right now creates merged images with a simply wrong opcode list. It should do the opcode-list work for every image before merging to gain best quality but also must not use the opcode list of one image for the merged one. So some homework for you to be done :-) Most people will not recognize the remaining problem on your side yet ...

LebedevRI:
> Licenses are incompatible though

Are you saying that it is okay to copy-paste gpl3 code into lgpl3 library?

PR #219, opened 2020-01-21
Enable run from binary dir

codingdave:
When I was trying to create out-of-binary-tree runs (be able to run darktable from the build directory without installing to enhance IDE experience and benefit from integrated cmake tools and thererlike) I figured out dt would not load because it was hanging in rawspeed. The reason was that I was missing a file next to the lib, which I did not yet copy.

This MR will copy the file to the build-tree. It does not cost much time but you might argue that it is not the task of the library to copy its dependencies during build time as it could be done by the integrator. This decision is up to you. However, having this PR would make out-of-binary-tree runs more easy to achive for every integration.

Kind regards
David


data/CMakeLists.txt
```diff
@@ -8,5 +8,6 @@ if(USE_XMLLINT AND XMLLINT_FOUND)
 endif()
 
 if(WITH_PUGIXML)
+  FILE(COPY cameras.xml showcameras.xsl DESTINATION ${DARKTABLE_DATADIR}/rawspeed)
```
LebedevRI, line 11: What's `DARKTABLE_DATADIR`?
If rs is being built outside of darktable, this now tries to copy a file into root file system directory.

LebedevRI (commented):
I'm very much sympathetic to running without installing,
but i'm thinking this should be solved on darktable's side.

data/CMakeLists.txt
codingdave: yeah, sorry, my bad. I have to work on that.

codingdave: I will remove that PR because the lib (for dt) is built in (CMAKE_CURRENT_BINARY_DIR) ./build/lib/darktable/rawspeed/data/rawspeed/cameras.xml but darktable expects to find it in build/share/darktable/rawspeed/cameras.xml. So the copying needs to take place outside the CMakeLists.txt scripts from rawspeed.

codingdave:
see note above

PR #221, opened 2020-01-21
Fix vscode cmake issue

codingdave:
Using Visual Studio Code with the cmake extension does not work with rawspeed. The reason is that the BUILD_TYPE is changed and such the cmake extension gets confused. You will find this error in the console:

[cms-driver] Found no matching code model for the current build type. This shouldn't be possible

See this issue here: https://github.com/microsoft/vscode-cmake-tools/issues/774

This patch changes the file `build-type.cmake` to not modify the BUILD_TYPE but instead do its sanity checks on another copy of it. I hope it does not break the current logic, at least it makes rawspeed work with VS Code and the cmake extension work flawlessly.

Kind regards
David


codingdave:
Closing this one because its on the wrong branch - contains merge commits. The new one is here: https://github.com/darktable-org/rawspeed/pull/222

PR #222, opened 2020-01-22
do not modify CMAKE_BUILD_TYPE to not confuse IDE

codingdave:
(This is the same but fixed PR as https://github.com/darktable-org/rawspeed/pull/221)

Using Visual Studio Code with the cmake extension does not work with rawspeed. The reason is that the BUILD_TYPE is changed and such the cmake extension gets confused. You will find this error in the console:

[cms-driver] Found no matching code model for the current build type. This shouldn't be possible

See this issue here: microsoft/vscode-cmake-tools#774

This patch changes the file build-type.cmake to not modify the BUILD_TYPE but instead do its sanity checks on another copy of it. I hope it does not break the current logic, at least it makes rawspeed work with VS Code and the cmake extension work flawlessly.

Kind regards
David


LebedevRI:
So TLDR is that the cmakefiles must not modify `CMAKE_BUILD_TYPE`, but keep it
exactly as it was set? If that is the only thing i suspect it could be okay, but then you also need to
fix every place where cmakelists relied on the canonicalization you are removing.

cmake/compiler-flags.cmake
```diff
@@ -23,10 +23,10 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 include(debug-info)
 
-if(CMAKE_BUILD_TYPE STREQUAL "RELEASE")
+if(CMAKE_BUILD_TYPE STREQUAL "Release")
```
LebedevRI, line 26: Note that this is not correct fix, this will now errneously not match `rELEASE` string

LebedevRI (changes requested):

codingdave:
> So TLDR is that the cmakefiles must not modify CMAKE_BUILD_TYPE, but keep it
exactly as it was set?

Yes, that is correct. 

> If that is the only thing i suspect it could be okay, but then you also need to
fix every place where cmakelists relied on the canonicalization you are removing.

With the last commit I have been only looking at Release, Debug, and RelWithDebInfo. I did not change COVERAGE, SANITIZE, TSAN, or FUZZ.

cmake/compiler-flags.cmake
codingdave: ok, I will check tomorrow. Thanks for the feedback.

codingdave: Hmm, how come you say `rELEASE` is a valid input? Looking at https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html:
> Possible values are empty, Debug, Release, RelWithDebInfo and MinSizeRel.

LebedevRI: case-insensitivity is how it is currently handled, it should stay that way be it either via
the LHS of the diff, or by fixing the every CMAKE_BUILD_TYPE-dependent check
to be case-insensitive.

codingdave:
Please read the commit message from: removes CMAKE_BUILD_TYPE_UPPERCASE for 2 reasons: f7fafb5

cmake/compiler-flags.cmake
codingdave: ok. please read my last commit message. If you persist on making it case-insensitive I will see how I can achieve that goal.

codingdave:
Why are travis and appveyor failing? Will they pass when my patches are taking care of the case-insensitivity?

LebedevRI:
Anything else?

codingdave:
Thank you @LebedevRI, I can confirm that the issue has been fixed. 

Kind regards
David

PR #236, opened 2020-06-28
Added basic support for Sony Alpha 9M2 (ILCE-9M2)

nixwieweg:
Sensor and RAW File Format is same as ILCE-9 -> so all values copied. Changed BlackLevel to value read from exiv2 after conversion. Have checked WhiteLevel.


LebedevRI:

PR #240, opened 2020-09-10
cmake: fix incorrect usage of CACHE variables

parafin:
* setting CACHE variable recursively will keep adding parameters on cmake
  re-runs
* updating cached value of CMAKE_C_FLAGS and friends breaks the whole
  logic of it being passed on from a user initially and then necessary
  flags added on top, specifically it circumvents the
  `set(CMAKE_C_FLAGS_SAVE "${CMAKE_C_FLAGS}")`
  `do_stuff()`
  `set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS_SAVE}")`
  trick utilized in several places (e.g. in darktable) when cmake is
  re-run
* that still leaves a general problem of updating global variables from
  inside the library, probably it should be done only at the top level
  of application cmake files, not ad-hoc like that, instead targets and
  other modern cmake features should be used

For context see darktable-org/darktable#6258 - LTO flags leak to global darktable scope, which I don't think is intended (or is it?).

There is probably more of such incorrect usage, e.g. check LLVMOpenMP.cmake. Basically every usage of `set(... CACHE ... FORCE)` should be suspect.

### DIFF
--- a/cmake/compiler-flags.cmake
+++ b/cmake/compiler-flags.cmake
@@ -50,17 +50,13 @@ ENDIF()
 # NOT CMAKE_STATIC_LINKER_FLAGS
 SET(CMAKE_SHARED_LINKER_FLAGS
     "${CMAKE_SHARED_LINKER_FLAGS} ${linkerflags}"
-    CACHE STRING "" FORCE )
+    )
 SET(CMAKE_EXE_LINKER_FLAGS
     "${CMAKE_EXE_LINKER_FLAGS} ${linkerflags}"
-    CACHE STRING "" FORCE )
+    )
 SET(CMAKE_MODULE_LINKER_FLAGS
     "${CMAKE_MODULE_LINKER_FLAGS} ${linkerflags}"
-    CACHE STRING "" FORCE )
-MARK_AS_ADVANCED(
-    CMAKE_SHARED_LINKER_FLAGS
-    CMAKE_EXE_LINKER_FLAGS
-    CMAKE_MODULE_LINKER_FLAGS )
+    )
 
 if(RAWSPEED_ENABLE_LTO)
   if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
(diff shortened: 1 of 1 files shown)


LebedevRI:
Right. This makes sense, thanks.

> For context see [darktable-org/darktable#6258](https://github.com/darktable-org/darktable/issues/6258) - LTO flags leak to global darktable scope, which I don't think is intended (or is it?).

Hm. I don't recall it being intentional, but does building e.g. darktable without LTO,
but with `-DRAWSPEED_ENABLE_LTO=ON` still work?

parafin:
> > For context see [darktable-org/darktable#6258](https://github.com/darktable-org/darktable/issues/6258) - LTO flags leak to global darktable scope, which I don't think is intended (or is it?).
> 
> Hm. I don't recall it being intentional, but does building e.g. darktable without LTO,
> but with `-DRAWSPEED_ENABLE_LTO=ON` still work?

That's a good question. So without this PR (and doing only one cmake run) `-flto=thin` gets passed down to darktable in linker flags (darktable only saves C flags, see https://github.com/darktable-org/darktable/blob/157a9d92f3abc6a5c6a07eb90c59583da40babc3/src/external/CMakeLists.txt#L44), so rawspeed was compiled with LTO, darktable was compiled without LTO, but everything was linked with LTO. With this PR LTO flags stay local to rawspeed it seems, so only its object files are built with LTO, but darktable is compiled and linked without LTO. I'm not sure if original mixed LTO scenario works. If it does, then this PR breaks RAWSPEED_ENABLE_LTO support in darktable. If it doesn't, then it never really worked (unless you pass down LTO flags using `-DCMAKE_C_FLAGS`). What do you think?
P.S.
In both cases darktable builds and runs fine.

LebedevRI:
Thinking about it, LTO needs to be added to the `PUBLIC` link flags of `rawspeed` build target,
but i guess i'll do that myself, since i don't see any nice way to do that.

parafin:
Please ping me once you do, so I could test it.

LebedevRI:
Done, seemed to work for me (took more than one try), but feel free to double-check.

parafin:
Works on my side too, thanx. Let's hope this mixed LTO setup actually makes sense:) I don't want to enable LTO on darktable itself, some issues are reported from time to time from people building with LTO.

LebedevRI:
> Works on my side too, thanx. Let's hope this mixed LTO setup actually makes sense:)

It does, at least for clang.

> I don't want to enable LTO on darktable itself, some issues are reported from time to time from people building with LTO.

Yeah, i'd be worried if it were any other way. It's rather broken even without LTO.

PR #241, opened 2020-09-27
Canon EOS Rebel T7i: add another alias

Zanhard:
Added another alias for Canon EOS Rebel T7i with Rebel in all capitals (REBEL). Darktable was unable to open the raw camera file as rawpseed was unable to find a corresponding camera due to Rebel being in all capitals. This was reported here: https://github.com/darktable-org/darktable/issues/5973


LebedevRI:
I have thought about this, and let's just go with this for now.

PR #242, opened 2020-10-07
Add Panasonic DC-S5

JosefR:
Just copying the configuration of the S1H/S1 as the S5 is using the same sensor. Works fine for standard resolution files when using the color matrix of the S1H in darktable. Just the sensor-shift high resolution files are not working. They are using a 4:3 aspect ratio but that was still not working when set up in in the config.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -7239,6 +7239,28 @@
 		<Crop x="0" y="0" width="-50" height="0"/>
 		<Sensor black="510" white="16380"/>
 	</Camera>
+	<Camera make="Panasonic" model="DC-S5">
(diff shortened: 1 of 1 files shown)


LebedevRI:
Thank you for the patch.

I just re-checked, and the most recent Adobe DNG Converter (12.4, from August) still doesn't support that camera.
So until it does, we don't have a color matrix, and are therefore stuck.

LebedevRI:
```
[rawspeed] (dc-s5_hr_96MP.RW2) virtual rawspeed::RawImage rawspeed::Rw2Decoder::decodeRawInternal(), line 151: Version 7 is unsupported
```

PR #243, opened 2020-10-15
add nikon z 5 (camera only supports compressed raw)

matt-maguire:
Added support for Nikon Z5 to cameras.xml, based on samples uploaded to raw.pixls.us (refer darktable ticket darktable-org/darktable#6173). This camera doesn't seem to support uncompressed raws, just compressed (lossy & lossless). Black and White points extracted out of both lossy and lossless compressed raw samples are identical, and are also consistent with values in existing Z6 entry. 

Used Adobe DNG converter 12.4.0.555 on macOS.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -2655,6 +2655,28 @@
 			<Hint name="msb_override" value=""/>
 		</Hints>
 	</Camera>
+        <Camera make="NIKON CORPORATION" model="NIKON Z 5" mode="14bit-compressed">
+                <ID make="Nikon" model="Z 5">Nikon Z 5</ID>
+                <CFA width="2" height="2">
+                        <Color x="0" y="0">RED</Color>
+                        <Color x="1" y="0">GREEN</Color>
+                        <Color x="0" y="1">GREEN</Color>
+                        <Color x="1" y="1">BLUE</Color>
+                </CFA>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Please can you help with the test coverage for this camera? I can't really verify the crop otherwise here.
There are several samples on RPU, but they are all somewhat unsatisfactory.

There should be 4 raws (14 bits lossless, 14bit lossy, 12bit lossless, 12bit lossy) of the ~same scene,
with low ISO, of something like daylight landscape.

matt-maguire:
I don't have a copy of this camera, I just used the 4 samples that were uploaded to raw.pixls.us. When I built darktable and opened the NEF files alongside the DNGs, the color tint from the photographed LCD looked the same (not that there is really much color there to compare), but really it would be better to have something more colorful to properly compare. I think the ISO of the photos was 400.

Maybe it would be better to wait for some better samples...

StadtMoewe:
Hi, I´m new here. I have a Nikon Z5 and uploaded some pictures of my motobike to https://raw.pixls.us. I did 12bit and 14bit with both compressions.
I hope this will help to bring support for my camera to darktable. When you need more or different pictures just tell me.

LebedevRI (approved):

PR #244, opened 2020-10-17
Added/updated black/white point data for Canon EOS 7D, 7D mk2, and 10D.

wpferguson:
Based on disscussion in https://github.com/darktable-org/darktable/issues/6548#issuecomment-711057304
Also added data for sRaw1 and sRaw2 for 7D and 7D mk2.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -193,7 +193,9 @@
 			<Color x="1" y="1">BLUE</Color>
 		</CFA>
 		<Crop x="64" y="12" width="0" height="0"/>
-		<Sensor black="127" white="4000"/>
+		<Sensor black="130" white="4000"/>
+		<Sensor black="781" white="32768" iso_list="100"/>
+		<Sensor black="255" white="4000" iso_list="1600 3200"/>
 	</Camera>
 	<Camera make="Canon" model="Canon EOS 20D">
 		<ID make="Canon" model="EOS 20D">Canon EOS 20D</ID>
@@ -894,8 +896,9 @@
 			<Color x="1" y="1">GREEN</Color>
 		</CFA>
 		<Crop x="158" y="52" width="5202" height="3464"/>
-		<Sensor black="1025" white="8150" iso_min="12800" iso_max="12800"/>
-		<Sensor black="2050" white="16300"/>
+		<Sensor black="2049" white="15400"/>
+		<Sensor black="2048" white="13600" iso_list="100 125"/>
+		<Sensor black="2048" white="12600" iso_list="160 320 640 1250 2500"/>
 		<BlackAreas>
 			<Vertical x="8" width="156"/>
 			<Horizontal y="32" height="18"/>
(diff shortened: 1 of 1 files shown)


data/cameras.xml
```diff
@@ -193,7 +193,9 @@
 			<Color x="1" y="1">BLUE</Color>
 		</CFA>
 		<Crop x="64" y="12" width="0" height="0"/>
-		<Sensor black="127" white="4000"/>
+		<Sensor black="130" white="4000"/>
+		<Sensor black="781" white="32768" iso_list="100"/>
```
LebedevRI: That 16-bit whitelevel doesn't look right.

data/cameras.xml
```diff
@@ -904,14 +907,16 @@
 	<Camera make="Canon" model="Canon EOS 7D" mode="sRaw1">
 		<ID make="Canon" model="EOS 7D">Canon EOS 7D</ID>
 		<Crop x="0" y="0" width="0" height="0"/>
-		<Sensor black="0" white="44200" iso_list="100 125 160 320 640 1250"/>
-		<Sensor black="0" white="54000"/>
```
LebedevRI, line 908: I'm looking at ISO1250 sample [`raw.pixls.us/Canon/EOS 7D/RAW_CANON_EOS_7D-mraw.CR2` ](https://raw.pixls.us/data/Canon/EOS%207D/RAW_CANON_EOS_7D-mraw.CR2),
and there is no way it should have white level of just `9000` instead of `44200`.

LebedevRI (commented):
I don't recall how i usually got the whitelevel values for subsampled raws, 
but i'm pretty confident that we can't just get them from DNG's.

data/cameras.xml
```diff
@@ -904,14 +907,16 @@
...
+		<Sensor black="0" white="11400"/>
+		<Sensor black="0" white="9850" iso_list="100 125"/>
```
LebedevRI: That being said, (1<<16)/(1<<14) is 4 and 44200/11400 is ~3.8, so i think we just need to multiply the values by 4

data/cameras.xml
wpferguson: I didn't think so either, so I shot another set and it came out the same way.  I looked at the generated DNG and that's what it had.  I just tested these numbers to see what would really happen.  It appears the default (127, 4000) works much better.  I'll change that back.

data/cameras.xml
wpferguson: I'll multiply by 4 and update them.

wpferguson: I'll also shoot a few of the subsampled raws and import them to check and make sure the values make sense.

wpferguson: I shot some mRaw and sRaw.  The 0 to 44000 matched the embedded jpegs best, so I'll set it back to that.

LebedevRI:
I obviously did not manually verify each change, but this seems ok overall i think.
Thank you.

PR #254, opened 2020-11-29
Fujifilm X-S10 camera support

floemker:
Added support for the Fujifilm X-S10 for uncompressed raw files and for
lossless compressed raw files. The X-S10 is using the same sensor and
CPU as the X-T4.

I took two test shots at every ISO setting. Booth ISO 500 shots had a black
level of 1021. One of the two 10000 ISO shots had a black level of 1023.
ISO settings were 160, 200, 250, 320, 400, 500, 640, 800, 1000, 1250, 1600,
2000, 2500, 3200, 4000, 5000, 6400, 8000, 10000, 12800. I ignored the
extended ISO settings 80, 100, 125, 25600, 51200. I assume these small
black level variations can be ignored?

I uploaded as well three same files to https://raw.pixls.us/, an uncompressed,
a lossy compressed, and a lossless compressed RAF file.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -10276,6 +10276,32 @@
 		<Crop x="4" y="0" width="-52" height="0"/>
 		<Sensor black="1024" white="16383"/>
         </Camera>
+	<Camera make="FUJIFILM" model="X-S10">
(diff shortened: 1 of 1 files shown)


LebedevRI:
Thanks!

PR #260, opened 2021-01-31
FIX DC-GF10 alias.

elizeusdsantos:
After some local tests I got that as a typo in the alias making impossible to open DC-GF10 raw files that was just included.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -6404,7 +6404,7 @@
 		<Crop x="0" y="0" width="-210" height="0"/>
 		<Sensor black="143" white="4095"/>
 		<Aliases>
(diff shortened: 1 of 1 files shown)


data/cameras.xml
```diff
@@ -6404,7 +6404,7 @@
 		<Crop x="0" y="0" width="-210" height="0"/>
 		<Sensor black="143" white="4095"/>
 		<Aliases>
-			<Alias>GF10</Alias>
+			<Alias>DC-GF10</Alias>
 			<Alias>GF90</Alias>
```
LebedevRI, line 6408: I'm guessing `GF90` is also a typo?

PR #269, opened 2021-05-05
Support for Leica D-LUX 7

jenshannoschwalm:
This is basically the same camera as Pana LX100 Mark II, just some other housing. So:

- copy in cameras.xml
- add "LEICA CAMERA AG" as a valid make in the decoder. Almost all recent firmware from Leica
  uses this btw.


jenshannoschwalm:
@LebedevRI anything more required or not correct?

LebedevRI:
@jenshannoschwalm thank you!

PR #277, opened 2021-05-22
Fix wrong ISO values in Cr2Decoder for ISO > 65535

### LINKED ISSUE #276: RECOMMENDEDEXPOSUREINDEX must be used instead of ISOSPEEDRATINGS for Hi-ISO settings
ISOSPEEDRATINGS is a unsigned SHORT.
For cameras supporting ISO values over 65535, the correct ISO value is stored in RECOMMENDEDEXPOSUREINDEX, which is a unsigned LONG.

This is the case for Canon cameras, not sure about other manufacturers.

I will provide a PR for the Cr2Decoder.

cytrinox:
The EXIF ISOSPEEDRATINGS tag is a unsigned SHORT int (2^16).
For lager ISO values, we must use RECOMMENDEDEXPOSUREINDEX instead.

Fixes #276.


LebedevRI:

PR #279, opened 2021-05-29
Add Ricoh GR III

johnny-bit:
I hope I made this one right.

This allows for full run of https://github.com/darktable-org/darktable/issues/9091 (which reads from cameras.xml) as well as makes darktable nicely accept DNG from https://raw.pixls.us/getfile.php/3115/nice/Ricoh%20-%20RICOH%20GR%20III%20-%2014bit%20(3:2).DNG with no rendition artifacts

(without this, the colors i get are wack).

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -9772,6 +9772,9 @@
 	<Camera make="RICOH" model="GR DIGITAL 2" mode="dng">
 		<ID make="Ricoh" model="GR II">Ricoh GR DIGITAL 2</ID>
 	</Camera>
+	<Camera make="RICOH IMAGING COMPANY, LTD." model="RICOH GR III" mode="dng">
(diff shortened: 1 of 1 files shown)


johnny-bit:
garbage bot...

grattus:
Thanks for taking the time. Ricoh GR Digital 3 and Ricoh GR III are different cameras so you may want to change your commit to `<ID make="Ricoh" model="GR III">Ricoh GR III</ID>` for greater clarity?

johnny-bit:
Sure.

btw: congrats ricoh! "Ricoh GR III digital " != "Ricoh GR Digital 3". one was released in 2019, other in 2009... No confusing names what so ever :)

LebedevRI (commented):
Well, we really shouldn't second-guess DNG's, but then this is consistent with the rest of this file, so whatever.

LebedevRI:
I don't think any of these Ricoh entries are right, but well...

PR #280, opened 2021-06-09
rawspeed: Pentax K-3 Mark III support

kmilos:
Addresses part of https://github.com/darktable-org/darktable/issues/8991

black and white levels come from DNG converter and `dngmeta.sh` script, crops from visual inspection of RPU PEF sample

Not sure about the make and model though as scheme seems to be changed from K-3 II...

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -6755,6 +6755,17 @@
                 <Crop x="4" y="0" width="-36" height="0"/>
                 <Sensor black="1" white="15865"/>
         </Camera>
+        <Camera make="RICOH IMAGING COMPANY, LTD." model="PENTAX K-3 Mark III">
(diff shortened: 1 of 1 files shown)


kmilos:
I think the <ID> field should be changed to match the current Pentax scheme, but that will then break the initial noise profile in https://github.com/darktable-org/darktable/issues/9047 that was included (maybe too early?) for the DNG variety...

data/cameras.xml
```diff
@@ -6755,6 +6755,17 @@
                 <Crop x="4" y="0" width="-36" height="0"/>
                 <Sensor black="1" white="15865"/>
         </Camera>
+        <Camera make="RICOH IMAGING COMPANY, LTD." model="PENTAX K-3 Mark III">
```
LebedevRI, line 6758: These aren't tabs, please fix your editor, or switch to another one.

PR #286, opened 2021-08-11
Add Sony ILCE-7RM4A alias

kmilos:
See https://github.com/darktable-org/darktable/issues/9347 and https://github.com/darktable-org/darktable/issues/9751


LebedevRI:
If there isn't going to be a sample for `Sony ILCE-7RM3A`, please drop it from the PR.
Thanks.

kmilos:
Ok, let's see in a while (i.e. dt 3.8 timeframe) what samples we get.

LebedevRI:
ILCE-7RM4A samples are up on RPU, so i'd merge that, if that was the only change here.

PR #288, opened 2021-08-11
Ignore .vscode folder

kmilos:


LebedevRI:
IMHO this is backwards (and has been that way since beginning, in every repo).
This will never stop, there will always be some cruft left by some program,
and once it's hardcoded here, we're stuck with it.
We should just drop this file instead.

kmilos:
It's you repo ;) 

What's your preferred way of avoiding accidental additions?

LebedevRI:
> It's you repo ;)
> 
> What's your preferred way of avoiding accidental additions?

Most importantly by not doing `git add -a`, actually reading the diff before pushing it,
and maybe a little of `$XDG_CONFIG_HOME/git/ignore`, `repo-checkout/.git/info/exclude`.
I think we shouldn't get stuck with preferential treatment though.

PR #292, opened 2021-08-24
Panasonic S5/S1/S1H: Fix Black Point

jdneumeyer77:
-Fix black point by reading shorts/u16 instead of u32. Don't add 15 or check for overflow.

There's discussion in this Darktable [issue](https://github.com/darktable-org/darktable/issues/9811) about this issue.

Jens suggested I check against the white point, but I don't see it in this decoder. There doesn't seem to be a reason why there was addition of 15 or the overflow check. It appears to be wrong assumption as it causes low-light images to be green.

Here's the Exiftools document about [Panasonic Raws](https://exiftool.org/TagNames/PanasonicRaw.html) (RW2) to verify the black levels are shorts.

Let me know if there's any other changes.

### DIFF
--- a/src/librawspeed/decoders/Rw2Decoder.cpp
+++ b/src/librawspeed/decoders/Rw2Decoder.cpp
@@ -214,13 +214,9 @@ void Rw2Decoder::decodeMetaDataInternal(const CameraMetaData* meta) {
   if (raw->hasEntry(static_cast<TiffTag>(0x1c)) &&
       raw->hasEntry(static_cast<TiffTag>(0x1d)) &&
       raw->hasEntry(static_cast<TiffTag>(0x1e))) {
-    const auto getBlack = [&raw](TiffTag t) -> int {
-      return raw->getEntry(t)->getU16();
-    };
-
-    const int blackRed = getBlack(static_cast<TiffTag>(0x1c));
-    const int blackGreen = getBlack(static_cast<TiffTag>(0x1d));
-    const int blackBlue = getBlack(static_cast<TiffTag>(0x1e));
+    const int blackRed = raw->getEntry(static_cast<TiffTag>(0x1c))->getU16();
+    const int blackGreen = raw->getEntry(static_cast<TiffTag>(0x1d))->getU16();
+    const int blackBlue = raw->getEntry(static_cast<TiffTag>(0x1e))->getU16();
 
     for(int i = 0; i < 2; i++) {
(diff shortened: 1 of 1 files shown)


src/librawspeed/decoders/Rw2Decoder.cpp
```diff
@@ -215,11 +215,7 @@ void Rw2Decoder::decodeMetaDataInternal(const CameraMetaData* meta) {
       raw->hasEntry(static_cast<TiffTag>(0x1d)) &&
       raw->hasEntry(static_cast<TiffTag>(0x1e))) {
     const auto getBlack = [&raw](TiffTag t) -> int {
```
LebedevRI: Just inline it then

LebedevRI:
Looking at the sample 
> I've attached a test image and I license it under under CC0 for testing purposes.
> [samples.zip](https://github.com/darktable-org/darktable/files/6987394/samples.zip)

@ https://github.com/darktable-org/darktable/issues/9285, black level of ~512 does seem about right.
Git speluking suggests that `+15` originates all the way back from eabe61462c40db1e39a734633dc0fecbc752db47,
so no idea why it is there.

This affects literally every RW2 image. I'm not going to be able to check every single one,
but the ones i checked, this //seemed// to be better. But i'm just not sure.

LebedevRI:
Thank you!

jdneumeyer77:
Eek fixing an 8 year old change. Yeah I didn't think about the Panasonic m43 cameras. I can sample a few of those images if you want me to.

LebedevRI:
> Eek fixing an 8 year old change. Yeah I didn't think about the Panasonic m43 cameras. I can sample a few of those images if you want me to.

The key question is: do we have a full sample set under CC0 license for $CAMERA on RPU?

jdneumeyer77:
That's a good question. I see `1,549 total entries` when I search Panasonic on RPU. I'm not familiar with the testing strategy for rawspeed or at least how to test such a change like mine across many cameras beyond checking a few samples with exiftools. The exifdata link on RPU doesn't seem to show the above individual black level values.

LebedevRI:
Just enter the camera name into the search field https://raw.pixls.us/#repo,
or look if there's an appropriately-named directory in https://raw.pixls.us/data/Panasonic/

LebedevRI:
And reverted in b701374539fa224daf6533b1d80963f4845363da / xref https://github.com/darktable-org/darktable/issues/10008.

Issue #226, opened 2020-03-25
OSX Build error


MStraeten:
found following error on building darktable on OSX after updating to XCode 11.4

```
[  6%] Building CXX object src/external/rawspeed/CMakeFiles/rawspeed.dir/src/librawspeed/common/RawImage.cpp.o
/Users/martinstraeten/src/darktable/src/external/rawspeed/src/librawspeed/common/RawImage.cpp:408:23: error:
      variable 'threads' must have explicitly specified data sharing attributes
  for (int i = 0; i < threads; i++) {
                      ^~~~~~~
/Users/martinstraeten/src/darktable/src/external/rawspeed/src/librawspeed/common/RawImage.cpp:404:34: note:
      explicit data sharing attribute requested here
#pragma omp parallel for default(none)                                         \
                                 ^
/Users/martinstraeten/src/darktable/src/external/rawspeed/src/librawspeed/common/RawImage.cpp:409:47: error:
      variable 'height' must have explicitly specified data sharing attributes
    int y_offset = std::min(i * y_per_thread, height);
                                              ^~~~~~
/Users/martinstraeten/src/darktable/src/external/rawspeed/src/librawspeed/common/RawImage.cpp:404:34: note:
      explicit data sharing attribute requested here
#pragma omp parallel for default(none)                                         \
                                 ^
/Users/martinstraeten/src/darktable/src/external/rawspeed/src/librawspeed/common/RawImage.cpp:412:33: error:
      variable 'task' must have explicitly specified data sharing attributes
    RawImageWorker worker(this, task, y_offset, y_end);
                                ^~~~
/Users/martinstraeten/src/darktable/src/external/rawspeed/src/librawspeed/common/RawImage.cpp:404:34: note:
      explicit data sharing attribute requested here
#pragma omp parallel for default(none)                                         \
                                 ^
/Users/martinstraeten/src/darktable/src/external/rawspeed/src/librawspeed/common/RawImage.cpp:409:33: error:
      variable 'y_per_thread' must have explicitly specified data sharing
      attributes
    int y_offset = std::min(i * y_per_thread, height);
                                ^~~~~~~~~~~~
/Users/martinstraeten/src/darktable/src/external/rawspeed/src/librawspeed/common/RawImage.cpp:404:34: note:
      explicit data sharing attribute requested here
#pragma omp parallel for default(none)                                         \
                                 ^
4 errors generated.
[...]

PR #294, opened 2021-08-27
Onwards to C++17 / OpenMP 4.5

LebedevRI:
Not really sure what this means for Apple XCode support.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,7 +1,7 @@
 ---
 Language:     Cpp
 BasedOnStyle: LLVM
-Standard:     Cpp11
+Standard:     c++17
 # Force pointers to the type for C++.
 DerivePointerAlignment: false
 PointerAlignment: Left
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -6,6 +6,8 @@ on:
   pull_request:
     branches: [ develop ]
 
+# FIXME: add no-openmp, and maybe -DCMAKE_BUILD_TYPE=Debug jobs
+
 jobs:
   linux:
     name: ${{ matrix.os }}.${{ matrix.compiler.compiler }}.${{ matrix.flavor }}
--- a/.github/workflows/ci-linux-llvm-legacy.yml
+++ b/.github/workflows/ci-linux-llvm-legacy.yml
@@ -7,69 +7,6 @@ on:
     branches: [ develop ]
 
 jobs:
-  linux-llvm6:
-    name: ${{ matrix.os }}.${{ matrix.compiler.compiler }}.${{ matrix.flavor }}
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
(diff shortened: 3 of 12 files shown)


simonspa:
Most language features should be in Xcode/AppleClang 9.3 but many library features have only been added in version 10: https://en.cppreference.com/w/cpp/compiler_support#cpp17

LebedevRI:
> Most language features should be in Xcode/AppleClang 9.3 but many library features have only been added in version 10: https://en.cppreference.com/w/cpp/compiler_support#cpp17

To me the "Apple Clang" columns read as "Yes", with no version specified..

Apple Clang/XCode 10 is already required, i think:
https://github.com/darktable-org/rawspeed/blob/ab3bf7290d5c6e086becdca4156be0254d8ee92a/cmake/compiler-versions.cmake#L17-L22

But that being said, as i've previously noted in https://github.com/darktable-org/rawspeed/issues/226#issuecomment-902792759

> @MStraeten @parafin sorry for necro-posting, but i'm revisiting this area of cmake, and i'm somewhat confused.
> This workarounds https://bugs.llvm.org/show_bug.cgi?id=35873, which was fixed in LLVM/Clang 7: https://godbolt.org/z/v9dcT9rK5
> But as per https://en.wikipedia.org/wiki/Xcode#Xcode_7.0_-_10.x_(since_Free_On-Device_Development)_2, the first XCode version based on LLVM 7 was `XCode 10.2` (aka `10.0.1 (clang-1001.0.46.3)`).
> 
> So how come
> 
> > (CMAKE_CXX_COMPILER_VERSION "11.0.0.11000033")
> 
> which i believe belongs to XCode 11.0, is still broken?
> 
> I guess my main question is: which is the first XCode version for which https://godbolt.org/z/v9dcT9rK5 is fixed?

simonspa:
Phew, to be honest, I never fully understood the versioning of AppleClang and XCode.

The linked cppreference table notes "yes" for most, but "10.0.0" for some of the library features, hence my comment on that. They might still have not implemented.

I guess we'd have to add specific XCode versions to the CI and simply test if they work?

LebedevRI:
> Phew, to be honest, I never fully understood the versioning of AppleClang and XCode.
> 
> The linked cppreference table notes "yes" for most, but "10.0.0" for some of the library features, hence my comment on that. They might still have not implemented.

> I guess we'd have to add specific XCode versions to the CI and simply test if they work?

That's what i'm currently trying to do..

LebedevRI:
> > Phew, to be honest, I never fully understood the versioning of AppleClang and XCode.
> > The linked cppreference table notes "yes" for most, but "10.0.0" for some of the library features, hence my comment on that. They might still have not implemented.
> 
> > I guess we'd have to add specific XCode versions to the CI and simply test if they work?
> 
> That's what i'm currently trying to do..

Ok, that was surprisingly easy to do, #295.
I've added only the latest minor release of each available major release, hopefully that will be good-enough.

Issue #176, opened 2019-05-22
Class RangesOverlapCmp cannot be used as a comparator for std::set in NORangesSet


govage:
std::set requires comparison functor for which cmp(x, y) == !cmp(y, x)
See https://en.cppreference.com/w/cpp/container/set and https://en.cppreference.com/w/cpp/named_req/Compare
But RangesOverlapCmp doesn't meet these criteria because it uses function RangesOverlap for which RangesOverlap(x, y) == RangesOverlap(y, x). When trying to run on Visual Studio, it shows assertion error message in debug mode.

Issue #287, opened 2021-08-11
This comparison operator is breaks weak ordering


Alexey-Danilchenko:
https://github.com/darktable-org/rawspeed/blob/4f03e0d42b1842f18b30a4e225e3bd39cbe00939/src/librawspeed/common/NORangesSet.h#L35

This comparison operator breaks weak ordering and on compilers with strict checking (GCC) breaks. It is quite understandable because for each a<b there also will be b<a so I am not sure how insering a buffer in a set will actually work for two non overlapping buffers. Specifically usage here breaking at runtime compiled with GCC and strict weak ordering checks:

https://github.com/darktable-org/rawspeed/blob/develop/src/librawspeed/tiff/TiffIFD.cpp?fbclid=IwAR3bEp2Hmm_Fj1ZbwDznU72_Nq23YNvnbNdOMFOoOgPIMqwLLauTYOpu3ew#L118

LebedevRI:
> Specifically usage here breaking at runtime compiled with GCC and strict weak ordering checks:

Which are enabled how?

Duplicate of #176

Alexey-Danilchenko:
Enabled by default with libstdc++. How this works in clang - no idea.

PR #299, opened 2021-09-03
(more) Correct NORangesSet implementation

LibRaw:
Description:
RangesOverlapCmp is not suitable as comparator for (ordered) std::set
because of RangesOverlapCmp(a,b) == RangesOverlapCmp(b,a) if ranges not
overlap.  Replaced with ordered comparator.

Because NORangesSet is used only for IFD overlap check via emplace call:
created own emplace implementation with RangesOverlap check within.

Note: Microsoft C++ library (in debug mode) compares _Pred(a,b) against _Pred(b,a) and asserts

### DIFF
--- a/src/librawspeed/common/NORangesSet.h
+++ b/src/librawspeed/common/NORangesSet.h
@@ -26,12 +26,26 @@
 
 namespace rawspeed {
 
-template <typename T> struct RangesOverlapCmp final {
-  constexpr bool operator()(const T& lhs, const T& rhs) const {
-    return !RangesOverlap(lhs, rhs);
-  }
+template <typename T> struct RangesSortCmp final {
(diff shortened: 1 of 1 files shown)


LibRaw:
This fixes: https://github.com/darktable-org/rawspeed/issues/176 (and https://github.com/darktable-org/rawspeed/issues/287 ) and allows clang-cl use (w/ Microsoft C++ library)

LebedevRI:
Three comments:
1. please ensure you follow https://github.com/darktable-org/darktable/blob/master/CONTRIBUTING.md#coding-style
2. you can't inherit from C++ STL classes
~3. this may or may not address whatever pedantic precondition violation, but the underlying algorithm is still incorrect.~

Let me just fix this.

PR #301, opened 2021-09-05
Remove final from CameraMetaData to allow derived classes

LibRaw:
CameraMetaData derived classes are needed if one want to create CameraMetaData ancestor not from file, but from other source (compressed resource, etc)


LebedevRI:
If this had come with a test it would have been merged faster :|

PR #305, opened 2021-09-10
[CI] Add entry with minimal supported OSX backdeployment target

LebedevRI:
cc @parafin

### DIFF
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -116,13 +116,14 @@ jobs:
           cmake -E make_directory "${INSTALL_PREFIX}";
           $(cygpath ${SRC_DIR})/.ci/ci-script.sh;
   macOS:
-    name: ${{ matrix.os }}.${{ matrix.compiler.compiler }}.${{ matrix.flavor }}
+    name: ${{ matrix.os }}.${{ matrix.compiler.compiler }}.${{ matrix.xcode }}.targeting.OSX.${{ matrix.deployment_target }}.${{ matrix.flavor }}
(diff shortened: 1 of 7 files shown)


LebedevRI:
@parafin hurray, that wasn't so bad.
But please, once you do happen to update XCode/`MACOSX_DEPLOYMENT_TARGET`, please ping about it :)

.github/workflows/ci-fast.yml
```diff
@@ -116,13 +116,14 @@ jobs:
...
   macOS:
-    name: ${{ matrix.os }}.${{ matrix.compiler.compiler }}.${{ matrix.flavor }}
+    name: ${{ matrix.os }}.${{ matrix.compiler.compiler }}.${{ matrix.xcode }}.targeting.OSX.${{ matrix.deployment_target }}.${{ matrix.flavor }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ macos-11 ]
         xcode: [ 12.5.1 ] # LLVM11
+        deployment_target: [ 12.5.1, 10.7 ]
```
parafin, line 126: 12.5.1 is Xcode version, not macOS. I guess you meant 11.3?

LebedevRI: Err, i suppose.
What i meant is "just target the current target"

parafin: Not sure if there's such (valid) value. But given that xcode version is fixed in CI config, you can fix deployment target to SDK version shipped with that Xcode version too. In case of Xcode 12.5.1 it's macOS 11.3 SDK.

PR #307, opened 2021-09-29
Fix Panasonic Black Levels for newer cameras.

jdneumeyer77:
Pivot black levels on raw version. Newer then version 5 should have correct black levels.

### DIFF
--- a/src/librawspeed/decoders/Rw2Decoder.cpp
+++ b/src/librawspeed/decoders/Rw2Decoder.cpp
@@ -226,16 +226,16 @@ void Rw2Decoder::decodeMetaDataInternal(const CameraMetaData* meta) {
       raw->hasEntry(static_cast<TiffTag>(0x1e))) {
     const auto getBlack = [&raw, version](TiffTag t) {
       const int val = raw->getEntry(t)->getU16();
+      int out;
       // After version 4 the black levels appears to be correct.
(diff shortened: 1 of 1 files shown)


jdneumeyer77:
@LebedevRI any feedback on this? Not sure why the `linux.LLVM12.SonarCloudStaticAnalysis` is failing. Not sure how to fix the test coverage. I'm guessing the tests don't include any version 5/6 Panasonic raw files. So it's not hitting the conditional. Any guidance on how to add those?

I also confirmed +15 is needed in older cameras by looking at the DNGs. Opening them includes the +15 while opening newer Panasonic (GH5S -> v5, S1, S5, etc. -> v6) files does not. These were the sample raw files from pixl.us converted using Adobe DNG convertor 14 via Wine.

LebedevRI:
@jdneumeyer77 i am sorry. I'm silent on this not because CI isn't silent/green,
but for the same reason there hasn't been much activity in this repo by me lately.

LebedevRI:
Ok, i've manually inspected all the affected RPU samples (you know, unlike the last time ^^),
and the _seem_ to be improvements.

jdneumeyer77:
@LebedevRI thanks for verifying and merging! There's no need to apologize. I was just checking in on the issue. We all get busy or life gets in the way sometimes.

PR #309, opened 2021-10-03
Ricoh GR II (firmware 3.0) alias, fixes denoise (profiled)

markotoplak:
The current release of darktable can not find the noise profile for the camera I use even though it is present among darktable noise profiles. This alias fixes the problem.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -9979,6 +9979,9 @@
 	<Camera make="RICOH" model="GR DIGITAL 2" mode="dng">
 		<ID make="Ricoh" model="GR II">Ricoh GR DIGITAL 2</ID>
 	</Camera>
+	<Camera make="RICOH IMAGING COMPANY, LTD." model="GR II" mode="dng">
+		<ID make="Ricoh" model="GR II">Ricoh GR II</ID>
+ 	</Camera>
 	<Camera make="RICOH IMAGING COMPANY, LTD." model="RICOH GR III" mode="dng">
 		<ID make="Ricoh" model="GR III">Ricoh GR III</ID>
 	</Camera>


LebedevRI:
I'm confident that you have verified that this fixes the issue for you.

kmilos:
At the same time, it looks like there might be a mistake on the line above: "RICOH GR DIGITAL 2" should not be normalized to "Ricoh GR II", these are completely different cameras (10Mpix 2007 vs 16Mpix 2015)... Nice one, Ricoh...

https://raw.pixls.us/getfile.php/667/exif/RAW_RICOH_GR2.DNG.exif.txt
https://raw.pixls.us/getfile.php/1955/exif/R0000443.DNG.exif.txt

@markotoplak You mind adding a quick fix? Best to remove the entire camera model as we're doing case-sensitive matching and this is already present as ["RICOH GR DIGITAL 2"](https://github.com/darktable-org/darktable/blob/a422bfd966af56a15c5fe0b8304dca94abad4977/src/external/adobe_coeff.c#L620) which you'll get by default from the metadata anyway if not present in cameras.xml... Luckily "GR DIGITAL 3 and "GR DIGITAL 4" are not present/aliased...

(Not even sure why there's an entry for these in adobe_coeffs since these are DNGs w/ valid color matrices anyway...)

markotoplak:
@kmilos, yes, I agree. I found it questionable, but did not want to change too much (perhaps some GRD2 user was relying on this bug). I'll open a PR.

kmilos:
@markotoplak I think you'll find you'll need to now fix https://github.com/darktable-org/darktable/blob/master/src/external/wb_presets.c as well

markotoplak:
@kmilos, that is a good find, thanks. Interesting how a small change propagates through the project: it happens a lot on my main project (biolab/orange3) too. :)

Do you know how are these WB numbers obtained? Could I somehow verify them with my DNG files? What surprises me is that Ricoh GR and GR II have slightly different numbers, even though they have an identical sensor (and supposedly processing).

kmilos:
> Do you know how are these WB numbers obtained? Could I somehow verify them with my DNG files?

Sure, see [here](https://github.com/darktable-org/darktable/wiki/Camera-support#white-balance-presets).

PR #310, opened 2021-10-04
Remove wrong Ricoh GR II alias (GR Digital 2 is a different camera)

markotoplak:
As suggested in #309 by @kmilos - and I agree.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -9976,9 +9976,6 @@
 	<Camera make="RICOH IMAGING COMPANY, LTD." model="GR" mode="dng">
 		<ID make="Ricoh" model="GR">GR</ID>
 	</Camera>
-	<Camera make="RICOH" model="GR DIGITAL 2" mode="dng">
-		<ID make="Ricoh" model="GR II">Ricoh GR DIGITAL 2</ID>
-	</Camera>
 	<Camera make="RICOH IMAGING COMPANY, LTD." model="GR II" mode="dng">
 		<ID make="Ricoh" model="GR II">Ricoh GR II</ID>
  	</Camera>


data/cameras.xml
```diff
@@ -9976,9 +9976,6 @@
 	<Camera make="RICOH IMAGING COMPANY, LTD." model="GR" mode="dng">
 		<ID make="Ricoh" model="GR">GR</ID>
 	</Camera>
-	<Camera make="RICOH" model="GR DIGITAL 2" mode="dng">
-		<ID make="Ricoh" model="GR II">Ricoh GR DIGITAL 2</ID>
```
LebedevRI, line 9980: Presumably this should be `<ID make="Ricoh" model="GR DIGITAL 2">Ricoh GR DIGITAL 2</ID>`?

kmilos: I think it's ok to remove completely - AFAICT rawspeed's [DngDecoder](https://github.com/darktable-org/rawspeed/blob/develop/src/librawspeed/decoders/DngDecoder.cpp#L577-L600) should get "RICOH"/"GR DIGITAL 2" into canonical_make/model, and "Ricoh GR DIGITAL 2" into canonical_id [from the DNG UniqueCameraModel tag without the help of cameras.xml in this case, as that DNG tag exists and is correct](https://raw.pixls.us/getfile.php/667/exif/RAW_RICOH_GR2.DNG.exif.txt).

markotoplak: I agree with @kmilos. Besides, if I remember correctly, there were 4 versions of GR DIGITAL cameras, so, if they need to be added, it would be best to add all models.

PR #312, opened 2021-10-18
Add Sony DSC-HX95 alias

kmilos:
See https://www.digitalcameraworld.com/uk/news/sony-unleashes-pocket-sized-hx95-and-hx99-travel-zoom-cameras

Also compared DCPs shipped w/ DNG converter, there is only the single byte (9 vs 5) difference.

([Related support request to be created once the samples are up.](https://discuss.pixls.us/t/sony-dsc-hx95-not-supported/27475))

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -9415,6 +9415,9 @@
 		</CFA>
 		<Crop x="10" y="0" width="-20" height="-10"/>
 		<Sensor black="800" white="16300"/>
+		<Aliases>
+			<Alias>DSC-HX95</Alias>
+		</Aliases>
 	</Camera>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Thanks.

LebedevRI:
Hm, i should not have merged this, there are no RPU samples.

kmilos:
The user requesting this should be uploading soon, as promised. Should be 99% ok ;)

LebedevRI:
If there isn't one by friday i'll revert.
In future, i will try to avoid making this mistake in the first place.

LebedevRI:
Thanks, resolved.

PR #322, opened 2021-11-13
More Leica C support

jenshannoschwalm:
For this model we also want a 3:2 ratio definer.


LebedevRI:
RPU sample?

jenshannoschwalm:
Just uploaded to raw.pixls; unfortunately when checking i can't find it and uploading again does not work.

LebedevRI:
> Just uploaded to raw.pixls; unfortunately when checking i can't find it and uploading again does not work.

That's how it works, yes.

PR #323, opened 2021-11-26
[CI] Add CIFuzz integration

DavidKorczynski:
Add CIFuzz workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which rawspeed already runs.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/CI-ossfuzz.yml
+++ b/.github/workflows/CI-ossfuzz.yml
@@ -1,0 +1,32 @@
+on:
+  workflow_call:
+    inputs:
+      sanitizer:
+        required: true
+        type: string
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: (OSS-FUZZ) Building with ${{ inputs.sanitizer }} sanitizer
+        id: build
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+        with:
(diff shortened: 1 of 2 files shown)


LebedevRI:
Is it required?

DavidKorczynski:
It's simple an added service - it can help catch regressions early or catch fuzzing breaks early. It was only recently released officially and there are already many projects on OSS-Fuzz that uses CIFuzz too. As such, it's not really a requirement but more of an extra feature.

LebedevRI:
I think, ignoring the question of per-project usefulness, my main issue with this is that it's a half of an integration test.
Because it is an integration test in the end, the fact that it might catch bugs in process is secondary.
It's a half because it only runs on PR's, and ignoring that, the problem with such an integration test is that
it inherently depends on external 3rd party consumer implementation detail.
E.g. https://github.com/google/oss-fuzz/tree/master/projects/librawspeed isn't in this repo,
so once they need to be modified, there's an issue.

DavidKorczynski:
> it only runs on PR's

We can set it to run on Push too. But, in the event you still don't find it useful then let's just close this PR!

LebedevRI:
Meh.
@DavidKorczynski thank you, after further consideration, i decided to take this,
at least to have build-coverage for sanitizers and/or for the clang trunk.

PR #324, opened 2021-12-05
Add GFX50S II support

Dannny1:


Dannny1:
from LibRaw code:
    { LIBRAW_CAMERAMAKER_Fujifilm, "GFX 50", 0, 0,  // same CMs: "GFX 50S", "GFX 50R", "GFX 50S II"
      { 11756,-4754,-874,-3056,11045,2305,-381,1457,6006 } },

LebedevRI:
There are no samples for those cameras on RPU.
We need to have full test coverage first.

LebedevRI:
Can anyone confirm that the dng converter believes this camera to be an alias to the GFX 50?

Dannny1:
I played with few files and dng converter showed same values as present in the libraw code.

data/cameras.xml
```diff
@@ -8588,6 +8588,10 @@
 		</CFA>
 		<Crop x="0" y="0" width="-938" height="-4"/>
 		<Sensor black="65" white="16383"/>
+		<Aliases>
+			<Alias>GFX 50S II</Alias>
```
LebedevRI, line 8592: Hm, why does this need two aliases? Which one is actually used by the camera?

kmilos:
They might use the same sensor, but IMHO they are different cameras: [added new compression mode, image stabilization](https://www.dpreview.com/reviews/fujifilm-gfx-50s-ii-review#WN)...

LebedevRI:
The samples are just too dark to tell, but i strongly believe treating it as an alias is wrong.

Dannny1:
GFX50S II is used

PR #329, opened 2021-12-14
Add Sony ILCE-7M4 camera support

### LINKED ISSUE #328: Add support for Sony A7 IV
Hi Roman,

I've just uploaded RAW files for the A7M4 to raw.pixls.us. Let me know if you need any additional pictures.

Thanks!

cryptomilk:
Fixes #328

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -7930,6 +7930,17 @@
 		<Crop x="0" y="0" width="0" height="0"/>
 		<Sensor black="512" white="16300"/>
 	</Camera>
+	<Camera make="SONY" model="ILCE-7M4">
+		<ID make="Sony" model="ILCE-7M4">Sony ILCE-7M4</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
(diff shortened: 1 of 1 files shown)


data/cameras.xml
```diff
@@ -7930,6 +7930,17 @@
...
 	</Camera>
+	<Camera make="SONY" model="ILCE-7M4">
+		<ID make="Sony" model="ILCE-7M4">Sony ILCE-7M4</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
+		</CFA>
+		<Crop x="0" y="0" width="0" height="0"/>
```
kmilos, line 7941: There's definitely a need to remove some pixels on the right side, the question is - how many exactly?

LebedevRI:
https://raw.pixls.us/ says
```
We are looking for shots that are:
* Image in focus and properly exposed
```
the RPU samples are not. I can not guess the right crop from them.
They really need to be landscape shots, bright, with sharp patterns all across.

LebedevRI:
I just can't tell what the right crop here is.
It seems like at least -8, so I've gone ahead with -10, but it might be too conservative still. (could be up to -30).

kmilos:
-10 was my best guess as well for the RPU sample...

PR #330, opened 2021-12-14
rawspeed: Fujifilm GFX50S II support

kmilos:
Output from dngmeta.sh script, w/ added compressed mode:

TODO (for both modes):

1. Verify crop (copy from 50S/50R?)
2. Verify black level (DNG says 64; 50S/50R have 65 in rawspeed and also in DNG, so Adobe does not treat them as 100% identical indeed)

Also, is 3rd (new lossy) compression mode missing? (also for GFX100S?) I have no idea how/if this new mode works out w/ rawspeed...

Addresses part of https://github.com/darktable-org/darktable/issues/10608

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -8644,6 +8644,28 @@
 		<Crop x="0" y="0" width="-938" height="-4"/>
 		<Sensor black="65" white="16383"/>
 	</Camera>
+        <Camera make="FUJIFILM" model="GFX50S II">
+                <ID make="Fujifilm" model="GFX50S II">Fujifilm GFX 50S II</ID>
+                <CFA width="2" height="2">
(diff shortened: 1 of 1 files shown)


data/cameras.xml
```diff
@@ -8644,6 +8644,28 @@
 		<Crop x="0" y="0" width="-938" height="-4"/>
 		<Sensor black="65" white="16383"/>
 	</Camera>
+        <Camera make="FUJIFILM" model="GFX50S II">
```
LebedevRI, line 8647: These aren't tabs.

kmilos: Sorry - copy/paste from MSYS2's mintty changes those then, I'll keep it in mind.

Issue #258, opened 2020-12-31
LJpegDecompressor predictor mode 7 support


LebedevRI:
Comes up in `iPhone 12 Pro` raws: https://raw.pixls.us/getfile.php/4264/raw/IMG_1361.DNG
```
/tmp/IMG_1361.DNG                                      : starting decoding ... 
/tmp/IMG_1361.DNG failed: ../src/librawspeed/decompressors/AbstractDngDecompressor.cpp:218: void rawspeed::AbstractDngDecompressor::decompress() const: Too many errors encountered. Giving up. First Error:
../src/librawspeed/decompressors/LJpegDecompressor.cpp:91: virtual void rawspeed::LJpegDecompressor::decodeScan(): Unsupported predictor mode: 7
Total decoding time: 0s

WARNING: the following 1 tests have failed:
/tmp/IMG_1361.DNG failed: ../src/librawspeed/decompressors/AbstractDngDecompressor.cpp:218: void rawspeed::AbstractDngDecompressor::decompress() const: Too many errors encountered. Giving up. First Error:
../src/librawspeed/decompressors/LJpegDecompressor.cpp:91: virtual void rawspeed::LJpegDecompressor::decodeScan(): Unsupported predictor mode: 7
==46741==LeakSanitizer has encountered a fatal error.
==46741==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==46741==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
```

MaykThewessen:
A fix is proposed in https://github.com/darktable-org/rawspeed/pull/963

Issue #189, opened 2019-06-30
LJpegDecompressor predictor mode 6 support


LebedevRI:
Comes up in Blackmagic cameras, Magic Lantern DNG's.
```
./Blackmagic/Pocket Cinema Camera/Blackmagic_Pocket_Cinema_Camera_1_2016-09-17_0053_C0022_000007.dng failed: ../src/librawspeed/decompressors/AbstractDngDecompressor.cpp:208: void rawspeed::AbstractDngDecompressor::decompress() const: Too many errors encountered. Giving up. First Error:
../src/librawspeed/decompressors/LJpegDecompressor.cpp:88: virtual void rawspeed::LJpegDecompressor::decodeScan(): Unsupported predictor mode: 6
./Blackmagic/URSA 4K/Blackmagic URSA_1_1999-12-31_2153_C0003_000001.dng failed: ../src/librawspeed/decompressors/AbstractDngDecompressor.cpp:208: void rawspeed::AbstractDngDecompressor::decompress() const: Too many errors encountered. Giving up. First Error:
../src/librawspeed/decompressors/LJpegDecompressor.cpp:88: virtual void rawspeed::LJpegDecompressor::decodeScan(): Unsupported predictor mode: 6
./Canon/EOS 5D Mark III/compress.dng failed: ../src/librawspeed/decompressors/AbstractDngDecompressor.cpp:208: void rawspeed::AbstractDngDecompressor::decompress() const: Too many errors encountered. Giving up. First Error:
../src/librawspeed/decompressors/LJpegDecompressor.cpp:88: virtual void rawspeed::LJpegDecompressor::decodeScan(): Unsupported predictor mode: 6
./Canon/EOS 60D/M14-1451_000085_cDNG_compressed.dng failed: ../src/librawspeed/decompressors/AbstractDngDecompressor.cpp:208: void rawspeed::AbstractDngDecompressor::decompress() const: Too many errors encountered. Giving up. First Error:
../src/librawspeed/decompressors/LJpegDecompressor.cpp:88: virtual void rawspeed::LJpegDecompressor::decodeScan(): Unsupported predictor mode: 6

```

kmilos:
Even w/ the predictor implemented, there is still some way to go - seems to be the BlackMagick TIFF tile tags vs LJpeg SOF3 frame geometry is a bit wonky/unexpected compared to "normal" CFA lossless DNGs?

```
[...]

Issue #325, opened 2021-12-07
Investigate Halide dependency


LebedevRI:
Now that more time has passed, and it is more widely packaged,
it is time to see if it can help improve on the selling point of the lib.

It is TBD what that means for support for non-LLVM compilers,
but they will likely become unsupported.

This may or may not be a preparatory step for further invasion.

PR #334, opened 2021-12-30
Add LJpeg predictors 2-7

kmilos:
Proof of concept (non-performant) to address #189 and #258

### DIFF
--- a/src/librawspeed/decompressors/LJpegDecompressor.cpp
+++ b/src/librawspeed/decompressors/LJpegDecompressor.cpp
@@ -53,10 +53,12 @@ LJpegDecompressor::LJpegDecompressor(RawImage img, iRectangle2D imgFrame_,
                                      Frame frame_,
                                      std::vector<PerComponentRecipe> rec_,
                                      int numRowsPerRestartInterval_,
+                                     int predictorMode_,
                                      Array1DRef<const uint8_t> input_)
     : mRaw(std::move(img)), input(input_), imgFrame(imgFrame_),
       frame(std::move(frame_)), rec(std::move(rec_)),
-      numRowsPerRestartInterval(numRowsPerRestartInterval_) {
+      numRowsPerRestartInterval(numRowsPerRestartInterval_),
+      predictorMode(predictorMode_) {
 
   if (mRaw->getDataType() != RawImageType::UINT16)
     ThrowRDE("Unexpected data type (%u)",
@@ -100,7 +102,7 @@ LJpegDecompressor::LJpegDecompressor(RawImage img, iRectangle2D imgFrame_,
     ThrowRDE("Unsupported number of components: %u", frame.cps);
 
   if (rec.size() != static_cast<unsigned>(frame.cps))
-    ThrowRDE("Must have exactly one recepie per component");
+    ThrowRDE("Must have exactly one recipe per component");
 
   for (const auto& recip : rec) {
     if (!recip.ht.isFullDecode())
@@ -136,7 +138,7 @@ LJpegDecompressor::LJpegDecompressor(RawImage img, iRectangle2D imgFrame_,
   }
 
   if (numRowsPerRestartInterval < 1)
-    ThrowRDE("Number of rows per restart interval must be positives");
+    ThrowRDE("Number of rows per restart interval must be positive");
 
   // How many full pixel blocks will we produce?
   fullBlocks = tileRequiredWidth / frame.cps; // Truncating division!
@@ -169,7 +171,8 @@ std::array<uint16_t, N_COMP> LJpegDecompressor::getInitialPreds() const {
 
 template <int N_COMP, bool WeirdWidth>
 void LJpegDecompressor::decodeRowN(
-    CroppedArray1DRef<uint16_t> outRow, std::array<uint16_t, N_COMP> pred,
+    CroppedArray1DRef<uint16_t> outRow, CroppedArray1DRef<uint16_t> prevRow,
+    std::array<uint16_t, N_COMP> pred, int predMode,
     std::array<std::reference_wrapper<const PrefixCodeDecoder<>>, N_COMP> ht,
     BitStreamerJPEG& bs) const {
   // FIXME: predictor may have value outside of the uint16_t.
@@ -179,10 +182,39 @@ void LJpegDecompressor::decodeRowN(
   // For x, we first process all full pixel blocks within the image buffer ...
   for (; col < N_COMP * fullBlocks; col += N_COMP) {
     for (int i = 0; i != N_COMP; ++i) {
-      pred[i] =
+      outRow(col + i) =
           uint16_t(pred[i] + (static_cast<const PrefixCodeDecoder<>&>(ht[i]))
                                  .decodeDifference(bs));
-      outRow(col + i) = pred[i];
+      if (col < N_COMP * (fullBlocks - 1)) {
+        int32_t predA = outRow(col + i);
+        int32_t predB = predMode > 1 ? prevRow(col + N_COMP + i) : 0;
+        int32_t predC = predMode > 1 ? prevRow(col + i) : 0;
+        switch (predMode) {
+        case 1:
+          pred[i] = predA;
+          break;
+        case 2:
+          pred[i] = predB;
+          break;
+        case 3:
+          pred[i] = predC;
+          break;
+        case 4:
+          pred[i] = predA + predB - predC;
+          break;
+        case 5:
+          pred[i] = predA + ((predB - predC) >> 1);
+          break;
+        case 6:
+          pred[i] = predB + ((predA - predC) >> 1);
+          break;
+        case 7:
+          pred[i] = (predA + predB) >> 1;
+          break;
+        default:
+          __builtin_unreachable();
+        }
+      }
     }
   }
 
@@ -197,10 +229,38 @@ void LJpegDecompressor::decodeRowN(
     invariant(trailingPixels < N_COMP);
     int c = 0;
     for (; c < trailingPixels; ++c) {
-      pred[c] =
+      // Continue predictor update skipped at last full block
+      int32_t predA = outRow(col - N_COMP + c);
+      int32_t predB = predMode > 1 ? prevRow(col + c) : 0;
+      int32_t predC = predMode > 1 ? prevRow(col - N_COMP + c) : 0;
+      switch (predMode) {
+      case 1:
+        pred[c] = predA;
+        break;
+      case 2:
+        pred[c] = predB;
+        break;
+      case 3:
+        pred[c] = predC;
+        break;
+      case 4:
+        pred[c] = predA + predB - predC;
+        break;
+      case 5:
+        pred[c] = predA + ((predB - predC) >> 1);
+        break;
+      case 6:
+        pred[c] = predB + ((predA - predC) >> 1);
+        break;
+      case 7:
+        pred[c] = (predA + predB) >> 1;
+        break;
+      default:
+        __builtin_unreachable();
+      }
+      outRow(col + c) =
           uint16_t(pred[c] + (static_cast<const PrefixCodeDecoder<>&>(ht[c]))
                                  .decodeDifference(bs));
-      outRow(col + c) = pred[c];
     }
     // Discard the rest of the block.
     invariant(c < N_COMP);
@@ -251,8 +311,9 @@ ByteStream::size_type LJpegDecompressor::decodeN() const {
   ByteStream inputStream(DataBuffer(input, Endianness::little));
   for (int restartIntervalIndex = 0;
        restartIntervalIndex != numRestartIntervals; ++restartIntervalIndex) {
-    auto pred = getInitialPreds<N_COMP>();
-    auto predNext = Array1DRef(pred.data(), pred.size());
+    auto predInit = getInitialPreds<N_COMP>();
+    auto predNext = Array1DRef(predInit.data(), predInit.size());
+    std::array<uint16_t, N_COMP> pred;
 
     if (restartIntervalIndex != 0) {
       auto marker = peekMarker(inputStream);
@@ -283,14 +344,17 @@ ByteStream::size_type LJpegDecompressor::decodeN() const {
       }
 
       auto outRow = img[row];
+      auto prevRow = row > 0 ? img[row - 1] : img[row];
       copy_n(predNext.begin(), N_COMP, pred.data());
       // the predictor for the next line is the start of this line
       predNext = outRow
                      .getBlock(/*size=*/N_COMP,
                                /*index=*/0)
                      .getAsArray1DRef();
+      // the predictor mode is always horizontal on the first line
+      int predMode = row == 0 ? 1 : predictorMode;
 
-      decodeRowN<N_COMP, WeirdWidth>(outRow, pred, ht, bs);
+      decodeRowN<N_COMP, WeirdWidth>(outRow, prevRow, pred, predMode, ht, bs);
     }
 
     inputStream.skipBytes(bs.getStreamPosition());
--- a/fuzz/librawspeed/decompressors/LJpegDecompressor.cpp
+++ b/fuzz/librawspeed/decompressors/LJpegDecompressor.cpp
@@ -87,10 +87,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* Data, size_t Size) {
                     });
 
     const int numRowsPerRestartInterval = bs.getI32();
+    const int predictorMode = bs.getByte();
 
     rawspeed::LJpegDecompressor d(
         mRaw, rawspeed::iRectangle2D(mRaw->dim.x, mRaw->dim.y), frame, rec,
-        numRowsPerRestartInterval,
+        numRowsPerRestartInterval, predictorMode,
         bs.getSubStream(/*offset=*/0).peekRemainingBuffer().getAsArray1DRef());
     mRaw->createData();
     (void)d.decode();
--- a/src/librawspeed/decompressors/AbstractLJpegDecoder.cpp
+++ b/src/librawspeed/decompressors/AbstractLJpegDecoder.cpp
@@ -208,7 +208,7 @@ void AbstractLJpegDecoder::parseSOS(ByteStream sos) {
 
   // Get predictor, see table H.1 from the JPEG spec
   predictorMode = sos.getByte();
-  // The spec says predictoreMode is in [0..7], but Hasselblad uses '8'.
+  // The spec says predictorMode is in [0..7], but Hasselblad uses '8'.
   if (predictorMode > 8)
     ThrowRDE("Invalid predictor mode.");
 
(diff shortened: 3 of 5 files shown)


LebedevRI:
I'm currently crossing all t's and dotting all i's so that #325 can happen.

src/librawspeed/decompressors/LJpegDecompressor.cpp
```diff
@@ -194,15 +194,47 @@ template <int N_COMP, bool WeirdWidth> void LJpegDecompressor::decodeN() {
...
     for (; col < N_COMP * fullBlocks; col += N_COMP) {
       for (int i = 0; i != N_COMP; ++i) {
-        pred[i] = uint16_t(pred[i] + ht[i]->decodeDifference(bitStream));
-        img(row, col + i) = pred[i];
+        img(row, col + i) =
+            uint16_t(pred[i] + ht[i]->decodeDifference(bitStream));
+        if (col < N_COMP * (fullBlocks - 1)) {
+          int predA = img(row, col + i);
+          int predB = predMode > 1 ? img(row - 1, col + N_COMP + i) : 0;
+          int predC = predMode > 1 ? img(row - 1, col + i) : 0;
```
LebedevRI: FWIW, these are out-of-bounds reads for first row and/or last column.

kmilos: Yep, hence all the checks around...

src/librawspeed/decompressors/LJpegDecompressor.cpp
```diff
@@ -217,8 +249,37 @@ template <int N_COMP, bool WeirdWidth> void LJpegDecompressor::decodeN() {
...
+          pred[c] = uint16_t(predA);
+          break;
+        case 2:
+          pred[c] = uint16_t(predB);
+          break;
+        case 3:
+          pred[c] = uint16_t(predC);
+          break;
+        case 4:
+          pred[c] = uint16_t(predA + predB - predC);
```
cytrinox: Predictor 4 is special because A + B - C can overflow uint16_t. The prediction and diff calculation must be done in int32_t. Prediction value can be negative, too.

src/librawspeed/decompressors/LJpegDecompressor.cpp
```diff
@@ -217,8 +249,37 @@ template <int N_COMP, bool WeirdWidth> void LJpegDecompressor::decodeN() {
...
+          break;
+        case 5:
+          pred[c] = uint16_t(predA + ((predB - predC) >> 1));
+          break;
+        case 6:
+          pred[c] = uint16_t(predB + ((predA - predC) >> 1));
+          break;
+        case 7:
+          pred[c] = uint16_t((predA + predB) >> 1);
```
cytrinox: When swiching to int32_t, predictor 7 is not calculated by arithmetic right shift, but with division by 2 (see ITU 81, table H.1).

cytrinox (changes requested):

src/librawspeed/decompressors/LJpegDecompressor.cpp
kmilos: Yep, that's why predA, predB, predC are declared as ints above. Are you saying I should check and clamp to 0 here as well?

src/librawspeed/decompressors/LJpegDecompressor.cpp
kmilos: I was split about this, as indeed the spec does not have same footnote for pred 7 as it does for 5 and 6... What's the difference, it is integer division so there is no rounding, no?

cytrinox:
For prediction > 4 and CFA images, it is common practice to pack two lines into one, so instead of:
````
RGRGRGRG
GBGBGBGB
RGRGRGRG
GBGBGBGB
````
the jpeg image becomes:
````
RGRGRGRGGBGBGBGB
RGRGRGRGGBGBGBGB
````

Because this pattern works better for prediction modes > 4 and component count = 2.

This is currently not handled by the code, as it is assumed that the dimension of the ljpeg stream is the same as the image/tile.

````
[rawspeed] (dummyreg5.dng) void rawspeed::AbstractDngDecompressor::decompress() const, line 217: Too many errors encountered. Giving up. First Error:
virtual void rawspeed::LJpegDecompressor::decodeScan(), line 112: LJpeg frame (1184, 79) is smaller than expected (592, 158)
````
I will generate you a set of DNG files with pred 1-7, comp=2 and packed lines later this day.

File: src/librawspeed/decompressors/LJpegDecompressor.cpp (hunk 3 above)
Comment (cytrinox) [reply src/librawspeed/decompressors/LJpegDecompressor.cpp]: For unsigned int, yes. But for signed int, division and shift is not the same. And the prediction formulas are written for signed integers.

**Comment by kmilos** [comment]
My understanding is that this reshaping happens at a different level of abstraction, and that the number of SOF3 channels at this stage is already correct, and could indeed be different from the CFA channels (i.e. 1). This is how it already works in rawspeed for "normal" DNG lossless compression (2 channels in SOF3, pred 1)

See also my findings in https://github.com/darktable-org/rawspeed/issues/189

The current code works out if you're reshaping by widening (Adobe), but doesn't work if you're reshaping by squeezing (which is what Blackmagic do).

File: src/librawspeed/decompressors/LJpegDecompressor.cpp (hunk 2 above)
Comment (cytrinox) [reply src/librawspeed/decompressors/LJpegDecompressor.cpp]: The reconstruction using diff and prediction value must be done with signed integers, then you can safely cast the result back to uint16_t. If you cast the prediction value first (loosing some bits) and then do simple unsigned integer math, I assume you get wrong results.

**Comment by kmilos** [comment]
In any case, I will not take this further (apart from fixing any overflows/underflows you can demonstrate), as it works on Apple's DNGs which I wanted to give people a potential patch to use on until Roman implements this in Halide.

File: src/librawspeed/decompressors/LJpegDecompressor.cpp (hunk 2 above)
Comment (kmilos) [reply src/librawspeed/decompressors/LJpegDecompressor.cpp]: My reading of itu-t81 is that only the intermediate calculation is to be done as signed int. The predictor is stored and used "modulo 2^16", i.e. unsigned:

> The difference between the prediction value and the input is calculated modulo 2^16. In the decoder the difference is
decoded and added, modulo 2^16, to the prediction.

But I do need to check how the two are added back together earlier indeed.

**Comment by cytrinox** [comment]
> The current code works out if you're reshaping by widening (Adobe), but doesn't work if you're reshaping by squeezing (which is what Blackmagic do).

There are multiple options. First, use comp=1, then w/h of the ljpeg is identical to tile w/h.
For CFA images, it's wise to use comp=2, then lpjeg width is divided by two (squeezing) without reducing the height. And for prediction 4-7, you pack two lines into one, leads to streching (width*2) and (height/2). You can use line packing with just one component, event four is save if the implementation respects it.

Adobe uses prediction 1 and comp=2 and because of that, the width is half but height is unchanged. BlackMagic uses comp=2, too, but because of the prediction mode, SOF3 width becomes (width/2)*2 and SOF3 height (height/2)

File: src/librawspeed/decompressors/LJpegDecompressor.cpp (hunk 3 above)
Comment (kmilos) [reply src/librawspeed/decompressors/LJpegDecompressor.cpp]: For predictor 7, you can't have a negative sum, both quantities (previous samples) are positive.

**Comment by kmilos** [comment]
> BlackMagic uses comp=2

Blackmagic use comp=1 and 2H x W/2

**Comment by LebedevRI** [comment]
(FYI if it isn't obvious i'm in `#pixls.us @ OFTC`)

> <...> until Roman implements this in Halide.

I'm basically having an existential crisis of conciseness here, and that has been happening for some time now.
I *really* don't like how all of our (not just darktable-org's, but more in general in the whole greater open graphics ecosystem)
image processing loops are just hacked together with shitty unreadable unmodifyable unperformant C-like spagetti.
Moving to higher-level abstraction *should* be a huge win, but i'm afraid of being false messiah, and just making things worse.

**Comment by kmilos** [comment]
The sentiment is shared, I don't think it's ideal either. But as you already know there are a limited number of motivated/interested contributors who can properly deal w/ the C++ templated abstraction already laid out here and confident in what LLVM will compile that code to, imagine how many there will be proficient in Halide... In the meantime, the requests to support new codecs keep piling on :/

**Comment by cytrinox** [comment]
> > BlackMagic uses comp=2
> 
> Blackmagic use comp=1 and 2H x W/2

Oh... I just assumed this because otherwise the rawspeed error in #189 wouldn't make sense. Let me check this deeper when I find some time, from the errors and width/height values reported it should be theoretically the same issue as with my line-packed dng files.

**Comment by kmilos** [comment]
The error in #189 is there because it precisely assumes 2 components so W/2 * 2 = W (or in general it assumes you'll always have W/N *N comp = W). Blackmagic don't do that (I guess because of the predictor), so that assumption brakes down.

Mind you, we also have [comp = 4 in the Sony case](https://github.com/darktable-org/darktable/issues/8523#issuecomment-885688715), but I haven't been motivated to do the plumbing for that one, which should work out hopefully better than the Blackmagic case.

File: src/librawspeed/decompressors/LJpegDecompressor.cpp (hunk 3 above)
Comment (kmilos) [reply src/librawspeed/decompressors/LJpegDecompressor.cpp]: I really wished they were clearer on this in the spec...

**Comment by LebedevRI** [comment]
> > (FYI if it isn't obvious i'm in `#pixls.us @ OFTC`)
> > 
> > > <...> until Roman implements this in Halide.
> > 
> > I'm basically having an existential crisis of conciseness here, and that has been happening for some time now. I _really_ don't like how all of our (not just darktable-org's, but more in general in the whole greater open graphics ecosystem) image processing loops are just hacked together with shitty unreadable unmodifyable unperformant C-like spagetti. Moving to higher-level abstraction _should_ be a huge win, but i'm afraid of being false messiah, and just making things worse.
>
> The sentiment is shared, I don't think it's ideal either. But as you already know there are a limited number of motivated/interested contributors who can properly deal w/ the C++ templated abstraction already laid out here and confident in what LLVM will compile that code to, imagine how many there will be proficient in Halide... In the meantime, the requests to support new codecs keep piling on :/

Let me put it it this way. While i have not made a decisions _yet_,
i am starting to believe that this is a "survival of the fittest" question.
As in, either our ecosystem does make that evolutionary jump,
or it does deserve to go extinct.

File: src/librawspeed/decompressors/LJpegDecompressor.cpp (hunk 3 above)
Comment (cytrinox) [reply src/librawspeed/decompressors/LJpegDecompressor.cpp]: You are right, A and B could not be negative, my fault. But I remembered that part of the spec and the *wtf* moment while reading it.

File: src/librawspeed/decompressors/LJpegDecompressor.cpp (hunk 2 above)
Comment (kmilos) [reply src/librawspeed/decompressors/LJpegDecompressor.cpp]: I _think_ all the uint16_t casting takes care of the modulo operations described above by design, and no changes are in fact necessary, but happy to be proven wrong.

Comment (cytrinox) [reply src/librawspeed/decompressors/LJpegDecompressor.cpp]: Well that's difficult to prove. I've added a test case to my implementation where `A = B = 65535` and `C=0` and the compression/decompression verification fails if I calculate this with u16 instead of i32. I can not tell you if some integer magic solves this in your code but it seems legit to assume there is an overflow which leads to wrong results in edge cases.

For me, the edge case was linear raw data (RGB instead of CFA) that uses the full 2^16 value range, writing into DNG with ljpeg compression. For regular CFA images, the input is ~ 2^14, so you never hit overflows.

It's difficult to test as we don't have a ljpeg compressor in rawspeed to write such small tests.

**Comment by cytrinox** [comment]
Here is a set of DNG files with various predictors: https://chaospixel.com/pub/temp/dng_pred_tests.tar.bz2

**Comment by cytrinox** [comment]
@kmilos Just for fun, I've looked into https://github.com/yanburman/dng_sdk/blob/master/source/dng_lossless_jpeg.cpp#L2501 and https://github.com/yanburman/dng_sdk/blob/master/source/dng_lossless_jpeg.cpp#L1428 which comes from the Adobe DNG SDK lossless decoder.
There is a special case when `Px = Ra + ((Rb – Rc)/2)` leads to `Ra + (-1 / 2)` which is different to `Ra + (-1 >> 1)` and I wanted to know how Adobe solved this. They just use signed integers and doing a right shift, which leads to `-1 >> 1 = -1`.

**Comment by kmilos** [comment]
Yet more spaghetti, but I believe this is now functionally correct for the intermediate predictor calculations.

**Comment by kmilos** [comment]
Btw, the Sony case won't work neither w/o some extra reshaping because we're not in the W/4 x 4 comp arrangement assumed, but H/2 x W/2 x 4.

**Comment by LebedevRI** [comment]
FWIW, i've been thinking about this, and i suppose i know a way to implement
the desired support without hurting the current cases.

I haven't checked, but i'm not sure we have all the necessary samples
on RPU though -- i suspect we only have the ones for predictors 2 and 7.
(ignoring the obvious predictor 0), and for sure i don't expect that we have
the full permutation matrix of all 4 variants of component per pixel (1, 2, 3, 4).

So essentially we need at least 8*4 samples. (i may be forgetting some other permutation).
It's obviously not fatal, it just requires someone to write a testcase generator for them.

**Comment by LebedevRI** [comment]
> So essentially we need at least 8*4 samples. (i may be forgetting some other permutation).

Right, restart interval.

**Comment by kmilos** [comment]
> I haven't checked, but i'm not sure we have all the necessary samples
> on RPU though -- i suspect we only have the ones for predictors 2 and 7.

Blackmagic DNGs have predictor 6, but have another problem (component arrangement as mentioned above and in https://github.com/darktable-org/rawspeed/issues/189) preventing its testing...

So it might be an idea to generalize the component arrangement first (1x1, 1x2, 2x1, 2x2), thus ticking Sony lossless off the list?

Or perhaps you prefer to handle both component arrangement and predictors in one go? I hope @cytrinox could help with the 4*8 test vectors then... 🙏

**Comment by kmilos** [comment]
> > I haven't checked, but i'm not sure we have all the necessary samples
> > on RPU though -- i suspect we only have the ones for predictors 2 and 7.
> 
> Blackmagic DNGs have predictor 6, but have another problem as mentioned preventing the testing today...

**Comment by LebedevRI** [comment]
> So it might be an idea to generalize the component arrangement first (1x1, 1x2, 2x1, 2x2), thus ticking Sony lossless off the list?

Sony FUBAR'ed their "LJpeg" implementation to the point of it no longer being a proper LJpeg.
We won't be able to support it in a generic LJpeg decompressor.

**Comment by kmilos** [comment]
> to the point of it no longer being a proper LJpeg

Why do you think so? AFAICT it is, just that the component scatter/gather is different from the Adobe one... I.e. for a 2x2 Bayer CFA Adobe does 2 components like this

**1 2** 1 2 ...
**1 2** 1 2 ...

While Sony does 4 components like this:

**1 2** 1 2 ...
**3 4** 3 4 ...

Each component in both Adobe and Sony cases seems to be regular LJpeg SOF3 (w/ already supported predictor 1)... The arrangement of components has nothing to do w/ LJpeg itself, and [current LJpeg code seems to work fine once the arrangement is taken into account..](https://github.com/darktable-org/rawspeed/pull/386)

What they FUBAR'ed is on the TIFF container level by having both Tile* and Strip* tags simultaneously, and making ImageWidth/Height an integer of tile size, when it doesn't have to be ([which is why I advocated for the absolute crop for these models covering all modes instead of a relative, negative one](https://github.com/darktable-org/rawspeed/pull/274#issue-891092324); but this then makes APS-C crop more complicated so it's a lose-lose either way)...

**Comment by kmilos** [comment]
And BlackMagic do something even weirder, from a 2x2 Bayer CFA:

1 1 ...
1 1 ...

to a single components LJpeg like this:

1 1 1 1

(w/ predictor 6)

**Comment by LebedevRI** [comment]
> and [current LJpeg code seems to work fine once the arrangement is taken into account..](https://github.com/darktable-org/rawspeed/pull/386)

That's my point exactly actually. As long as that `sonyArrange_` can't be
deduced from the LJpeg container itself, it's broken.

**Comment by kmilos** [comment]
> As long as that `sonyArrange_` can't be
> deduced from the LJpeg container itself, it's broken.

Sure, and I already mentioned there that should be generalized - one just needs to parse the SOF3 header for the no. of components and their frame size, and compare to the TIFF CFA tile size, and that gives the arrangement:

E.g.

Sony case: TIFF tile is 512x512, SOF3 is 4 components of 256x256

Adobe case: TIFF tile is 256x256, SOF3 is 2 components of 256x128

BlackMagic case: TIFF tile is 512x512, SOF3 is 1 component of 256x1024

The arrangement is calculated from the TIFF tile HxW divided by SOF3 YxX, and to check conformance, one must have N_comp\*Y\*X = H\*W (I think already checked).

Both the Sony and BlackMagic SOF3 headers are correct in what they implement AFAICT, so nothing is really at odds here, "just" missing implementation in rawspeed...

**Comment by LebedevRI** [comment]
I see.
So for cpp=2, we have two possible layouts:
```
x x
```
and
```
x
x
```
and for cpp=4 we have three:
```
x x x x
```
```
x x
x x
```
```
x
x
x
x
```
.. that multiplies the needed sample set by another x2 or so.

kmilos:
> that multiplies the needed sample set by another x2 or so

😮

kmilos:
> and for cpp=4 we have three

Not quite. For cpp=4 you can have two:

1 2
3 4

or

1 3
2 4

i.e. row or column major scatter of the 2x2 Bayer into 4 separate component planes. Like this for row major: 

image

(In theory, one can of course have all the other ways of ordering 4 components, 4! in total, but that would be a bit silly...)

For cpp=1 you could have 2 (well, 3 if also you count identity, i.e. no rearrangement, just like you illustrated above).

LebedevRI:
Sorry, not following the logic there. I can see how the arrangements i listed can be specified
via different frame sizes, but not the other orders.

kmilos:
> Sorry, not following the logic there. I can see how the arrangements i listed can be specified
> via different frame sizes, but not the other orders.

Np, I'll try to do a clearer drawing for the three (Adobe, Sony, and BlackMagic) cases a bit later and add here.

kmilos:
@LebedevRI I hope the following illustration of the scatter/gather schemes makes it clearer:

dng_ljpeg

These are all perfectly valid w.r.t. ITU-T T.81. The components highlighted in yellow make up one MCU, and the dotted ones are used as predictors (usually 1, i.e. just horizontal, but 6 for Blackmagic/CinemaDNG case).

(There is also the trivial case of DNG spp=3 LinearRaw with 1:1 mapping to SOF3 Nf=3, which is already supported for predictor 1, but not predictor 7 using the row above like Apple do.)

LebedevRI:
One thing i can say right away -- if nothing, it's a pretty nice infographics!

artemist:
~In the JFIF SOF3 Sony advertises blocks as 2W x H/2.~ Wait no you're right I was thinking of the number of samples I have to write per "row" as a consequence of decoding in one pass.

 I have a somewhat dirty PR in #386 that supports the Sony arrangement and might support blackmagic arrangement if I understand it correctly but I can't test.

kmilos:
I think this was the last time I rebased this - the new code structure that imposes use of a single row only (effectively supporting only predictor 1) makes working around this even uglier than before, if that's possible 😮

PR #344, opened 2022-02-14
Update compiler-warnings-gcc.cmake

squid-f:
typo mistake, I assume. Option old-stlyle-casts doesn't exit. It should rather be old-style-cast

### DIFF
--- a/cmake/compiler-warnings-gcc.cmake
+++ b/cmake/compiler-warnings-gcc.cmake
@@ -13,7 +13,7 @@ set (GCC_WARNING_FLAGS
   "extra-semi"
   "format=2"
 #  "missing-prototypes"
-  "old-style-casts"
+  "old-style-cast"
   "pointer-arith"
 #  "strict-prototypes"
 #  "suggest-attribute=const"


squid-f:
Hi. Not sure why the checks below have failed. 
Personally, I have packaged darktable with rawspeed bundled, using gcc 10.3.0, with this change.

squid-f:
Hi. Not sure why the checks below have failed. Personally, I have packaged darktable with rawspeed bundled, using gcc 10.3.0, with this change ; mageia 8 (rpm distro).

LebedevRI:
It fails because it the now-enabled diagnostic triggers.
It's sorta expected, and unfortunate, because i'm not sure how to deal with some of those cases.
It worked for you because dt disables all warnings for rawspeed.

It might be better to, instead of fixing the typo, to instead completely drop the misspelled diag flag.

squid-f:
Ok. As it didn’t have any effect so far, as misspelled, it should not hurt in anything to remove it. I will issue another pull request.

PR #345, opened 2022-02-15
remove misspelled flag

squid-f:
Based on https://github.com/darktable-org/rawspeed/pull/344 , remove misspelled old-style-cast flag instead of correcting its spelling.


LebedevRI:
Thanks.

PR #347, opened 2022-02-24
Nikon Z 9 support (lossless only)

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/10889 and https://github.com/darktable-org/darktable/issues/11067

@LebedevRI This is a stub only, I have not verified the CFA pattern nor the crop.

There is also no support (nor detection & exception throwing!) added for the new "High Efficiency" codec.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -4379,6 +4379,24 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="NIKON CORPORATION" model="NIKON Z 9" mode="14bit-compressed">
+		<ID make="Nikon" model="Z 9">Nikon Z 9</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
+		</CFA>
+		<Crop x="0" y="0" width="0" height="0"/>
+		<Sensor black="1008" white="15892"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">13389 -6049 -1441</ColorMatrixRow>
+				<ColorMatrixRow plane="1">-4544 12757 1969</ColorMatrixRow>
+				<ColorMatrixRow plane="2">229 498 7390</ColorMatrixRow>
+			</ColorMatrix>
+		</ColorMatrices>
+	</Camera>
 	<Camera make="NIKON CORPORATION" model="NIKON Z 50" mode="12bit-compressed">
 		<ID make="Nikon" model="Z 50">Nikon Z 50</ID>
 		<CFA width="2" height="2">


kmilos:
Just a little bit of info I have found re the new HE modes:

They do not have the usual [0x96 NEFLinearizationTable tag](https://github.com/darktable-org/rawspeed/blob/0c854a18ac6f664e374a28623ce10d1480c8c6c2/src/librawspeed/decoders/NefDecoder.cpp#L117-L121) in the MakerNote

The fallback 0x8c is junk/unknown, so any further parsing crashes rawspeed

There seems to be a new 0x51 tag instead in Z 9 files, here's the comparison of lossless vs HE exiftool output:

image

Note that the NEFCompression field of that tag seems to correspond to tag [0x93](https://exiftool.org/TagNames/Nikon.html#NEFCompression) (which is absent for Z 9).

There is also a [new (unknown) 0xc7d5 NEFInfo IFD](https://exiftool.org/TagNames/Nikon.html#NEFInfo) in the Z bodies as well.

kmilos:
The RPU lossless sample loads fine in dt master and there are no artifacts at the edges, so this might be good to go w/o the crop (like the Z 7_2).

Loading the HE files luckily does not crash dt (the unsupported message is displayed), so additional checks might not be needed after all.

colin-adams:
This seems to be stuck. Can I help at all (I'm a software engineer by profession)? Looking at the Details of the failing check, it seems to be a python module that is no longer available, but a replacement or two are indicated, so it ought to be easy to fix.

kmilos:
Cleared that hurdle, but a new one appears of course, and I have zero interest in addressing it: CodeChecker (external repository) requires older psutil 5.8.0 exactly and a) wheel is not available for Python 3.10 (psutil 5.9.0 is already installed) so it tries to build 5.8.0 from source, and as a consequence b) setuptools can't figure out it needs to use the LLVM linker, it assumes GCC instead.

In any case, this is not blocking this particular PR, as there is no code change.

colin-adams:
What's holding this up? Anything I can do to help?

ValZapod:
So this will add ticoraw for 2.0 firmware? https://www.nikoncafe.com/threads/nikon-z9-raw-file-support.331229/

Sturmflut:
Any updates? I've verified the settings in the XML with my own Z 9, they're fine.

Sturmflut:
> So this will add ticoraw for upcoming 2.0 firmware, like in Canon and Red?

No, this only adds support for the sensor in the Z 9.

data/cameras.xml
```diff
@@ -4379,6 +4379,24 @@
...
 	</Camera>
+	<Camera make="NIKON CORPORATION" model="NIKON Z 9" mode="14bit-compressed">
+		<ID make="Nikon" model="Z 9">Nikon Z 9</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
+		</CFA>
+		<Crop x="0" y="0" width="0" height="0"/>
```
LebedevRI, line 4390: I can not tell from existing samples, but it looks like this needs a bit of a crop on the right.

colin-adams:
I don't know if my remark here is relevant, but I have been using the Z9
with darktable since the beginning of August, and I haven't ever seen any
superfluous columns of pixels on the right (or anywhere else for that
matter).

LebedevRI:
It's basically off by at most 6-12 pixels, not really noticeable unless you really look.

LebedevRI:
@kmilos thank you!

robcampbell95:
What does it mean that the nikon_z9 branch was deleted?  There is still no official support for Nikon Z9?  I've updated my cameras.xml file but my path is /usr/share/darktable/rawspeed/cameras.xml not /usr/share/darktable/rawspeed/data/cameras.xml which seems to be the guidance (data/cameras.xml) for the file.

kmilos:
The Z 9 is supported in the just released darktable 4.2 OOTB, except the HE modes.

PR #351, opened 2022-04-01
`ColorFilterArray::shiftDcrawFilter()`: avoid out-of-bounds shifts

maruncz:
when `y==0`, `(filter << (32 - y))` shifts by 32 is too large for 32-bit integer

### DIFF
--- a/src/librawspeed/metadata/ColorFilterArray.cpp
+++ b/src/librawspeed/metadata/ColorFilterArray.cpp
@@ -153,9 +153,8 @@ uint32_t ColorFilterArray::shiftDcrawFilter(uint32_t filter, int x, int y) {
   // A shift in y direction means rotating the whole int by 4 bits.
   y *= 4;
   y = y >= 0 ? y % 32 : 32 - ((-y) % 32);
-  if (y != 0) {
+  if (y != 0)
     filter = (filter >> y) | (filter << (32 - y));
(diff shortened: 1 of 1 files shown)


src/librawspeed/metadata/ColorFilterArray.cpp
```diff
@@ -153,7 +153,9 @@ uint32_t ColorFilterArray::shiftDcrawFilter(uint32_t filter, int x, int y) {
   // A shift in y direction means rotating the whole int by 4 bits.
   y *= 4;
   y = y >= 0 ? y % 32 : 32 - ((-y) % 32);
-  filter = (filter >> y) | (filter << (32 - y));
+  if (y != 0) {
```
LebedevRI: Isn't this dead code due to https://github.com/darktable-org/rawspeed/blob/1a1b723420bd3c923b0ed242287e6c615cd87af4/src/librawspeed/metadata/ColorFilterArray.cpp#L150-L151 ?

maruncz: if y=8 then y*4=32

LebedevRI: Please elide the braces

LebedevRI:
Thanks.

PR #353, opened 2022-05-05
LJpegDecompressor: don't look for markers beyond end of buffer

kmilos:
Provides a chance to raise the desired exception and help w/ debugging rather than the generic "Out of bounds" one seen in https://github.com/darktable-org/darktable/issues/11695

### DIFF
--- a/src/librawspeed/decompressors/AbstractLJpegDecompressor.cpp
+++ b/src/librawspeed/decompressors/AbstractLJpegDecompressor.cpp
@@ -30,6 +30,7 @@
 #include <array>                                // for array
 #include <cassert>                              // for assert
 #include <memory>                               // for unique_ptr, make_unique
+#include <optional>                             // for optional
 #include <utility>                              // for move
 #include <vector>                               // for vector
 
@@ -250,17 +251,28 @@ void AbstractLJpegDecompressor::parseDHT(ByteStream dht) {
 }
 
 JpegMarker AbstractLJpegDecompressor::getNextMarker(bool allowskip) {
-  uint8_t c0;
-  uint8_t c1 = input.getByte();
-  do {
-    c0 = c1;
-    c1 = input.getByte();
-  } while (allowskip && !(c0 == 0xFF && c1 != 0 && c1 != 0xFF));
+  auto peekMarker = [&]() -> std::optional<JpegMarker> {
+    uint8_t c0 = input.peekByte(0);
+    uint8_t c1 = input.peekByte(1);
 
-  if (!(c0 == 0xFF && c1 != 0 && c1 != 0xFF))
-    ThrowRDE("(Noskip) Expected marker not found. Probably corrupt file.");
+    if (c0 == 0xFF && c1 != 0 && c1 != 0xFF)
+      return static_cast<JpegMarker>(c1);
+    return {};
+  };
 
-  return static_cast<JpegMarker>(c1);
+  while (input.getRemainSize() >= 2) {
+    if (std::optional<JpegMarker> m = peekMarker()) {
+      input.skipBytes(2); // Skip the bytes we've just consumed.
+      return *m;
+    }
+    // Marker not found. Might there be leading padding bytes?
+    if (!allowskip)
+      break; // Nope, give up.
+    // Advance by a single(!) byte and try again.
+    input.skipBytes(1);
+  }
+
+  ThrowRDE("(Noskip) Expected marker not found. Probably corrupt file.");
 }
 
 } // namespace rawspeed


LebedevRI:
IIRC , that happens when looking for the `EOI` (end of image) marker
*after* we've already successfully found `SOS`(start of scan) marker,
and decoded the image, correct?

I've seen that, many Hasselblad images are affected. I don't really know what to do about it.

What we _could_ do, is to *not* look for `EOI`, but immediately return after handling `SOS`,
and then those images would be supported, without safety issues even.

Perhaps i need to re-re-read the [spec (itu t81 (09/92))](https://github.com/darktable-org/rawspeed/files/8660295/itu-t81.pdf) (is there something newer?),
maybe that's the expected behavior/handling?

kmilos:
> IIRC , that happens when looking for the `EOI` (end of image) marker
> _after_ we've already successfully found `SOS`(start of scan) marker,
> and decoded the image, correct?

Exactly, as was the case with some those Leica files in the linked issue - SOS decoding was successful, but EOI is not found and then we fail.

> What we _could_ do, is to _not_ look for `EOI`, but immediately return after handling `SOS`,
> and then those images would be supported, without safety issues even.

Sure, but don't know about the safety aspect - I don't think there's any out of bounds checking of the (same?) buffer in SOS parsing like it is when parsing markers (data buffer created from wrong StripByteCounts in that failing case), but this would would mean a security risk exists as is today. If it uses a different buffer/stream and checks it somewhere else, then it's all fine I guess...

kmilos:
The spec seems to be clear on this:

"the non-hierarchical interchange format shall begin with an SOI marker, shall contain one frame, and shall end with an EOI marker"

src/librawspeed/decompressors/AbstractLJpegDecompressor.cpp
```diff
@@ -255,7 +255,7 @@ JpegMarker AbstractLJpegDecompressor::getNextMarker(bool allowskip) {
   do {
     c0 = c1;
     c1 = input.getByte();
-  } while (allowskip && !(c0 == 0xFF && c1 != 0 && c1 != 0xFF));
+  } while (allowskip && !(c0 == 0xFF && c1 != 0 && c1 != 0xFF) && input.getPosition() < input.getSize());
```
LebedevRI: How about something like this:
```c++
JpegMarker AbstractLJpegDecompressor::getNextMarker(bool allowskip) {
  auto peekMarker = [&]() -> std::optional<JpegMarker> {
    uint8_t c0 = input.peekByte(0);
    uint8_t c1 = input.peekByte(1);

    if (c0 == 0xFF && c1 != 0 && c1 != 0xFF)
      return static_cast<JpegMarker>(c1);
    return {};
  };

  while (input.getRemainSize() >= 2) {
    if (std::optional<JpegMarker> m = peekMarker()) {
      input.skipBytes(2); // Skip the bytes we've just consumed.
      return *m;
    }
    // Marker not found. Might there be leading padding bytes?
    if (!allowskip)
      break; // Nope, give up.
    // Advance by a single(!) byte and try again.
    input.skipBytes(1);
  }

  ThrowRDE("Expected marker not found. Probably corrupt file.");
}

```

LebedevRI:
> > What we _could_ do, is to _not_ look for `EOI`, but immediately return after handling `SOS`,
> > and then those images would be supported, without safety issues even.
> 
> Sure, but don't know about the safety aspect - I don't think there's any out of bounds checking of the (same?) buffer in SOS parsing like it is when parsing markers (data buffer created from wrong StripByteCounts in that failing case), but this would would mean a security risk exists as is today. If it uses a different buffer/stream and checks it somewhere else, then it's all fine I guess...

I'm confident that there is sufficient bounds checking.
And even if there isn't, that's just a bug that would need to be fixed.

src/librawspeed/decompressors/AbstractLJpegDecompressor.cpp
kmilos: Thanks, better indeed.

LebedevRI: (awaiting diff update)

kmilos: Done

kmilos:
I get the more informative output now from e.g. `rs-identify`:

```
ERROR: [rawspeed] C:/msys64/home/kmilos/rawspeed/src/librawspeed/decompressors/AbstractDngDecompressor.cpp:217: void rawspeed::AbstractDngDecompressor::decompress() const: Too many errors encountered. Giving up. First Error:
C:/msys64/home/kmilos/rawspeed/src/librawspeed/decompressors/AbstractLJpegDecompressor.cpp:275: rawspeed::JpegMarker rawspeed::AbstractLJpegDecompressor::getNextMarker(bool): (Noskip) Expected marker not found. Probably corrupt file.
```

LebedevRI:
Thank you!

PR #355, opened 2022-05-10
Add support for DNG DefaultScale tag

cytrinox:
Some cameras like Nikon D1X has non-square pixels.
For the NEF file, this is controlled in cameras.xml with:

    <Hint name="pixel_aspect_ratio" value="0.5"/>

When NEF is converted to DNG, the required scale is encoded in DEFAULTSCALE tag. So we can derive the aspect ratio from
this tag.

Attached is a DNG file converted with AdobeDNG from D1X NEF. I've implemented it the same way in dnglab.

[NIKON D1X_ISO_200_12bits_Uncompressed.zip](https://github.com/darktable-org/rawspeed/files/8658630/NIKON.D1X_ISO_200_12bits_Uncompressed.zip)


LebedevRI:
Thank you!
Looks fine i guess.

PR #362, opened 2022-05-19
Add Olympus E-M10 Mark IIIs alias

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/11803

Reviews also confirm these are identical wrt to sensor parameters, e.g. https://www.dpreview.com/news/0621457211/olympus-releases-the-lightly-updated-om-d-e-m10-mark-iiis

There is a sample on RPU and loads ok after this addition.


LebedevRI:
@kmilos thank you!

PR #365, opened 2022-05-29
Fuji" Fix 16 bit "lossless" support for GFX 100

### LINKED ISSUE #228: Fujifilm GFX100 16-bit "lossless" compressed raw support
Support for that camera was requested in https://github.com/darktable-org/darktable/issues/4323 by @mgrayson3
It appears, only the 16-bit compressed raws don't work out-of-the-box.

cytrinox:
decBits limit must be 14 instead of 12 (still compatible with 14 bit images). See libraw bitDiff() function for reference.

fixes #228


LebedevRI:
Thank you!

PR #339, opened 2022-01-10
Fujifilm X-T30 II support

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/10755

@LebedevRI ~~This is a stub only, I have not verified the CFA pattern nor the crop.~~

PR #369, opened 2022-06-03
Fix shiftLeft() and shiftDown() for CFAs larger than 2x2

cytrinox:
If CFA is larger than 2x2 (like X-Trans with 6x6 for example), the shift needs to be corrected by subtracting from CFA width/height.

You can't hit this bug with 2x2 CFAs because `1 == 2 - 1`.

### DIFF
--- a/src/librawspeed/metadata/ColorFilterArray.cpp
+++ b/src/librawspeed/metadata/ColorFilterArray.cpp
@@ -89,8 +89,8 @@ void ColorFilterArray::shiftLeft(int n) {
     ThrowRDE("No CFA size set (or set to zero)");
 
   writeLog(DEBUG_PRIO::EXTRA, "Shift left:%d", n);
-  n %= size.x;
-  if (n == 0)
+  n = size.x - (n % size.x);
+  if (n == 0 || n == size.x)
     return;
 
   vector<CFAColor> tmp(size.area());
@@ -107,8 +107,8 @@ void ColorFilterArray::shiftDown(int n) {
     ThrowRDE("No CFA size set (or set to zero)");
 
   writeLog(DEBUG_PRIO::EXTRA, "Shift down:%d", n);
-  n %= size.y;
-  if (n == 0)
+  n = size.y - (n % size.y);
+  if (n == 0 || n == size.y)
     return;
   vector<CFAColor> tmp(size.area());
   for (int y = 0; y < size.y; ++y) {


cytrinox:
Here is an example DNG file that triggers the bug: https://chaospixel.com/pub/misc/X-T30II_active_area_shifted.dng

The ActiveArea begins not at CFA boundary:

````
Active Area                     : 5 0 4181 6252
````
The correct n value is `1 (by 6 - (5 % 6)` but old code calculates `5` which produces an invalid shift.

**Comment by cytrinox** [comment]
Just confused by the name: shiftDown() don't shift down but up, but I don't want to touch the name :smile:

**Comment by LebedevRI** [comment]
I don't understand the problem, but i guess we ask it to reinterpret it the CFA so it's point of origin
has been shifted inwards towards the center of the image by the shift amounts, and it does something else?
Really, the actual bug is that we do that in the first place.

That being said, can you add some basic test with 3x3 CFA (e.g. all-RED, single GREEN in the centre)
by duplicating https://github.com/darktable-org/rawspeed/blob/dbe8a6960752612fdd33e05b3eb0b594c5b5775f/test/librawspeed/metadata/ColorFilterArrayTest.cpp#L260 and asserting that the middle element goes where expected for various shift combinations?

**Review by LebedevRI — CHANGES_REQUESTED** [review]
Actually, please explain what the actual problem is.
This test passes currently:
```
TEST(ColorFilterArrayTestBasic, shift3x3Filter) {
  ColorFilterArray cfaOrig;
  cfaOrig.setSize({3, 3});
  cfaOrig.setColorAt({1, 1}, CFAColor::RED);

  {
    ColorFilterArray c = cfaOrig;
    c.shiftLeft(1);
    ASSERT_EQ(c.getColorAt(0, 1), cfaOrig.getColorAt(1, 1));
  }
  {
    ColorFilterArray c = cfaOrig;
    c.shiftDown(1);
    ASSERT_EQ(c.getColorAt(1, 0), cfaOrig.getColorAt(1, 1));
  }
}
```
and with this patch the shift direction is inversed.

LebedevRI:
I've added a test, and i suspect the users are wrong.
I can't test in darktable because it does not build for me any longer.

cytrinox:
I've produced multiple DNG files (for all xtrans RAFs in raw samples) with ActiveArea that is not aligned at CFA boundary, for example `5 0`. All these DNGs opens perfectly in Rawtherapee and Lightroom but not in Darktable. In Darktable, there is clearly a wrong CFA applied and the colors are totally wrong.
I've debugged this with prtinting cfa.asString() after all the shift corrections in rawspeed and it was cleary the wrong layout. This PR fixed the issue for me, but maybe the problem root is somewhere else?

But I'm sure the math in the shift functions is wrong. Let me show an example:

If the initial Xtrans pattern at (0, 0) is:

````
row[0] AAAAAA
row[1] BBBBBB
row[2] CCCCCC
row[3] DDDDDD
row[4] EEEEEE
row[5] FFFFFF
````
and the ActiveArea is `5 0 ? ?` (shift 5 on y axis), the encoded pattern in DNG becomes this (because CFAPattern in DNG must be aligned at ActiveArea):

````
row[0] FFFFFF
row[1] AAAAAA
row[2] BBBBBB
row[3] CCCCCC
row[4] DDDDDD
row[5] EEEEEE
````

Rawspeed tries to revert that process when reading DNGs to get the initial alignment at (0, 0). For this `shiftDown(n=5)` is called. Because of `n %= size.y` we get `5 % 6 => 5`. Which means the new cfa row[0] gets EEEEEE assigned, which is clearly wrong.
The attempt of this PR was to fix this so shiftDown(n=5) really shifts the CFA pattern by 5 rows the restore the correct pattern.

I'm still confused by the shiftDown name, but I think this is the main problem. After thinking again about it, the function behaves like it says: it shifts the pattern down, so by `n = 5` the new pattern starts with EEEEEE. As rawspeed tries to *revert* the ActiveArea shift, we need a second function `shiftUp()` with my change `n = size.y - (n % size.y)` and call this instead.
Do you agree with this or I'm totally wrong..?

cytrinox:
Another (much simpler solution would be): 

Just changing to negative values:
https://github.com/darktable-org/rawspeed/blob/586d1492492d9df8a7683401f3017c54a770ffce/src/librawspeed/decoders/DngDecoder.cpp#L209-L210

````
 mRaw->cfa.shiftRight(-aa[1]); 
 mRaw->cfa.shiftDown(-aa[0]); 
````

Then the CFA is correct, too, without the need of bungle around shiftRight and shiftDown. I wasn't aware of the fact we can put negative numbers here safely.

kmilos:
Can you please verify this can also address https://github.com/darktable-org/rawspeed/pull/339 please?

It seems the CFA pattern shifts in compressed Vs uncompressed mode?

cytrinox:
@kmilos checked and responded at #339. There is no difference between comp/uncomp. The CFA definition for X-T30 II in your PR needs to be corrected then it should work.

LebedevRI:
> Another (much simpler solution would be):
> 
> Just changing to negative values:
> 
> https://github.com/darktable-org/rawspeed/blob/586d1492492d9df8a7683401f3017c54a770ffce/src/librawspeed/decoders/DngDecoder.cpp#L209-L210
> 
> ```
>  mRaw->cfa.shiftRight(-aa[1]); 
>  mRaw->cfa.shiftDown(-aa[0]); 
> ```
> 
> Then the CFA is correct, too, without the need of bungle around shiftRight and shiftDown. I wasn't aware of the fact we can put negative numbers here safely.

Yes, this was the solution i was thinking of. Let's go with it. But then other uses may still be wrong. (in dt e.g.)
While there, let's unmangle the function names, they should be `rotateLeft` and `rotateUp`, i messed that up.

cytrinox:
Superseded by #370

Issue #375, opened 2022-07-24
Some different white points at different ISO, Canon EOS 1Ds


PeterWem:
I used dngmeta.rb for my 1Ds raw files and I got the following:

```
<Camera make="Canon" model="Canon EOS-1DS">
		<ID make="" model="">Canon EOS-1Ds</ID>
		<Sensor black="0" white="3617"/>
		<Sensor black="0" white="3500" iso_list="50"/>
		<Sensor black="0" white="3615" iso_list="100"/>
		<Sensor black="0" white="3614" iso_list="125"/>
		<Sensor black="0" white="3611" iso_list="160"/>
		<Sensor black="0" white="3616" iso_list="250 320"/>
		<Sensor black="0" white="3621" iso_list="500"/>
		<Sensor black="0" white="3625" iso_list="800"/>
		<Sensor black="0" white="3623" iso_list="1000"/>
		<Sensor black="0" white="3622" iso_list="1250"/>
	</Camera>
```

![Skärmbild från 2022-07-24 23-12-55](https://user-images.githubusercontent.com/34200649/180666023-59559b2e-b52a-4feb-9baa-2445c35e6a4d.png)

I also checked the color matrix. I got:
$ exiv2 -pt 20220724E53W1894.dng 2>/dev/null | grep CalibrationIlluminant2
Exif.Image.CalibrationIlluminant2            Short       1  D65

and

$ exiv2 -pt 20220724E53W1895.dng 2>/dev/null | grep ColorMatrix2
Exif.Image.ColorMatrix2                      SRational   9  3925/10000 4060/10000 -1739/10000 -8973/10000 16552/10000 2545/10000 -3287/10000 3945/10000 8243/10000

PR #378, opened 2022-07-26
Update cameras.xml

PeterWem:
Added white points/ISO for 1Ds and renewed its color matrix. Adobe DNG Converter 14.2 was used.
This PR solves https://github.com/darktable-org/rawspeed/issues/375

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -1784,7 +1784,6 @@
 			<Color x="1" y="1">GREEN</Color>
 		</CFA>
 		<Crop x="0" y="0" width="0" height="0"/>
-		<Sensor black="0" white="3617"/>
 		<Sensor black="0" white="3500" iso_list="50"/>
 		<Sensor black="0" white="3615" iso_list="100"/>
 		<Sensor black="0" white="3614" iso_list="125"/>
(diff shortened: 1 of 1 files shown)


data/cameras.xml
```diff
@@ -1784,12 +1784,21 @@
 			<Color x="1" y="1">GREEN</Color>
 		</CFA>
 		<Crop x="0" y="0" width="0" height="0"/>
-		<Sensor black="0" white="3500"/>
-		<ColorMatrices>
-			<ColorMatrix planes="3">
-				<ColorMatrixRow plane="0">4374 3631 -1743</ColorMatrixRow>
-				<ColorMatrixRow plane="1">-7520 15212 2472</ColorMatrixRow>
-				<ColorMatrixRow plane="2">-2892 3632 8161</ColorMatrixRow>
+		<Sensor black="0" white="3617"/>
```
kmilos: I think this should go last (fallback value if ISO not matched): https://github.com/darktable-org/rawspeed/tree/develop/data#sensor-info

LebedevRI:
If somebody really cares about this, just implement reading of those values from the metadata.

PeterWem:
Hm, I will close this PR, redo it and make it less messy. 

Reading from Exif would be the best solution. It seems Canon has the black/white point in many different tags depending on camera model?

kmilos:
Wrt Canon, black and white levels are indeed all over the place depending on MakerNote version: https://exiftool.org/TagNames/Canon.html

PR #379, opened 2022-07-26
Updated 1Ds color matrix and sensor white

PeterWem:
Added white points/ISO for 1Ds and renewed its color matrix. Adobe DNG Converter 14.2 was used.
This PR solves #375

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -1784,12 +1784,21 @@
 			<Color x="1" y="1">GREEN</Color>
 		</CFA>
 		<Crop x="0" y="0" width="0" height="0"/>
-		<Sensor black="0" white="3500"/>
+		<Sensor black="0" white="3500" iso_list="50"/>
+		<Sensor black="0" white="3615" iso_list="100"/>
+		<Sensor black="0" white="3614" iso_list="125"/>
+		<Sensor black="0" white="3611" iso_list="160"/>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Yep, incorrect black/white levels is a rather common problem.
We are really ought to stop hardcoding it, and reading it from EXIF instead.

iiborg:
EXIF contains programmed black level, not the real one, especially for the cameras where the black level is calculated from the optical black, per row / per column.

PeterWem:
> Yep, incorrect black/white levels is a rather common problem. We are really ought to stop hardcoding it, and reading it from EXIF instead.

Black level would be better from optical black, wouldn't it? About white point we don't have anything to measure, I assume?

Some black level tags for Canon, but Canon changes all the time https://www.dpreview.com/forums/thread/3789550?page=2

kmilos:
There are many black level and white level tag locations for Canon cameras depending on model and makernote version:

https://exiftool.org/TagNames/Canon.html

iiborg:
> Black level would be better from optical black, wouldn't it? 

Absolutely. That helps preventing pattern noise.

> About white point we don't have anything to measure, I assume?

White point is calculated from raw data if raw data is clipped.

LebedevRI:
@PeterWem thank you!

PR #380, opened 2022-09-06
DngDecoder: improve ColorMatrix extraction

kmilos:
See https://discuss.pixls.us/t/dng-hdr-created-in-dt-has-unsupported-input-profile/32528

Not really implementing "rationals support" throughout (I'll bring the comment back in), but should be an ok workaround (better than returning a zero matrix).

### DIFF
--- a/src/librawspeed/decoders/DngDecoder.cpp
+++ b/src/librawspeed/decoders/DngDecoder.cpp
@@ -637,28 +637,41 @@ void DngDecoder::decodeMetaDataInternal(const CameraMetaData* meta) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


LebedevRI:
Thank you!

kmilos:
Thanks for merging and extending.

I guess this is now no longer compatible w/ https://github.com/darktable-org/darktable/blob/5a1a1845fa32378daa9db4b5c3a5705051b1ac49/src/common/imageio_rawspeed.cc#L268 as the division will have already been done when the float cast is requested?

LebedevRI:
Right, it is expected that places like that will have to be updated.

PR #385, opened 2022-10-01
ARW: Get black level and white point from EXIF

artemist:
This implements the change suggested in #349 so that Sony cameras with white and black level encoded in their metadata do not require sensor rules in cameras.xml

Unfortunately I haven't been able to test on high ISO samples because I cannot find samples at ISO 102400 with a properly set camera model name and I do not currently own an ILCE-7M4


LebedevRI:
@artemist Thank you!

PR #388, opened 2022-10-13
Read black levels from Nikon makernotes

kmilos:
Using https://exiftool.org/TagNames/Nikon.html

### DIFF
--- a/src/librawspeed/decoders/NefDecoder.cpp
+++ b/src/librawspeed/decoders/NefDecoder.cpp
@@ -599,6 +599,20 @@ void NefDecoder::decodeMetaDataInternal(const CameraMetaData* meta) {
   auto id = mRootIFD->getID();
   std::string mode = getMode();
   std::string extended_mode = getExtendedMode(mode);
+
+  // Read black levels (seem to be recorded for 14bit always)
+  if (mRootIFD->hasEntryRecursive(TiffTag::NIKON_BLACKLEVEL)) {
+    const TiffEntry* bl =
+        mRootIFD->getEntryRecursive(TiffTag::NIKON_BLACKLEVEL);
+    if (bl->count != 4)
+      ThrowRDE("BlackLevel has %d entries instead of 4", bl->count);
+    const int sh = mode.rfind("12bit", 0) == 0 ? 2 : 0;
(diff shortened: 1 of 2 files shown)


src/librawspeed/decoders/NefDecoder.cpp
```diff
@@ -607,6 +607,19 @@ void NefDecoder::decodeMetaDataInternal(const CameraMetaData* meta) {
     setMetaData(meta, id, "", iso);
   }
 
+  // Read any additional black levels, but only if not already overriden
+  if (!hints.has("override_cfa_black") &&
```
LebedevRI: This should not be here.

src/librawspeed/decoders/NefDecoder.cpp
```diff
@@ -607,6 +607,19 @@ void NefDecoder::decodeMetaDataInternal(const CameraMetaData* meta) {
...
+      mRootIFD->hasEntryRecursive(static_cast<TiffTag>(0x003d))) {
```
LebedevRI: Please use named Tiff Tag enum, add one if it's not there yet.

kmilos:
Ah, this won't work, as I just read for the D5600 samples (ISO 200)

```
BlackLevel                      : 600 600 600 600
```

in both 12-bit and 14-bit modes...

Converting to draft until someone comes up with an additional fix...

kmilos:
One possibility is just going by `mode`.

LebedevRI:
```c++
uint32_t bitPerPixel = raw->getEntry(TiffTag::BITSPERSAMPLE)->getU32();
if(bitPerPixel != 12 && bitPerPixel != 14)
      ThrowRDE("Bad bit per pixel: ", bitPerPixel);
unsigned shamt = 14-bitPerPixel;
for(auto&b : mRaw->blackLevelSeparate)
  b >>= shamt;
mRaw->whiteLevel >>= shamt;
```

LebedevRI:
Looking at the results, not really sure how much this helps anything, but it's certainly a step in the right direction.
Thank you!

kmilos:
Thanks. Indeed, in limited testing I haven't yet come across a sample where the values are different from the ones in cameras.xml...

PR #394, opened 2022-11-03
CI: update actions, streamline MSYS2 setup

kmilos:
[Security related changes:](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates)

* [GitHub actions are updated to their latest available version](https://github.com/darktable-org/rawspeed/network/dependencies)
  * w/ the exception of Codecov action (as breaking changes were mentioned), leaving that to someone more clued up...

Efficiency/environmental impact related changes:

* [In-progress CI jobs are cancelled if a new commit is pushed to an existing PR](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-to-cancel-any-in-progress-job-or-run)

* MSYS2 setup now installs the minimal set of needed packages instead of entire toolchain
  * also, [`pacboy` is used for arch/msystem independent package specification for install](https://github.com/msys2/setup-msys2#pacboy)

  Before : `Total Installed Size:  1603.82 MiB`

  After: `Total Installed Size:  116.40 MiB` + `Total Installed Size:  1141.37 MiB`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -141,7 +141,7 @@ jobs:
         run: |
           set -xe
           git config --global --add safe.directory "$GITHUB_WORKSPACE"
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         name: Fetch/Checkout RawSpeed git repo
         with:
           path: 'rawspeed'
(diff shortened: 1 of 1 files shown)


LebedevRI:
This could use some more words in the description part i think.

kmilos:
> This could use some more words in the description part i think.

Done.

LebedevRI:
> * w/ the exception of Codecov action (as breaking changes were mentioned), leaving that to someone more clued up...

Where? https://github.com/codecov/codecov-action only mentions breaking changes when going from v1.

kmilos:
> https://github.com/codecov/codecov-action only mentions breaking changes when going from v1.

True. Still,

> leaving that to someone more clued up...

LebedevRI:
I'm not really sure about `concurrency` thing, the rest seems fine i guess.

LebedevRI:
Thanks!

kmilos:
> I'm not really sure about `concurrency` thing

I added the same to darktable and exiv2, works as advertised.

LebedevRI:
E.g. i don't see why non-PR builds should be aborted.

kmilos:
Yeah, not quite sure how to separate the "on push" and "on PR" references, but haven't seen any side effects yet with this setup...

PR #395, opened 2022-11-08
Samsung NX mini support

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/12820

~~Note: I have not verified the crop.~~


kmilos:
Dropping in favor of #396

PR #396, opened 2022-11-08
Support Samsung NX mini mirrorless camera from 2014

ge0rg:
This is based on a conversion of an SRW (Samsung RAW file) to DNG and the output of `dngmeta.sh` in addition to manually inspecting the crop area and black areas.

The sensor range has been confirmed over all ISO settings from 160 to 12800.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -12108,6 +12108,28 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="SAMSUNG" model="NX mini">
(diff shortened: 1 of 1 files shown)


LebedevRI:
@ge0rg thank you, but! Please contribute proper sample to RPU.
The existing sample is almost unusable. I can not tell what the right crop is, even after fixing it.

ge0rg:
Whoops, I need to look into my github mailbox more often. I've just uploaded a more scenic NX mini photo to RPU. Please let me know if you need different ISO, lighting conditions or a better view ;-)

PR #397, opened 2022-11-08
CI: improved PR job cancelling

kmilos:
This time, only works on pull_request events as per https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [ develop ]
 
+concurrency: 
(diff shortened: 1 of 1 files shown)


LebedevRI:
If there are many jobs within a workflow, they do not cancel each other, right?

kmilos:
All the jobs of a workflow get cancelled together if they have the same reference - `head_ref` i.e. source branch for a pull request; for push events, nothing gets cancelled as `run_id` is unique.

LebedevRI:
Thanks!

PR #404, opened 2022-11-21
Sony A74 (ILCE-7M4) support

nugecom:
Hi,

I use darktable, which uses rawspeed, and am having problems with:

- Blown highlights showing as lilac colour (I have a work around for this)
- Lossless compressed raw files (.arw) not being supported

I raised a case with darktable here:
https://github.com/darktable-org/darktable/issues/12508

... where they mentioned their reliance on rawspeed for file support.

Does rawspeed support Sony A74 (ILCE-7M4)? On your website it is not listed but the above darktable pull request has some code showing rawspeed does support my camera.

Is the lack of support for lossless compressed a rawspeed issue or a darktable issue?

If rawspeed, is there a timescale for addressing this / these issues?

Any help appreciated. Thanks!

John


LebedevRI:
Why is this a PR?

PR #406, opened 2022-11-22
 Fujifilm X-T5 support

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/12952

Tested the X-T5 RPU samples w/ ADC 15.0, and indeed the values are the same as the X-H2.

I've used the same crop as X-H2 as well.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -15755,6 +15755,46 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="FUJIFILM" model="X-T5">
+		<ID make="Fujifilm" model="X-T5">Fujifilm X-T5</ID>
+		<CFA2 width="6" height="6">
+			<ColorRow y="0">GGRGGB</ColorRow>
+			<ColorRow y="1">GGBGGR</ColorRow>
+			<ColorRow y="2">BRGRBG</ColorRow>
+			<ColorRow y="3">GGBGGR</ColorRow>
+			<ColorRow y="4">GGRGGB</ColorRow>
+			<ColorRow y="5">RBGBRG</ColorRow>
+		</CFA2>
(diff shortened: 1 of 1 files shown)


kmilos:
@LebedevRI This, and the OM System OM-5 (depending on the larger OM-1 patch), are the last relevant cameras [supported by current ADC 15.0](https://helpx.adobe.com/camera-raw/kb/camera-raw-plug-supported-cameras.html) (likely the last release before dt 4.2 in December), so it would be great if we could get them in please?

LebedevRI:
This i will get merged.
The problem with merging larger code-affecting parts, is that they cause chain reaction
of having to deal with bugs in them, which prevents propagating that new code to `stable`.
So if i do merge that now, it won't make into next darktable release.

kmilos:
> This i will get merged.

Great, thanks.

> The problem with merging larger code-affecting parts, is that they cause chain reaction
of having to deal with bugs in them, which prevents propagating that new code to stable.

Fair enough. Maybe it works out for the spring update ;)

LebedevRI:
> > The problem with merging larger code-affecting parts, is that they cause chain reaction
> > of having to deal with bugs in them, which prevents propagating that new code to stable.
> 
> Fair enough. Maybe it works out for the spring update ;)

FWIW, i agree that this is not great, i just don't really see a way around that.
Either i deal with things as they come in, or they will just pile up the backlog
of things to be cleaned up sometimes later, and that almost never happens.

kmilos:
> Either i deal with things as they come in

Speaking of which, can you at least please release the OM Digital Solutions OM-5 samples on RPU (the related PR claims they have been uploaded)?

As for the rest, I only wish I have time to invest into honing my C++ skills, so my contributions must remain limited, if not distracting 🤷

LebedevRI:
Thanks!
Looks like here the crop was correct.

PR #408, opened 2022-11-28
Misc CI changes

LebedevRI:


.github/workflows/CI.yml
```diff
@@ -329,62 +84,16 @@ jobs:
             msys2: { msystem: MINGW64, arch: x86_64 }
```
kmilos: FYI, MINGW64 is being sort of deprecated, UCRT64 is now the recommended MSYS2 default: https://www.msys2.org/news/#2022-10-29-changing-the-default-environment-from-mingw64-to-ucrt64

LebedevRI: Good to know, but i'm not sure what to do about it, just dropping it in favor of the latter seems lossy?

Issue #409, opened 2022-12-10
Sony A7RM5 (ILCE-7RM5) camera support


ghost:
Hello

Would it be possible to include raw support for the Sony A7RM5 (ILCE-7RM5) please?

Thank you very much!

kmilos:
Duplicate of https://github.com/darktable-org/darktable/issues/13035 and https://github.com/darktable-org/darktable/issues/12985

da-phil:
Rawspeed already supports uncompressed & lossy compressed images, support for lossless compression (L) is prepared in the following PR: https://github.com/darktable-org/rawspeed/pull/483
Lossless compression M/S is going to be future work I guess.

LebedevRI:
c38746dae23d6f8fca594be50678bc55607da268

PR #410, opened 2022-12-12
Sony ILCE-7RM5 support

kmilos:
Using the just released ADC 15.1 for the color matrix.

Partially addresses https://github.com/darktable-org/darktable/issues/13035 (and https://github.com/darktable-org/darktable/issues/12985, and #409), i.e. no lossless mode support included yet.

As the RPU sample has uniform background, the crop might not be 100% correct and needs to be verified (currently copied from 7RM4).

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -13702,6 +13702,24 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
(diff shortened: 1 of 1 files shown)


LebedevRI:
I was waiting for this patch :)
Hopefully still not too late.

Issue #246, opened 2020-10-27
Panasonic DC-S5 96MP raws are version 7 - new decompressor needed


LebedevRI:
```
[rawspeed] (dc-s5_hr_96MP.RW2) virtual rawspeed::RawImage rawspeed::Rw2Decoder::decodeRawInternal(), line 151: Version 7 is unsupported
```

_Originally posted by @LebedevRI in https://github.com/darktable-org/rawspeed/pull/242#issuecomment-717512473_

ntsarb:
A quick note to add that Darktable currently crashes when attempting to open a 96MP raw image file from Panasonic DC-S5.

Also, when I enable the option in the camera to produce a standard resolution file alongside the 96MP one, this (smaller) file will also cause Darktable to crash upon opening.

Is there any way I can support the development of a fix, e.g. by providing sample image files?

LebedevRI:
That was addressed by https://github.com/darktable-org/rawspeed/pull/412.

Issue #367, opened 2022-06-01
Panasonic DC-GH6 raws are version 8 - new decompressor needed


LebedevRI:

uvolfen:
I uploaded a full set (all variants of crops) on raw.pixls.us with Creative Commons Zero (CC0) license at August, 22nd.

Uwe

Opendreas:
Any progress on adding GH6 to Darktable?

yaiol:
Unfortunately last version of darktable 4.2.0 didn't add support to GH6, same with GH5M2 though files are available for both cameras on raw.pixls.us. 

Only way to use darktable with those cameras is to convert RAW pictures to DNG

kmilos:
Tracker for other models depending on v8 decoder:

DC-S5M2: https://github.com/darktable-org/darktable/issues/13511
DC-S5M2X: https://github.com/darktable-org/darktable/issues/15446
DC-G9M2: https://github.com/darktable-org/darktable/issues/15679
DC-S9: https://github.com/darktable-org/darktable/issues/18609
DC-GH7
DC-S1RM2

bringlein:
Any update on this?
Does it depend on `libraw` releasing the GH6 support as open source? (see [here](https://www.libraw.org/node/2710) and [here](https://www.libraw.org/comment/6626#comment-6626))? 
Thanks!

kmilos:
> Does it depend on `libraw` releasing the GH6 support as open source?

It depends on _someone_ reverse engineering the new codec and releasing the spec or code as open source, it doesn't necessarily have to come from LibRaw.

kmilos:
Just landed for porting: https://github.com/LibRaw/LibRaw/blob/master/src/decoders/pana8.cpp @jdneumeyer77 @mlouielu

jdneumeyer77:
That code is pretty ugly. Not sure when I'll get a chance to work on it, but hopefully over the next couple weeks I'll take crack at an initial port.

LebedevRI:
> That code is pretty ugly.

:)

> Not sure when I'll get a chance to work on it, but hopefully over the next couple weeks I'll take crack at an initial port.

If you do want to take a look at this, i would strongly suggest the PR should be *just* the code being copied over,
with no rewrites silencing compiler/clang-tidy diagnostics, or using rawspeed's native data structures.
That'll give us a known-"good" baseline change.

LibRaw:
[...]

PR #412, opened 2022-12-14
Support Panasonic RW2 v7 formats

jdneumeyer77:
Support 12 and 14bit V7 RW2 raw format.

I tested various files from my Panasonic S5 in various modes that produce this format:
- Hi-res/Pixel-shift
- Multiple exposure mode (like double exposures from film days)
- Live Composite mode

I ported some code from Libraw. Plus, I cleaned it up a bit to make the bit manipulation easier to understand and came up with an easy way to support both 12 bit and 14 bit. My camera produces only 14 bit. Even Libraw mentions having never seen 12 bit in the wild.

This should address: darktable-org/darktable#8338 darktable-org/darktable#9768 #246 . It might fix others as well.

I believe there are already samples available. I mimicked the existing fuzzers, but I'm not sure how they work. Nor am I sure of other automated tests that could be enabled. Happy to add those.


LibRaw:
You ported our (LibRaw) code, please don't forget to mention it in copyright section

kmilos:
> I think one of the GH5S or something required v7 as well.

GH5M2 (https://github.com/darktable-org/darktable/issues/9370) is actually **v6**, but w/ just 14b mode, while only 12b is currently implemented by rawspeed. GH5S is a **v5** where dual 12b/14b is already supported.

jdneumeyer77:
@LibRaw No worries. I can't take credit for reverse engineering that decodes the block. Feel free to use the readability improvements.

@kmilos You're right about the GH5M2, but I think it needs 12 bit support not 14 bit. I was thinking about tackling that issue as well since I have some extra free time. I don't own that camera, but I can test some samples and make sure my S5's files don't break. I thought there was a G* series camera that used the v7, but I think I'm mistaken. GH6 uses a new v8 format.

LibRaw:
@jdneumeyer77  This format is so simple that no reverse engineering was needed, just accurate look into RAW data.
Thank you for adding our copyright.

LebedevRI:
> You ported our (LibRaw) code, please don't forget to mention it in copyright section

Does LGPL really require the `Copyright (C)` line to be present, not just the license as a whole?
Quote please?

LibRaw:
I have not asked about this specific form, just about 'mentioning'
It is up to original code author how to mention LibRaw. If @jdneumeyer77  preferred to put (C) then we agree

jdneumeyer77:
I was just following existing conventions. v6, which looks to be ported as well, had the `Copyright (C) LibRaw LLC (info@libraw.org)` already.

Good to know that's a simple format. I'm not familiar bayer sensor data. So I wouldn't know. I don't typically work with binary formats nor C++. I'll have to dig into it sometime to better understand what it is doing.

Edit: Ok I think I understand it clearer. now. Sensor data before this step is 14/12 bit pixels written out in big endian. It's read as byte chunks. In order to convert it the next 14/12 bits read and converted into little endian which is stored it in the next largest type size (2 bytes ushort).

kmilos:
> You're right about the GH5M2, but I think it needs 12 bit support not 14 bit.

It's quite possible that I mixed them up. In any case, one of the two is missing for V6.

LebedevRI:
@jdneumeyer77 so i'm finally looking into this, and i'm a bit troubled
by the amount of samples on RPU. Is this really only a single camera?

More importantly, in the original comment you wrote:


> I tested various files from my Panasonic S5 in various modes that produce this format:
> 
> * Hi-res/Pixel-shift
> * Multiple exposure mode (like double exposures from film days)
> * Live Composite mode

... but there is only a single (High resolution (sensor shift) mode 96 MP) affected sample.
Could you at least please contribute one more affected sample, one with minimal image size?

jdneumeyer77:
@LebedevRI this is for several cameras. I believe S1 and S1H have the same features and use this format. The S1R uses a different sensor, but I think it uses this format as well for hi-res/live composite/etc.. I'm not sure if Gx series series uses this format for building images like hi-res mode. 

I'll upload samples for live composite and multiple-exposure from my camera. I'll check if I can open hi-res samples from the S1/S1H/S1R to verify those as well. It's been working great for my camera the last couple weeks.

LebedevRI:
> I'll upload samples for live composite and multiple-exposure from my camera.

At a bare minimum, that is indeed what i would like to see (on RPU) before merging this.
Mainly, i just want to have some smallest image with such format,
so i can tag it to be part of masterset, so it participates in perf testing etc.

It would be really good to have appropriate samples from other affected cameras.
Especially, if it suddenly turns out that one of them writes 12-bit raws...

jdneumeyer77:
@LebedevRI  Ok. I've added two quick samples for live composite and multiple-exposure modes. They aren't much larger than normal raw files.

This [issue](https://github.com/darktable-org/darktable/issues/8338) indicates someone uploaded a high res sample for S1R. The sample doesn't appear on RPU. I did manually check if it's v7 format and it is with a sample I found elsewhere. It works locally. Same with a hi-res sample of the S1. No samples available, but I found a sample raw elsewhere. It also works.

> It would be really good to have appropriate samples from other affected cameras. Especially, if it suddenly turns out that one of them writes 12-bit raws...

This PR includes support for both 12 bit and 14 bit. A comment from Libraw source code says they've not encountered a 12 bit file in this format yet.

LebedevRI:
> @LebedevRI Ok. I've added two quick samples for live composite and multiple-exposure modes. They aren't much larger than normal raw files.

Thanks you, that'll do.

> > It would be really good to have appropriate samples from other affected cameras. Especially, if it suddenly turns out that one of them writes 12-bit raws...
> 
> This PR includes support for both 12 bit and 14 bit. A comment from Libraw source code says they've not encountered a 12 bit file in this format yet.

My point is that it's dead code if we don't have any evidence any such raws were ever produced.
So i'm just doing to rip it out. Which is why i'm asking to double check that other cameras don't produce them.

pgassmann:
> It would be really good to have appropriate samples from other affected cameras. Especially, if it suddenly turns out that one of them writes 12-bit raws...

I'm waiting for GH5M2 support which uses [12bit according to some comments](https://github.com/darktable-org/rawspeed/pull/412#issuecomment-1353415227). 

Not sure if this is in the scope or where GH5M2 support is tracked. There's only one other issue here mentioning this camera: #367 and one for darktable: https://github.com/darktable-org/darktable/issues/9370

@LebedevRI can you look at this too or does it make sense to create a separate issue?

LebedevRI:
As #367 mentions, that camera produces yet another format, v8, this adds support for v7.

kmilos:
> As #367 mentions, that camera produces yet another format, v8, this adds support for v7.

[GH5M2 is v6](https://github.com/darktable-org/darktable/issues/9370#issuecomment-996021411), GH6 is v8...

jdneumeyer77:
@pgassmann GH5M2 uses v6 with 12 bit. I don't own the camera, but once I get time I can port that code over as well in a separate PR.

@LebedevRI I'm all for removing dead code, but it seems like Panasonic added 12 bit later for v6. Not sure if it was a similar case where there weren't any cameras that produced v6 12 bit at the time, so the support was excluded/ripped out.

Also I tried a few other tests such as using electronic shutter (ec) or ec + noise reduction with hi-res mode to see if they produced 12 bit. I know some manufactures bin to 12 bit using electronic shutter mode. Still produced 14 bit raws. So at least with my camera I can't produce 12 bit. I'll ask a friend to test and provide examples from S1R in different modes like multiple-exposure and using ec.

LebedevRI:
Thanks!
Here it will be trivial to add should the need emerge.

LebedevRI:
@jdneumeyer77 thank you! This should not have taken this long.

pgassmann:
> @pgassmann GH5M2 uses v6 with 12 bit. I don't own the camera, but once I get time I can port that code over as well in a separate PR.

Thank you. How much time does it take to port that code over? Is it worth a speparate issue to track this in rawspeed or is https://github.com/darktable-org/darktable/issues/9370 enough? That one is currently assigned to @LebedevRI

PR #413, opened 2022-12-16
Nikon Z 30 support

kmilos:
From https://github.com/darktable-org/darktable/issues/12239

Forgot about it, too bad it slipped through the cracks for dt 4.2 🤦

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -4406,6 +4406,42 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="NIKON CORPORATION" model="NIKON Z 30" mode="12bit-compressed">
+		<ID make="Nikon" model="Z 30">Nikon Z 30</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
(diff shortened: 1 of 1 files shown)


LebedevRI:
> Forgot about it, too bad it slipped through the cracks for dt 4.2

Hm, why is there a tag already? I was told 22'th.

LebedevRI:
@kmilos thank you!
If you want some pointers, things like [this](https://github.com/darktable-org/rawspeed/blob/develop/src/librawspeed/decoders/ArwDecoder.cpp#L70-L73) are prime candidates
for refactoring -- there should be some Entries in that TIFF that provide
the same data, we shouldn't hardcode it.
Same for anything that is driven by `hints.has("<???>")`.

kmilos:
> Hm, why is there a tag already? I was told 22'th.

Yeah, that's including the binary packaging for Windows and macOS, so AFAIK there's always a couple of days of buffer to allow those volunteers to do their work and some last minute testing... Pascal usually sends an email just before the freeze...

kmilos:
Can't promise, but I might look into those pointers if I find the time. Thanks!

PR #419, opened 2023-01-18
Fix format string type mismatch

kmilos:


LebedevRI:
The sign is the problem, i guess?

kmilos:
> The sign is the problem, i guess?

Yes, this just avoids warnings by cppcheck and potential build failures if setting -Werror=all

LebedevRI:
Thanks.

Issue #354, opened 2022-05-08
GFX 100S 16 bits white point


jas01:
When importing a 16 bits RAW uncompress file from the GFX100S into darktable, they are a problem with the white level. Every image are way too bright.

To fix this problem someone give me a solution and it's to change in the cameras.xml
```xml
<Sensor black="63" white="16383"/>
```
to
```xml
 <Sensor black="63" white="65535"/>
```
But when looking in the cameras.xml that would also affect 14bits file. 

So is they are any way to differentiate the GFX 100S 16 bits with the 14 bits ?, so darktable pickup the right white point

kmilos:
See also https://github.com/darktable-org/rawspeed/issues/228

dllu:
I can confirm that this fix is needed even after checking out the latest `develop` branch of `rawspeed` and compiling Darktable.

LebedevRI:
I believe this was fixed by #421.

PR #421, opened 2023-01-30
Fix GFX100(S) 16-bit sensor levels

twbaumann:
This branch corrects the black and white sensor levels for Fujifilm GFX100(S) cameras in 16-bit mode, fixing the following issues:

- [GFX 100S 16 bits white point #354](https://github.com/darktable-org/rawspeed/issues/354)
- [Fujifilm GFX 100 support issues #11071](https://github.com/darktable-org/darktable/issues/11071) (partly)

Currently, cameras.xml lists 62 or 63 for the black and 16383 for the white sensor level of these cameras, regardless of output bit depth. These numbers are correct for 14-bit output. For 16-bit they should be 255 (GFX100S) or 256 (GFX 100) for black and 65535 for white, as found in the exif tags _Exif.Fujifilm.BlackLevel_, _Exif.SubImage1.BlackLevel_ and _Exif.SubImage1.WhiteLevel_, the last two extracted from DNG converted files. Darktable uses these values per default which leads to apparently overexposed raws, unless you change the white level manually.

New modes '16bit-uncompressed' and '16bit-compressed' with adapted sensor levels are added to cameras.xml (and cameras.xsd) for Fujifilm GFX 100 and GFX100S cameras. The only code change is within _decodeMetaDataInternal_ of RafDecoder.cpp. I did not change _checkSupportInternal_ and _decodeRawInternal_, which should be fine as long as the standard modes '' and 'compressed' exist.

The change of black and white levels does not affect existing edits in darktable, but only new image imports and history resets.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -14889,6 +14889,42 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="FUJIFILM" model="GFX 100" mode="16bit-uncompressed">
+		<ID make="Fujifilm" model="GFX 100">Fujifilm GFX 100</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
+		</CFA>
+		<Crop x="0" y="4" width="-146" height="-6"/>
+		<Sensor black="256" white="65535"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">16212 -8423 -1583</ColorMatrixRow>
+				<ColorMatrixRow plane="1">-4336 12583 1937</ColorMatrixRow>
+				<ColorMatrixRow plane="2">-195 726 6199</ColorMatrixRow>
+			</ColorMatrix>
+		</ColorMatrices>
+	</Camera>
+	<Camera make="FUJIFILM" model="GFX 100" mode="16bit-compressed">
+		<ID make="Fujifilm" model="GFX 100">Fujifilm GFX 100</ID>
+		<CFA width="2" height="2">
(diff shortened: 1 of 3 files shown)


LebedevRI:
Is that data really not present in the files themselves?
In general, we should be removing such things from `cameras.xml`, not adding more of it...

LebedevRI:
Failures are unrelated.
Github is starting to do the google thing of breaking everyone and pretend everything is fine.
https://github.com/orgs/community/discussions/45830 https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/

twbaumann:
> Is that data really not present in the files themselves? In general, we should be removing such things from `cameras.xml`, not adding more of it...

Well, the black point is available as Exif.Fujifilm.BlackLevel, the white point could be set to 2 ^ bit_depth - 1. I would then either override the cameras.xml sensor levels when the bit depth is 16 (safe, as it only affects GFX100(S)) or perhaps I could remove the sensor data for these cameras and use that as trigger for taking the exif data.

LebedevRI:
> > Is that data really not present in the files themselves? In general, we should be removing such things from `cameras.xml`, not adding more of it...
> 
> Well, the black point is available as Exif.Fujifilm.BlackLevel, the white point could be set to 2 ^ bit_depth - 1. I would then either override the cameras.xml sensor levels when the bit depth is 16 (safe, as it only affects GFX100(S)) or perhaps I could remove the sensor data for these cameras and use that as trigger for taking the exif data.

I think we should default to the data from EXIF (== derive white level from bit depth here),
and override it with the data from `cameras.xml`. So yes, that sounds like a better plan.
You will probably need to adjust some things to still work with missing `<Sensor>` entry.

twbaumann:
It's finished. For any Fujifilm camera to use EXIF data for black and white point just set the sensor black and white values in cameras.xml to '0', as I did for GFX 100 and GFX100S. But check the data availability first, specifically Exif.Fujifilm.BlackLevel and Exif.Fujifilm.BitsPerSample.

LebedevRI:
Thanks!

PR #425, opened 2023-02-08
Assert on out of bounds access

kmilos:
From cppcheck 2.10:

[rawspeed/src/librawspeed/decompressors/PentaxDecompressor.cpp:121] (error,inconclusive) Out of bounds access in 'v2[sm_num]', if 'v2' size is 16 and 'sm_num' is 255 [containerOutOfBounds]

### DIFF
--- a/src/librawspeed/decompressors/PentaxDecompressor.cpp
+++ b/src/librawspeed/decompressors/PentaxDecompressor.cpp
@@ -117,6 +117,7 @@ HuffmanTable<> PentaxDecompressor::SetupHuffmanTable_Modern(ByteStream stream) {
         sm_val = v2[j];
       }
     }
+    assert(sm_num < 16);
     codeValues.push_back(sm_num);
     v2[sm_num] = 0xffffffff;
   }


LebedevRI:
I'd advise against using cppcheck nowadays.

That situation it suggests is impossible, and if it was possible,
we certainly wouldn't want to hide it (since originally we'd get
a sanitizer diagnostic/crash instead).

If you really care, just add an assert that `sm_num` is in-range.

kmilos:
> If you really care, just add an assert that `sm_num` is in-range.

Done. Although you're right, it cannot happen, and in that sense, it doesn't matter what `sm_num` is initialized to, so one could just as well make the cppcheck warning go away... (or we simply drop the whole thing)

LebedevRI:
The more general/long-term thing will be enabling,
or at least fixing instances highlighted by,
clang's `-Wunsafe-buffer-usage` diagnostic,
by using `Array1DRef` instead of directly indexing `array`s.

LebedevRI:
@kmilos thanks.

PR #426, opened 2023-02-09
Add Canon PowerShot SX160 IS

kmilos:


LebedevRI:
No idea why we have these, but consistent with existing practice.

LebedevRI:
@kmilos thanks.

kmilos:
It's just for sanitizing UI strings, i.e. avoiding "Canon Canon ..." in this case.

PR #432, opened 2023-02-17
Add Phase One P45 and IQ180 - white levels ?

twbaumann:
I analyzed IIQ L raw files of two Phase One digital backs, P45 and IQ180, which I also converted to DNG format to extract Exif.SubImage1.BlackLevel, Exif.SubImage1.WhiteLevel and Exif.Image.ColorMatrix2 (for D65 illuminant). For the crop margins I compared exports from Capture One (original raw) and darktable (DNG file) to the uncropped versions and chose the narrowest borders maintaining clean data.

Surprisingly I had to set the white level to 16383 for both P45 and IQ180 raw files to get correct image preview, histogram and highlight reconstruction, while the corresponding DNG converted versions use a white level of 65535, in line with RawDigger by LibRaw. For consistency I also divided the DNG derived black levels by 4. I don't know the reason for this 14/16-bit discrepancy and hope it does not adversely affect the quadrant correction in IiqDecoder::CorrectQuadrantMultipliersCombined, the spline curve of which extends from 0 to 65535.

For both digital backs I uploaded a full set of sample images (IIQ L, IIQ S, Sensor+ for IQ180) to raw.pixls.us.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -16502,6 +16502,24 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="Phase One A/S" model="P45">
+		<ID make="Phase One" model="P45">Phase One P45</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">GREEN</Color>
+			<Color x="1" y="0">RED</Color>
+			<Color x="0" y="1">BLUE</Color>
+			<Color x="1" y="1">GREEN</Color>
+		</CFA>
+		<Crop x="44" y="26" width="-24" height="-32"/>
+		<Sensor black="0" white="16383"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">5053 -24 -117</ColorMatrixRow>
+				<ColorMatrixRow plane="1">-5685 14077 1703</ColorMatrixRow>
+				<ColorMatrixRow plane="2">-2619 4491 5850</ColorMatrixRow>
+			</ColorMatrix>
+		</ColorMatrices>
+	</Camera>
 	<Camera make="Phase One A/S" model="P65+">
 		<ID make="Phase One" model="P65+">Phase One P65+</ID>
 		<CFA width="2" height="2">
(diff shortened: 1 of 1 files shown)


LebedevRI:
> Surprisingly I had to set the white level to 16383 for both P45 and IQ180 raw files to get correct image preview, histogram and highlight reconstruction

Yeah, i remember seeing that in some other IIQ samples i looked at. I don't know if there is a decoding issue or not.

LebedevRI:
@twbaumann thank you for contributing the sample sets!

LebedevRI:
@twbaumann thank you!

LebedevRI:
Thinking about it, given that "we" rescale the data so black level becomes `0.0f`,
and white level becomes `1.0f` (and anything above that considered clipped),
maybe we need to have additional "exposure bias" to be applied after normalization.
With that, while level would be 65535, and bias +2EV.

twbaumann:
Yes, but for highlight reconstruction based on clipped raw data you then have to set the clipping threshold to 0.25 or less which does not feel right.

dcraw and LibRaw both shift the decompressed IIQ raw data into 16-bit range, I found the `<< 2` operation in the sources. The question remains if the quadrant correction expects 16-bit data. I will try to find out ...

LebedevRI:
> Yes, but for highlight reconstruction based on clipped raw data you then have to set the clipping threshold to 0.25 or less

Why?

twbaumann:
The raw data clip at 16383, but when the white level is set to 65535 the highlight reconstruction will not normally detect clipping. In darktable exposure is done after highlight reconstruction. - Unless by exposure bias you mean shifting the raw data << 2, as done by LibRaw and Adobe.

LebedevRI:
I'm asking because the sensor, if grossly oversimplified, is an array of solar panels,
each one filling their own "capacitor" during the exposure time, and then
the full-ness of the "capacitor" is measured, let's say 1.1V..3.3V, and digitized.
Obviously, "capacitor" can overfill (hard clipping), and in reality it starts resisting
before that (non-linearity area above white point but below hard clipping).
But what matters is, the ADC bit depth is effectively epsilon,
the "width" of an analog step between two consecutive digital numbers.

My actual point is, if the 14bit ADC produces 0...16383 for 1.1V..3.3V,
does that 16bit ADC really still produce 0...16383 for 1.1V..3.3V,
and 16383..65535 is for some higher voltages (3.3V..9.9V)
that aren't even produced by the "capacitor", and are beyond it's clipping point,
instead of producing 0...65535 for 1.1V..3.3V, the normal range,
thus producing more precise measurements?

Let me know if if that does not make sense.

kmilos:
Without knowing the details of this particular sensor, there are several schemes to "extend" the the photodiode well capacity and its 14-bit ADC:

- Dual conversion gain, [found e.g. in the recent Panasonic GH6.](https://www.dpreview.com/articles/1570070253/what-is-dual-gain-and-how-does-it-work)

- Split-pixel (large diode + small diode)

The two techniques can even be combined, [as mentioned e.g. here.](https://markets.vision/new-pixel-architectures-for-high-dynamic-range-imaging/)

Don't know if and what is used by Phase One sensors, and if it really outputs usable 16 bits.

twbaumann:
To my knowledge, all currently supported Phase One backs only output 14-bit raw data, irrespective of the internal ADC capabilities. Allegedly anything more would be noise and not worth the additional file size. All you can do is append two zero bits. That changes with IQ3 and IQ4 backs which optionally offer real 16-bit output (Format IIQ_L16).

PR #436, opened 2023-02-20
Add Canon PowerShot SX220 HS and SX530 HS

kmilos:
Addresses the DNG mode only of https://github.com/darktable-org/darktable/issues/2802 and https://github.com/darktable-org/darktable/issues/10121


kmilos:
Also added all other PowerShot SX models for which there are DNG samples for on RPU so their model strings get sanitized.

LebedevRI:
@kmilos thank you.

PR #439, opened 2023-03-08
Fujifilm FinePix F550EXR support

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/8619

Essentially a copy of F600EXR.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -14227,6 +14227,24 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="FUJIFILM" model="FinePix F550EXR">
+		<ID make="Fujifilm" model="FinePix F550EXR">Fujifilm FinePix F550EXR</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Thank you.

As usual, the crop is slightly wrong, it may be good to pay a bit more attention to that
(use 800% zoom, and move the image without letting it finish processing,
you'll notice when the border pixels are wrong-looking.) (when doing so, it's best to
disable most of the modules, e.g. filmic seems to produce pixellated garbage
on the overexposed pixels that are directly on the border of an image.)

kmilos:
Yeah, it did look pretty tight at 400%...

I went finally with a simple F600EXR copy as [it really is almost the same camera (they were released a month apart).](https://www.photographyblog.com/reviews/fujifilm_finepix_f600exr_review#conclusion) Otherwise, I guess both would need to be updated...

LebedevRI:
That may simply mean that F600EXR's crop is also wrong.

LebedevRI:
Note the

> use 800% zoom, and **move the image without letting it finish processing**,

bit. Wrong crop is not as apparent without using that second part of the trick,
but it is very jarring on an pixellated "preview" image.

kmilos:
Thanks, good trick to keep in mind for the future.

PR #444, opened 2023-03-17
Add chroma and luma calibration for Phase One

twbaumann:
With some of my IQ180 images I saw clear quarter paneling artefacts in darktable, which did not show up after Adobe DNG conversion, but also not using RawTherapee and of course C1. After comparing the correction meta data present in the various Phase One raw samples with the applied corrections in darktable, dcraw and LibRaw, I decided to adapt phase_one_flat_field() for rawspeed, following the LibRaw source. The corresponding tags are 0x40b for chroma calibration and 0x410 for luma calibration, both present in all supported backs. 

The result makes me happy, see the before/after images of my specially designed step gradient test print. I can upload the raw file if that is helpful. The code changes were tested using darktable release-4.2.1 (rawspeed @ 5f991a0).

The red highlights have nothing to do with darktable, they also appear in C1. The image is correctly ETTR exposed but apparently the CCD sensor has a non-linearity in the highlights at ISO speeds below 200 - here the green raw channels are highest, but not clipping, so the highlights turn magenta.

_Before_
230307_9577_rel

_After_
230307_9577_dev

### DIFF
--- a/src/librawspeed/decoders/IiqDecoder.cpp
+++ b/src/librawspeed/decoders/IiqDecoder.cpp
@@ -143,6 +143,8 @@ std::optional<IIQFormat> getAsIIQFormat(uint32_t v) {
 }
 } // namespace
 
+enum class IiqDecoder::IiqCorr { LUMA, CHROMA };
+
 RawImage IiqDecoder::decodeRawInternal() {
   const Buffer buf(mFile.getSubView(8));
   const DataBuffer db(buf, Endianness::little);
@@ -292,6 +294,12 @@ void IiqDecoder::CorrectPhaseOneC(ByteStream meta_data, uint32_t split_row,
       correctSensorDefects(meta_data.getSubStream(offset, len));
       SensorDefectsSeen = true;
       break;
+    case 0x40b: // Chroma calibration
+      PhaseOneFlatField(meta_data.getSubStream(offset, len), IiqCorr::CHROMA);
+      break;
+    case 0x410: // Luma calibration
+      PhaseOneFlatField(meta_data.getSubStream(offset, len), IiqCorr::LUMA);
+      break;
     case 0x431:
       if (QuadrantMultipliersSeen)
         ThrowRDE("Second quadrant multipliers entry seen. Unexpected.");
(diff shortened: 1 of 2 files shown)


LibRaw:
It would be great if you mentioned LibRaw somewhere in your patch

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -384,6 +390,81 @@ void IiqDecoder::CorrectQuadrantMultipliersCombined(ByteStream data,
...
+  std::array<float, 4> mult;
+  const Array2DRef<uint16_t> img(mRaw->getU16DataAsUncroppedArray2DRef());
+
+  for (i = 0; i < 8; i++)
+    head[i] = data.getU16();
+  if (head[2] == 0 || head[3] == 0 || head[4] == 0 || head[5] == 0)
+    return;
+  wide = head[2] / head[4] + (head[2] % head[4] != 0);
+  high = head[3] / head[5] + (head[3] % head[5] != 0);
+  mrow = (float *) calloc(nc * wide, sizeof *mrow);
```
LebedevRI:
```
std::vector<float> mrow_storage;
const Array2DRef<float> mrow = Array2DRef<float>::create(mrow_storage, wide, nc);
```

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -292,6 +292,12 @@ void IiqDecoder::CorrectPhaseOneC(ByteStream meta_data, uint32_t split_row,
       correctSensorDefects(meta_data.getSubStream(offset, len));
       SensorDefectsSeen = true;
       break;
+    case 0x40b: // Chroma calibration
+      PhaseOneFlatField(meta_data.getSubStream(offset, len), 0, 4);
```
LebedevRI: 1. `/*is_float=*/false` and so on
2. it is never float, let's drop that param (or should be a `bool`)
3. sink `nc` into `PhaseOneFlatField`, just pass a `bool` or better yet an `enum`

src/librawspeed/decoders/IiqDecoder.cpp
```diff
@@ -384,6 +390,81 @@ void IiqDecoder::CorrectQuadrantMultipliersCombined(ByteStream data,
...
+
+  for (i = 0; i < 8; i++)
+    head[i] = data.getU16();
+  if (head[2] == 0 || head[3] == 0 || head[4] == 0 || head[5] == 0)
+    return;
+  wide = head[2] / head[4] + (head[2] % head[4] != 0);
+  high = head[3] / head[5] + (head[3] % head[5] != 0);
+  mrow = (float *) calloc(nc * wide, sizeof *mrow);
+  for (y = 0; y < high; y++)
+  {
```
LebedevRI: Please clang-format-15 the changes

LebedevRI:
Thanks!
Just to be sure, this does appear in at least one sample on RPU, correct?

LebedevRI:
> It would be great if you mentioned LibRaw somewhere in your patch

Does libraw now own dcraw code? That looks like `phase_one_flat_field` to me.

LibRaw:
> Does libraw now own dcraw code? That looks like `phase_one_flat_field` to me.

LibRaw's PhaseOne flat field processing is different from dcraw's (and better)

LibRaw:
> LibRaw's PhaseOne flat field processing is different from dcraw's (and better)

Okay, for this particular patch, it seems enough to mention Dave Coffin. 
But I can recommend moving the entire linearization code from our library, it is improved relative to dcraw

LebedevRI:
> > LibRaw's PhaseOne flat field processing is different from dcraw's (and better)
> 
> Okay, for this particular patch, it seems enough to mention Dave Coffin. But I can recommend moving the entire linearization code from our library, it is improved relative to dcraw

FWIW, really not a fan of just copying code around.

LibRaw:
> FWIW, really not a fan of just copying code around.

That's all about users, not only code quality and beauty. Of course, one may completely re-write other's code, or re-create same idea from scratch. 

In any case, the task is to ensure that users can process their images (apply correction recorded in metadata in this particular case), and not software development for the sake of the process itself.

src/librawspeed/decoders/IiqDecoder.cpp
twbaumann: is_float is true in dcraw's correction tag 0x401 but I found no corresponding metadata in the supported raw files, so it's gone. Can be resurrected within PhaseOneFlatField() by a third enum value if needed.

twbaumann:
> Just to be sure, this does appear in at least one sample on RPU, correct?

Yes. The problem is only visible in homogeneous areas, but I found a sample on raw.pixls.us with a visible top/bottom segmentation that is resolved by this fix: [Phase One - IQ140 - IIQ L (4_3).TIF](https://raw.pixls.us/getfile.php/2020/nice/Phase%20One%20-%20IQ140%20-%20IIQ%20L%20(4:3).TIF)

There is a P65+ sample on RPU (_Another quadrant correction issue_) that still shows segmentation, but that is also not fixed using RawTherapee.

twbaumann:
> Okay, for this particular patch, it seems enough to mention Dave Coffin.

@LibRaw Thank you, I fixed that.

src/librawspeed/decoders/IiqDecoder.cpp
twbaumann: Implemented.

src/librawspeed/decoders/IiqDecoder.cpp
twbaumann: Fixed.

LebedevRI:
(CI still seem to be having issues with the code)

LebedevRI:
Thanks! Could you please squash the commits?

twbaumann:
> Thanks! Could you please squash the commits?

I squashed the last four into one - not sure about the (unnecessary) merge from upstream and the first (review) commit.

LebedevRI:
> > Thanks! Could you please squash the commits?
> 
> I squashed the last four into one - not sure about the (unnecessary) merge from upstream and the first (review) commit.

I did mean all, yes.

twbaumann:
> > > Thanks! Could you please squash the commits?
> > 
> > 
> > I squashed the last four into one - not sure about the (unnecessary) merge from upstream and the first (review) commit.
> 
> I did mean all, yes.

Ok, done. The failing checks seem to be unrelated to the code.

LebedevRI:
> Ok, done. The failing checks seem to be unrelated to the code.

Thanks.
That's https://github.com/codecov/feedback/issues/126, it's extremely frustrating,
but i'm not aware of any way as to how it can be solved.

LebedevRI:
Thank you!

PR #451, opened 2023-03-29
Remove SonarCloud cache and threads configuration as it is now by default

mpaladin:
No need to configure the cache and threads anymore, SonarCloud now has automatic analysis caching and threads configuration. See: https://docs.sonarcloud.io/advanced-setup/languages/c-c-objective-c/#analysis-cache

### DIFF
--- a/.github/workflows/CI-linux.yml
+++ b/.github/workflows/CI-linux.yml
@@ -410,4 +410,4 @@ jobs:
         run: |
           set -xe
           cd "${SRC_DIR}"
-          sonar-scanner --define project.settings="${SRC_DIR}/.ci/sonar-project.properties" --define sonar.host.url="${{ env.SONAR_SERVER_URL }}" --define sonar.projectKey="${{ secrets.SONAR_PROJECT_KEY }}" --define sonar.organization="${{ secrets.SONAR_ORGANIZATION }}" --define sonar.cfamily.compile-commands="${BUILD_DIR}/compile_commands.json" --define sonar.cfamily.threads="$(nproc --all)" --define sonar.cfamily.cache.enabled=false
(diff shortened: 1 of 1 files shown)


.github/workflows/CI-linux.yml
```diff
@@ -410,4 +410,4 @@ jobs:
         run: |
           set -xe
           cd "${SRC_DIR}"
-          sonar-scanner --define project.settings="${SRC_DIR}/.ci/sonar-project.properties" --define sonar.host.url="${{ env.SONAR_SERVER_URL }}" --define sonar.projectKey="${{ secrets.SONAR_PROJECT_KEY }}" --define sonar.organization="${{ secrets.SONAR_ORGANIZATION }}" --define sonar.cfamily.compile-commands="${BUILD_DIR}/compile_commands.json" --define sonar.cfamily.threads="$(nproc --all)" --define sonar.cfamily.cache.enabled=false
+          sonar-scanner --define project.settings="${SRC_DIR}/.ci/sonar-project.properties" --define sonar.host.url="${{ env.SONAR_SERVER_URL }}" --define sonar.projectKey="${{ secrets.SONAR_PROJECT_KEY }}" --define sonar.organization="${{ secrets.SONAR_ORGANIZATION }}" --define sonar.cfamily.compile-commands="${BUILD_DIR}/compile_commands.json"
```
LebedevRI, line 413: At least the `sonar.cfamily.cache.enable` change is incorrect.

LebedevRI:
Is there any particular problem you are trying to address? Knowing context might help here.

.github/workflows/CI-linux.yml
mpaladin: @LebedevRI I looked through git history and I couldn't find the reason that made you disable the cache, my guess is that you set it to `false` in the past to not get a warning in the UI.
Now there is an automatic analysis cache feature, and we recommend users to just use the default settings.

mpaladin:
> Is there any particular problem you are trying to address? Knowing context might help here.

@LebedevRI the automatic cache is going to speed up the analysis, and the default is always preferred. The less configuration, the better it is.

LebedevRI:
> > Is there any particular problem you are trying to address? Knowing context might help here.
> 
> @LebedevRI the automatic cache is going to speed up the analysis, and the default is always preferred. The less configuration, the better it is.

I'm sure that it is it's goal, but i have explicitly intentionally disabled it to avoid
any and all cache invalidation problems and make it as stateless as possible.

PR #453, opened 2023-04-01
Align Phase One sensor levels

twbaumann:
1. Set all sensor levels for Phase One backs in cameras.xml to 0 for black and 16383 for white. This affects the entries for IQ140, IQ180 and IQ250.<br/>
The **black level** for all supported Phase One backs is read from metadata. The value in cameras.xml is ignored, therefore it should be set to zero.<br/>
All currently supported Phase One backs use the 14-bit IIQ_L format. A real 16-bit raw data option (IIQ_L16) is only provided by unsupported IQ3 and IQ4 backs. Based on the unshifted 14-bit raw data the **white level** should be 16383. For verification I used the samples from RPU (for IQ140) and other sources (for IQ250).<br/>

2. Add an explanatory comment to IiqDecoder::decodeRawInternal() at the line where the black level is read from metadata. The level read is related to 16-bit data, and because rawspeed provides unshifted 14-bit raw data the level is shifted accordingly. Once IIQ_L16 is supported this shift has to be made conditional.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -16565,7 +16565,7 @@
 			<Color x="1" y="1">BLUE</Color>
 		</CFA>
 		<Crop x="42" y="8" width="-20" height="-54"/>
-		<Sensor black="0" white="65535"/>
+		<Sensor black="0" white="16383"/>
 		<ColorMatrices>
 			<ColorMatrix planes="3">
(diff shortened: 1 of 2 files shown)


LebedevRI:
Thank you!

PR #454, opened 2023-04-09
rawspeed: Sony ILME-FX30 support

helloworld1:
Add Sony ILME-FX30 Camera support. 
Resolving https://github.com/darktable-org/darktable/issues/14147

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -14132,6 +14132,24 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Thank you!
I can't really tell if the crop is right, it might need to be +2 pixels tighter.

helloworld1:
Thanks for merging. I do take other pictures that has better edges and zoomed in to 300% and counted garbage pixels as 28 on the right edge. If I notice any artifacts in the future, I will create another PR to adjust.

PR #457, opened 2023-04-14
Panasonic DMC-GF8 support

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/14196

Samples should be in the upload queue.

Really just a copy of GF7...


LebedevRI:
Thank you!

PR #461, opened 2023-04-18
Sony ILME-FX3 support

kmilos:
Waiting for samples on RPU...

[Same sensor as 7SM3 (and ZV-E1)](https://www.dpreview.com/reviews/sony-zv-e1-preview), all 3 profiles confirmed identical in ADC 15.3, so a straight copy.


LebedevRI:
@kmilos thank you!

PR #465, opened 2023-05-02
Hasselblad X1D support

kmilos:
Color matrix is verified to be the same as Mk II via ADC 15.3 - as expected, as sensor is supposedly the same.

Black and white levels are from ADC, but doesn't really matter what they are as they're read from raw file metadata anyway.

The crop has been revisited and updated for Mk II as well.

Depends on https://github.com/darktable-org/rawspeed/pull/466 for OOC 3FR uncompressed support.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -17138,6 +17138,24 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="Hasselblad" model="Hasselblad X1D">
+		<ID make="Hasselblad" model="X1D">Hasselblad 50-15-Coated5</ID>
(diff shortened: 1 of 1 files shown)


data/cameras.xml
```diff
@@ -17146,7 +17164,7 @@
 			<Color x="0" y="1">GREEN</Color>
 			<Color x="1" y="1">BLUE</Color>
 		</CFA>
-		<Crop x="48" y="104" width="-60" height="0"/>
+		<Crop x="46" y="96" width="-60" height="0"/>
```
LebedevRI, line 17167: The old crop was correct.

LebedevRI:
Thank you!

PR #463, opened 2023-04-28
Hasselblad X2D 100C support

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/14345

Stub only (using ADC 15.3), levels and crop need to be verified.

Depends on #466 for OOC 3FR uncompressed support.

PR #466, opened 2023-05-02
Add Hasselblad uncompressed 3FR support

kmilos:
As found in X1D, X1D II 50C and X2D 100C (all available on RPU).

Split from https://github.com/darktable-org/rawspeed/pull/463

Also bumps size checks to 12000 x 8842 to support both H6D-100 and X2D 100C sensors.

### DIFF
--- a/src/librawspeed/decompressors/HasselbladDecompressor.cpp
+++ b/src/librawspeed/decompressors/HasselbladDecompressor.cpp
@@ -44,7 +44,7 @@ HasselbladDecompressor::HasselbladDecompressor(const RawImage& mRaw_,
 
   // FIXME: could be wrong. max "active pixels" - "100 MP"
   if (mRaw->dim.x == 0 || mRaw->dim.y == 0 || mRaw->dim.x % 2 != 0 ||
-      mRaw->dim.x > 12000 || mRaw->dim.y > 8816) {
+      mRaw->dim.x > 12000 || mRaw->dim.y > 8842) {
     ThrowRDE("Unexpected image dimensions found: (%u; %u)", mRaw->dim.x,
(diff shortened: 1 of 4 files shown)


src/librawspeed/decompressors/HasselbladDecompressor.cpp
```diff
@@ -44,7 +44,7 @@ HasselbladDecompressor::HasselbladDecompressor(const RawImage& mRaw_,
 
   // FIXME: could be wrong. max "active pixels" - "100 MP"
   if (mRaw->dim.x == 0 || mRaw->dim.y == 0 || mRaw->dim.x % 2 != 0 ||
-      mRaw->dim.x > 12000 || mRaw->dim.y > 8816) {
+      mRaw->dim.x > 12000 || mRaw->dim.y > 8842) {
```
LebedevRI, line 47: (nitpick) Like i said, this is adding support for uncompressed, why is it touching something else?

LebedevRI:
Thank you!

src/librawspeed/decompressors/HasselbladDecompressor.cpp
LebedevRI: I ended up not merging this part.
After the main change of this PR, no RPU sample fails with this error,
which leads me to believe this was superficial.
This is why i'm asking for small separate changes.

kmilos: In theory, the tethered download from camera will produce a compressed .FFF file of this size, but indeed, there is no such sample in RPU currently. Maybe @kkempter-getswift can supply one as well please?

PR #467, opened 2023-05-02
canon EOS M6 White Level for stopped down lens

MStraeten:
found former settings are fine with open aperture (ef-m 22 f2.0), stopped down needs a different set (ef-m f5.6) did some investigations with a clipping flashlight with both f stops and all ISO values and checked that with real life images

so general question: which usecase should be covered by default?

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -1685,10 +1685,10 @@
 			<Color x="1" y="1">BLUE</Color>
 		</CFA>
 		<Crop x="266" y="42" width="-6" height="-4"/>
-		<Sensor black="2048" white="16000"/>
-		<Sensor black="512" white="16000" iso_list="100 125 200 250"/>
-		<Sensor black="512" white="13200" iso_list="160"/>
-		<Sensor black="2048" white="13200" iso_list="320 640 1250 2500 5000"/>
+		<Sensor black="2048" white="15600"/>
+		<Sensor black="512" white="15600" iso_list="100 125 200 250"/>
+		<Sensor black="512" white="12600" iso_list="160"/>
+		<Sensor black="2048" white="13200" iso_list="320 640 1250 2500 5000 "/>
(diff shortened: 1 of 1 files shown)


LebedevRI:
We should be reading it from EXIF.

MStraeten:
exif values doesn't differ related to aperture and values are even lower:
exiftool -H -G -a *.CR2 | grep -e 'Specular White Level' -e 'Black Level' -e '0x8827'

ISO | Black Level | White Level
-- | -- | --
100 | 512 | 11892
125 | 512 | 11892
160 | 512 | 10749
200 | 512 | 14338
250 | 512 | 14338
320 | 2048 | 10749
400 | 2048 | 14338
500 | 2048 | 14338
640 | 2048 | 10749
800 | 2048 | 14338
1000 | 2048 | 14338
1250 | 2048 | 10749
1600 | 2048 | 14338
2000 | 2048 | 14338
2500 | 2048 | 10749
3200 | 2048 | 14338
4000 | 2048 | 14338
5000 | 2048 | 10749
6400 | 2048 | 14338
8000 | 2048 | 14338
10000 | 2048 | 14338
12800 | 2048 | 14338
16000 | 2048 | 14338
20000 | 2048 | 14338
25600 | 2048 | 14338

MStraeten:
why other canon changes are merged? Is it worth to keep limited settings waiting for a solution that isn't available?

LebedevRI:
@MStraeten because it didn't explicitly mention anything about being incorrect in some cases,
and i didn't think of that.

I've finally got, err, sufficiently "motivated", to finally implement it: https://github.com/darktable-org/rawspeed/pull/578

MStraeten:
thanks, #578 works fine with M6

LebedevRI:
> thanks, #578 works fine with M6

Thank you for checking!

PR #468, opened 2023-05-03
Hasselblad X1D: adjust crop

kmilos:
Align w/ X1D II 50C (https://github.com/darktable-org/rawspeed/commit/b44f7d2c9e44761e1f08778c921ef9fa050c52c0) as sensor is identical.


LebedevRI:
I'm not sure this is needed, but bigger crop isn't smaller crop so i guess this is fine?
Thanks.

kmilos:
> I'm not sure this is needed

Yeah, there might be a more optimal for both, but let's keep X1D and X1D II in sync at least. Thanks.

PR #474, opened 2023-05-24
Simplify CMakeLists by using policy to honor CXX standard

kmilos:
According to https://cmake.org/cmake/help/latest/policy/CMP0067.html it is no longer required to set CXXFLAGS as well for try_compile().

### DIFF
--- a/cmake/compiler-flags.cmake
+++ b/cmake/compiler-flags.cmake
@@ -1,21 +1,4 @@
-include(CheckCXXCompilerFlag)
 include(CpuMarch)
-include(CheckCXXCompilerFlagAndEnableIt)
-
-# yes, need to keep both the CMAKE_CXX_FLAGS and CMAKE_CXX_STANDARD.
-# with just the CMAKE_CXX_STANDARD, try_compile() breaks:
-#   https://gitlab.kitware.com/cmake/cmake/issues/16456
-# with just the CMAKE_CXX_FLAGS, 'bundled' pugixml breaks tests
-#   https://github.com/darktable-org/rawspeed/issues/112#issuecomment-321517003
-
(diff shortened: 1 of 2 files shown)


LebedevRI:
Yup, makes sense.
Thank you!

cmake/compiler-flags.cmake
```diff
@@ -2,19 +2,9 @@ include(CheckCXXCompilerFlag)
...
 include(CheckCXXCompilerFlagAndEnableIt)
 
-# yes, need to keep both the CMAKE_CXX_FLAGS and CMAKE_CXX_STANDARD.
-# with just the CMAKE_CXX_STANDARD, try_compile() breaks:
-#   https://gitlab.kitware.com/cmake/cmake/issues/16456
-# with just the CMAKE_CXX_FLAGS, 'bundled' pugixml breaks tests
-#   https://github.com/darktable-org/rawspeed/issues/112#issuecomment-321517003
-
-message(STATUS "Checking for -std=c++17 support")
 CHECK_CXX_COMPILER_FLAG("-std=c++17" COMPILER_SUPPORTS_CXX17)
```
LebedevRI: I think we can drop this whole block too?

kmilos: I guess you'd still want to check if this level is supported, or you're thinking the prior min. compiler version check is already sufficient to cover this now?

LebedevRI: It's basically a TOCTOU issue.
I'm not sure what we get from that check nowadays,
since CMake (should?) handles it for us.

kmilos: Sure. I'll remove it.

PR #475, opened 2023-05-25
Do not throw if DNG crop is undefined (DJI bug)

kmilos:
The file is probably usable, so fall back to default values and log error instead.

This should address https://github.com/darktable-org/darktable/issues/11949 and https://discuss.pixls.us/t/unable-to-import-drone-hyperlapse-images/37222

Asked for samples...

### DIFF
--- a/src/librawspeed/decoders/DngDecoder.cpp
+++ b/src/librawspeed/decoders/DngDecoder.cpp
@@ -499,8 +499,17 @@ RawImage DngDecoder::decodeRawInternal() {
 
 void DngDecoder::handleMetadata(const TiffIFD* raw) {
   // Crop
-  if (const std::optional<iRectangle2D> aa = parseACTIVEAREA(raw))
-    mRaw->subFrame(*aa);
+  if (const std::optional<iRectangle2D> aa = parseACTIVEAREA(raw)) {
+    try {
+      if (!aa->hasPositiveArea())
+        ThrowRDE("No positive active area");
+      mRaw->subFrame(*aa);
+    } catch (const RawDecoderException& e) {
+      // We push back errors from the active area parser, since the image may
+      // still be usable
+      mRaw->setError(e.what());
+    }
+  }
 
   if (raw->hasEntry(TiffTag::DEFAULTCROPORIGIN) &&
       raw->hasEntry(TiffTag::DEFAULTCROPSIZE)) {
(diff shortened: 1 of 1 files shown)


LebedevRI:
Yeah i do not know about this... I guess they expect us to treat
such "rational"s w/ zero denominator by pretending that denominator is 1?
But i really don't see why that is what the TIFF spec prescribes,
and as far as i can tell neither dcraw nor libTIFF does that.

This is quite an unfortunate situation.

kmilos:
> Yeah i do not know about this... I guess they expect us to treat
> such "rational"s w/ zero denominator by pretending that denominator is 1?

I don't think we can assume that. For example, DJI really put (0/0, 0/0) in there, so for DefaultCropSize we can't and should not pretend it is (0,0).

> But i really don't see why that is what the TIFF spec prescribes,
> and as far as i can tell neither dcraw nor libTIFF does that.

The Exif (or TIFF/EP) spec mentions for some fields that 0/0 means "undefined".

> This is quite an unfortunate situation.

It is unfortunate, but I think falling back to default (entire image) is sensible in this specific case, instead of bailing out on the user. We do something similar already if the opcodes cannot be properly parsed (i.e. just ignore them, but not bail out)...

kmilos:
Yep, the Exif spec mentions in several places:

> This value is invalid when 0/0 is recorded.

kmilos:
This doesn't work yet btw, because `mRaw->dim` is (0,0) at this point?!

kmilos:
Argh, they also put ActiveArea as 0 0 0 0. Wonderful!

LebedevRI:
Why do we care to support those invalid files?

LibRaw:
To allow users to process files captured by their (invalid) equipment

LebedevRI:
Also, was the bug reported to the DJI? cc @DJI-SDK @lanyusea

kmilos:
> Also, was the bug reported to the DJI?

cc @mluckau @JooB1 @WiLars @MMCMA @da-phil @cdhodgdon @fran6t

lanyusea:
> Also, was the bug reported to the DJI? cc  @lanyusea

let me check

kmilos:
Looks like both `DefaultCropSize` and `ActiveArea` are fixed in the recently launched DJI Air 3, but I guess there's still a need to support older invalid files.

kmilos:
@LebedevRI Can we please reconsider this? Just came across another instance from the new [Adobe Indigo app](https://research.adobe.com/articles/indigo/indigo.html)...

LebedevRI:
RIIR project is entering active phase, once there's feature parity, this repo will be archived.
Until then, this repo will only be recieveing trivial new camera support.

kmilos:
Ah. Thanks for the update.

kmilos:
Just for reference, undefined `DefaultCropSize` can be found in some [Project Indigo DNGs](https://github.com/darktable-org/darktable/issues/19176) as well.

PR #479, opened 2023-05-30
Fujifilm X-S20 support

kmilos:
Used ADC 15.3.1 that introduced support, but waiting for RPU samples to verify coeffs and crop.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -15805,6 +15805,46 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="FUJIFILM" model="X-S20">
+		<ID make="Fujifilm" model="X-S20">Fujifilm X-S20</ID>
+		<CFA2 width="6" height="6">
+			<ColorRow y="0">GGRGGB</ColorRow>
+			<ColorRow y="1">GGBGGR</ColorRow>
+			<ColorRow y="2">BRGRBG</ColorRow>
+			<ColorRow y="3">GGBGGR</ColorRow>
+			<ColorRow y="4">GGRGGB</ColorRow>
+			<ColorRow y="5">RBGBRG</ColorRow>
+		</CFA2>
+		<Crop x="0" y="6" width="-90" height="-6"/>
+		<Sensor black="1022" white="16383"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">12836 -5909 -1032</ColorMatrixRow>
+				<ColorMatrixRow plane="1">-3087 11132 2236</ColorMatrixRow>
+				<ColorMatrixRow plane="2">-35 872 5330</ColorMatrixRow>
+			</ColorMatrix>
+		</ColorMatrices>
+	</Camera>
+	<Camera make="FUJIFILM" model="X-S20" mode="compressed">
+		<ID make="Fujifilm" model="X-S20">Fujifilm X-S20</ID>
+		<CFA2 width="6" height="6">
+			<ColorRow y="0">GGRGGB</ColorRow>
+			<ColorRow y="1">GGBGGR</ColorRow>
+			<ColorRow y="2">BRGRBG</ColorRow>
+			<ColorRow y="3">GGBGGR</ColorRow>
+			<ColorRow y="4">GGRGGB</ColorRow>
+			<ColorRow y="5">RBGBRG</ColorRow>
+		</CFA2>
+		<Crop x="0" y="6" width="-90" height="-6"/>
+		<Sensor black="1022" white="16383"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">12836 -5909 -1032</ColorMatrixRow>
+				<ColorMatrixRow plane="1">-3087 11132 2236</ColorMatrixRow>
+				<ColorMatrixRow plane="2">-35 872 5330</ColorMatrixRow>
+			</ColorMatrix>
+		</ColorMatrices>
+	</Camera>
 	<Camera make="FUJIFILM" model="X-T1">
 		<ID make="Fujifilm" model="X-T1">Fujifilm X-T1</ID>
 		<CFA2 width="6" height="6">


LebedevRI:
Hm, and nowadays dt's demosaic seems to always produce weird artifacts around the edges.
Will i really have to fork it.

LebedevRI:
@kmilos @frischnetz thank you!

kmilos:
Ok, so it's not just me then - I was finding your method for crop checking was getting unreliable recently... @jenshannoschwalm

jenshannoschwalm:
Being pinged here, I didn't yet see an issue with demosaicing, could you give me a hint?

Or did I misunderstand?

kmilos:
The way Roman checks the crop is this: turn on crop editing in rawprepare via darktablerc, then slide the image at 400-800% close to the edges. The hypothesis is that crop is good if there are no demosaicing artifacts (usually false color flashing when sliding for me). This kinda stopped working around the 4.4 release, there are now artefacts even with a crop heavier than theoretically needed...

kmilos:
For example, using the uncompressed X-S20 sample here and my `unprocessed_raw` dump method (from the LibRaw package), I would maintain the max valid crop is perhaps even (0,6,-84,-6)...

jenshannoschwalm:
1. Only for xtrans?
2. Do you suspect anything special already?
3. All borders affected?

kmilos:
No, not only for x-trans [1], and seems to be mostly top and right edges in my experience. No clues as to cause, sorry.

[1] E.g. I was seeing this trying to update and match the crops of Pentax K-3 Mark III, and the K-3 Mark III Monochrome (no PEF for mono on RPU, so bypassing the DNG values just to check the max possible). In theory they could be the same for color and mono, no? But just starting to change the crops of the color K-3 Mark III I couldn't get rid of the flashing for almost any value, including the original ones we have in cameras.xml... Oh, and no OpenCL here (Windows, blacklisted Intel Xe).

Here's the example of what happens e.g. if a drag at 800% along the top edge and release (color K-3 Mark III, no changes to rawspeed database crop):

image

P.S. Shall we open a separate issue on dt if confirmed?

jenshannoschwalm:
Did you try with demosaic bypass mode? No other fancy modules involved like hlr or lens? Somehow depending on zoom level? 

I guess, yes an issue...

jenshannoschwalm:
Aaah, I looked at the borders and missed the cyan errors. Clearly wrong. Should never happen and certainly not while changing the crop.

LebedevRI:
> Ok, so it's not just me then - I was finding your method for crop checking was getting unreliable recently... @jenshannoschwalm

Yeah, certainly not just you. As i have said in the now-locked thread, i'm really disappointed in the dt's trajectory.

jenshannoschwalm:
Sorry, I was not aware of such problems. My idea is, for some demosaicers we have optimized code for the inner part and the border region would be approximated by ppg. What the image shows would be exactly an issue where that outer ppg calculated border is too small.

jenshannoschwalm:
> Here's the example of what happens e.g. if a drag at 800% along the top edge and release (color K-3 Mark III, no changes to rawspeed database crop):

Just to make sure, this is the output after "settling" the fast-scaled ?

kmilos:
> Just to make sure, this is the output after "settling" the fast-scaled ?

Yes, in the case shown.

jenshannoschwalm:
Using the provided file for K3 Mk III, disabling pipecache, removed processing modules after demosaicer and no hlr algorithm i could reproduce some strange visible issue so far
1. Only on CPU
2. While resizing the main window in some cases, in some not
3. Not depending on the chosen demosaicer (could switch and observe the error in all cases)
4. Appearing not at the borders of the image but at the canvas border
5. 
Bildschirmfoto vom 2023-07-12 06-04-31
Bildschirmfoto vom 2023-07-12 06-10-48

6. Somewhat changing but practically independent of the selected settings in rawprepare crop
7. could confirm the issue is also present in passthru demosaicer modes
8. The affected locations change while changing white points in the dt rawprepare module !
9. I could not reproduce with white balance module switched off!
10. Affected locations seem to be depending on temperature setting in white balance module

When dumping the pipe intermediate results i think i can confirm the artefacts being introduced in white balance module so far. (images scaled and brightness-corrected in gimp)

0087_temperature_cpu_in_M


0088_temperature_cpu_out_M

Any ideas so far on this strange thing?

jenshannoschwalm:
Ok, i think we have a "gotcha" here.

In temperature.c i replaced the optimized loop using some __alignment magic__ with a brute-force-simple
```
      for(int i = 0; i < width; i++)
      {
        const size_t p = (size_t)j * width + i;
        out[p] = in[p] * d_coeffs[FC(j + roi_out->y, i + roi_out->x, filters)];
      }
```

@ralfbrown i think this is from https://github.com/darktable-org/darktable/commit/22d8dd9bff1a757722c3a98e7ceb9a9824f48a97

ralfbrown:
Unable to reproduce with the K3m3 "14-bit auto" PEF from RPU.
From the screenshots, my thought is that something is going awry in the initial unaligned group of max. 3 sensels.  What happens if you remove the alignment directives for `inp` and `outp` from `scaled_copy_4wide` (both preceding the function and in the for_four_channels) and eliminate the initial 0...alignment single-sensel loop, so that the Bayer branch is more like the Xtrans branch?  That would preserve the vectorization, though still be a bit slower due to unaligned reads and writes.

ETA: the corrupted pixels in that last screenshot are suspiciously regularly spaced, so maybe there's a race condition involving the OpenMP parallelization.  Does the set of corrupted pixels change if you change WB and then restore its settings?  Does the set change if you restrict the number of threads dt is using? (run with `-t N`)

jenshannoschwalm:
@kmilos, i did a lot of more testing but so far could never reproduce your "issue pattern". Also could not identify anything temperature related with xtrans images.

There is this pattern i sometimes find (depends on darkroom-centre window and rawimage size) always at the bottom-right edge like here both for xtrans and bayer raws
Bildschirmfoto vom 2023-07-13 22-31-57 

artefacts seem to be introduced in colorin module; dumping shows input ok, output **incorrect**.

jenshannoschwalm:
@kmilos pinging you again - coming back to the issue-pattern you reported and provided a screenshot for. **That** issue i can't yet reproduce here. Could you provide your xmp for that pattern and maybe some more hints?

kmilos:
It's just using defaults on import here, even starting w/ a clean config folder. Zoom to 800%, and then slide along the top edge relatively quickly and release. I haven't had a chance to try the patch yet.

jenshannoschwalm:
The patches won't help for this. Will continue investigating ...

Edit:
1. Could you check please if the issue is also present with PPG demosaicer or Amaze?
2. If not, cou you compile dt with `RCD_MARGIN 7` in `iop/demosaicing/rcd.c`?

jenshannoschwalm:
@kmilos @LebedevRI the good point, we could identify and fix 3 bugs elsewhere causing artefacts in your testing scenario.

Please let me know if you run into issues again...

LebedevRI:
@jenshannoschwalm thank you, but as far as i can tell, err, how do i say this gently,
nothing has changed in current up-to-date darktable master.

Take e.g. `Fujifilm-X-S10-lossless-compressed.RAF` sample from RPU,
zoom in 1600x, and look at the rightmost edge of the image.
The last column of pixels is black. Double the `crop right` setting
in in `raw black/white module`. It's still black.

It appears to originate from `inpaint opposed` mode of `highlight reconstruction` module.

jenshannoschwalm:
> It appears to originate from `inpaint opposed` mode of `highlight reconstruction` module.

I will have a look and fix.

jenshannoschwalm:
> The last column of pixels is black. Double the `crop right` setting
> in in `raw black/white module`. It's still black.

@LebedevRI Can you confirm the issue has been fixed in master / 4.4?

PR #482, opened 2023-06-04
ARW: Support new LJPEG compression on ILCE-7M4 and ILCE-7R5

da-phil:
This PR continues where https://github.com/darktable-org/rawspeed/pull/386 stopped, so all the work in the[ first commit](https://github.com/darktable-org/rawspeed/pull/482/commits/b383e9ead00c0ec3578d5e0b94cbcdcbda322766) has been done by @artemist, I just rebased her changes against recent develop and fixed the merge conflicts before starting to add more work on top.

I'm still not sure about the cropping, however could only verify that it is reasonable, even though a few valid border pixels get cropped away, compared to Sony A7R5 files, which are converted to DNG. This really needs a little bit more attention before we settle on something too early.


LebedevRI:
Thank you for rebasing this!

> This PR continues where #386 stopped, so all the work has been done by @artemist, I just rebased her changes against recent develop and fixed the merge conflicts.

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors might be relevant here.

> <...>

I tried to look in this general direction, and the main stopping point for me is the said boolean `interleaveRows`.
What i have failed to understand from ITU-T81, is how we are supposed to derive the actual MCU size,
from the LJpeg itself?

kmilos:
> What i have failed to understand from ITU-T81, is how we are supposed to derive the actual MCU size,
> from the LJpeg itself?

You won't find this anywhere, as this is AFAICT not the subject of ITU-T81. The only way is to relate the TIFF tile size and SOF3 header yourself, as I have mentioned already: https://github.com/darktable-org/rawspeed/pull/334#issuecomment-1412148764

How you arrange the channels is up to you, you are in charge of the TIFF container abstraction, not ITU-T81 - SOF3 here (restricting to DNG & ARW) just provides simple, **_non-interleaved_** multi-component scans (see section 4.8.1 - it doesn't know or care where components A, B, and C came from). Hence the different schemes concocted by Adobe (Ncomp=2), Blackmagic (Ncomp=1) and Sony (Ncomp=4).

Furthermore, section 4.8.2 says this for the **_non-interleaved_** multi-component scan: 

> For example, in Figure 12 the MCU for the noninterleaved case is a single data unit.

meaning - a single pixel.

So, let's stop confusing the ITU-T81 interleaving w/ CFA interleaving please, which is on a higher (and independent) level of abstraction.

da-phil:
Besides the ITU-T81 discussion I just noticed that the current decoding approach fails with M and S sized compressed raws from my Sony A7r5, [see images](https://www.file-upload.net/download-15154112/compressed_raws.zip.html
). Decoding those (M & S) files gives me the following exception:
``rawspeed::ArwDecoder::DecodeLJpeg(const rawspeed::TiffIFD*) const, line 292: Unexpected bits per pixel: 15``

However, it works with the files which were uploaded to raw.pixls.us: https://raw.pixls.us/data/Sony/ILCE-7RM5

It is possible that the LJPEG coding depends on some other settings (other than size), which apparently I set on my camera?

kmilos:
M & S ARW images are a different can of worms - they are not CFA raws, but demosaiced YCbCr...

src/librawspeed/decoders/ArwDecoder.cpp
```diff
@@ -270,6 +277,79 @@ void ArwDecoder::DecodeUncompressed(const TiffIFD* raw) const {
...
+      const uint32_t length = counts->getU32(tile);
+
+      LJpegDecoder decoder(ByteStream(DataBuffer(mFile.getSubView(offset, length),
+                                                Endianness::little)),
+                          mRaw, true);
+      decoder.decode(tileX * tilew, tileY * tileh, tilew, tileh, false);
+    }
+
+  const TiffEntry* origin_entry = raw->getEntry(TiffTag::DEFAULTCROPORIGIN);
+  const TiffEntry* size_entry = raw->getEntry(TiffTag::DEFAULTCROPSIZE);
```
kmilos: As I already mentioned, [let's not use these but the dimensions in the new tag 0x7038](https://github.com/darktable-org/rawspeed/pull/386#discussion_r1154150298). We then need a new mode that'll tell us to ignore any further cameras.xml crop.

You could even go a step further and use those directly as width and height on L281-L282 without needing to do anything extra here. The LJpeg decoder should take of the padding just fine.

da-phil:
> M & S ARW images are a different can of worms - they are not CFA raws, but demosaiced YCbCr...

OMG, why would you do that? this is quite f*cked up :sweat_smile: 

Answering myself to 
> However, it works with the files which were uploaded to raw.pixls.us: https://raw.pixls.us/data/Sony/ILCE-7RM5

it seems I picked the S35 files, which are further cropped in, however I don't understand why the "normal" M&S files use a completely different encoding (demosaiced YCbCr)?

da-phil:
> I tried to look in this general direction, and the main stopping point for me is the said boolean interleaveRows.
> What i have failed to understand from ITU-T81, is how we are supposed to derive the actual MCU size,
from the LJpeg itself?

@LebedevRI 
When I rename the compressed ARW files to CR2 or DNG, rawspeed seems to still be able to decode the files and somehow deal with the "interleaving". I haven't understood all of the code to explain that to myself yet :sweat_smile: 
But this already shows the potential for refactoring & generalization of all this LJPEG stuff...

kmilos:
> it seems I picked the S35 files, which are further cropped in,

Yes, the S35/APS-C crop is the same size _**in pixels**_ as the M output in FF, but the M are demosaiced and full field of view (i.e. scaled down), while S35/APS-C are still CFAs, only physically cropped FOV.

> why would you do that?

Even smaller files, if you don't care about the resolution, but maybe still care about dynamic range and want to push color and tone processing in post.

Again, different can of worms (CbCr sub-sampling and ITU-T81 interleaving might play a factor here), so I'd ignore M&S for now and tackle them separately later. The S35/APS-C cropped ARWs should just work once this first hurdle is done.

kmilos:
@LebedevRI Actually, after looking at the SOF3 and SOS headers again, there is component interleaving on the MCU level:

Sony SOF3 (14-bit, 256x256 frame, 4 components, 1x1 sampling each): `FF C3 00 14 0E 01 00 01 00 04 01 11 00 02 11 00 03 11 00 04 11 00`
Sony SOS (4 components _interleaved_, single entropy table, horiz. predictor): `FF DA 00 0E 04 01 00 02 00 03 00 04 00 01 00 00`

and, for Adobe DNG:

Adobe SOF3 (16-bit, 128x256 frame, 2 components, 1x1 sampling each): `FF C3 00 0E 10 01 00 00 80 02 00 11 00 01 11 00`
Adobe SOS (2 components _interleaved_, entropy table each, horiz. predictor): `FF DA 00 0A 02 00 00 01 10 01 00 00`

But the [connection between the SOF3 4/2-component frame and the TIFF single channel CFA tile](https://github.com/darktable-org/rawspeed/pull/334#issuecomment-1412148764) is still independent of ITU-T81 and on a different level of abstraction you (i.e. Adobe, Blackmagic, Sony) have to establish and maintain.

And similar for the Blackmagic case, where the number of components is 1, so there is even no MCU interleaving...

Blackmagic SOF3 (12-bit, 2064x1088, 1 component, 1x1 sampling): `FF C3 00 0B 0C 04 40 08 10 01 00 11 00`
Blackmagic SOS (1 component, predictor 6): `FF DA 00 08 01 00 00 06 00 00`

All of these are valid ITU-T81 - it doesn't know or care where the components it was fed came from in these 3 cases, but the bespoke TIFF decoder needs to know how to rearrange these SOF3 frames back into 512x512, 256x256, and 1032x2176 TIFF tiles respectively.

src/librawspeed/decoders/ArwDecoder.cpp
da-phil: Seems to work, I just pushed the changes!

> You could even go a step further and use those directly as width and height on L281-L282 without needing to do anything extra here. The LJpeg decoder should take of the padding just fine.

This however does not work, as the width & height need to be multiples of the tile size to make the tiling code work, or how did you mean that?

kmilos: They don't. The tiling code rounds up to the required multiple.

Well, at least it does in the DNG decoder, don't know if it'll just work out here.

da-phil: > We then need a new mode that'll tell us to ignore any further cameras.xml crop.

Can we make the mandatory cropping step according to the values in ``cameras.xml`` optional by just skipping it on a decoder basis? 
But I guess it would be cleaner and more flexible to introduce a new flag in ``cameras.xml``  a'la 
``<Crop x="0" y="0" width="0" height="0" read_from_exif="true"/>``
If cropping based on exif tags fails cropping shall fall back to values provided in  ``cameras.xml`` .

kmilos: Again, let's not call this "Exif crop". This is a pure kludge on Sony's part.

da-phil: > Well, at least it does in the DNG decoder, don't know if it'll just work out here.

As I said, it doesn't, see check here:
https://github.com/da-phil/rawspeed/blob/1972397b4a6091e3f13fd17a00860560bf15c827/src/librawspeed/decompressors/LJpegDecoder.cpp#L75

> Again, let's not call this "Exif crop". This is a pure kludge on Sony's part.

Is ``autocrop`` better? I don't think naming is our biggest problem with this LJPEG nightmare :sweat_smile:

kmilos: > As I said, it doesn't, see check here

Yeah, this is purely artificial, there is no reason for that constraint. The DNG decoder and its ljpeg decompressor work just fine without this constraint, as they should: https://github.com/darktable-org/rawspeed/blob/1ea45d926df9dbe904eb4b37f118277fbf6a16ab/src/librawspeed/decoders/DngDecoder.cpp#L289-L321

At this point, let's leave this PR as is if it's a working demo, and I'd let @LebedevRI decide on the refactoring going forward.

da-phil: As convoluted as the code is, ``DngDecoder`` eventually calls the same function (``LJpegDecoder::decode``)  in its own decompressor  ``AbstractDngDecompressor::decompressThread<7>()`` and additionally the ``LJpegDecompressor`` ctor does the same check [here](https://github.com/da-phil/rawspeed/blob/1972397b4a6091e3f13fd17a00860560bf15c827/src/librawspeed/decompressors/LJpegDecompressor.cpp#L76). Without those checks the code simply segfaults, as expected. For this to work we need to modify the code which runs over all tiles and allows to process the remaining fraction of tiles at the right and bottom borders. 

But yeah, this little "optimization" will not make a big difference. I'd rather continue with overall refactoring and creating a working solution for the M & S compressed files.

kmilos: > As convoluted as the code is, `DngDecoder` eventually calls the same function (`LJpegDecoder::decode`) in its own decompressor `AbstractDngDecompressor::decompressThread<7>()` and additionally the `LJpegDecompressor` ctor does the same check [here](https://github.com/da-phil/rawspeed/blob/1972397b4a6091e3f13fd17a00860560bf15c827/src/librawspeed/decompressors/LJpegDecompressor.cpp#L76). Without those checks the code simply segfaults, as expected. For this to work we need to modify the code which runs over all tiles and allows to process the remaining fraction of tiles at the right and bottom borders.

Again, we don't. In the DNG case, it really does work today (e.g. convert the uncompressed/lossy ILCE-7M4 to DNG: 7040x4688 is not a multiple of 256x256 tile size, and it doesn't throw). Before you even call `AbstractDngDecompressor::decompress()` and get to that check , the [DNG tile/slice dimensions are calculated for each one on creation](https://github.com/darktable-org/rawspeed/blob/1ea45d926df9dbe904eb4b37f118277fbf6a16ab/src/librawspeed/decoders/DngDecoder.cpp#L398), and [not constant/square on the last column/row](https://github.com/darktable-org/rawspeed/blob/1ea45d926df9dbe904eb4b37f118277fbf6a16ab/src/librawspeed/decompressors/AbstractDngDecompressor.h#L102-L108).

So, the correct behaviour (as outlined in the TIFF spec) is already implemented for the DNG case. Why Sony decided to make a mess of it, I'll never know...

As said, I'll let @LebedevRI decide how to refactor all of this best.

da-phil: Alright, due to the lack of understanding the rawspeed & LJPEG big picture yet I have to believe you ;)

Another question: after the changes in this PR we do not need the cropping which applies the values from `cameras.xml`, where is this applied and can we easily bypass it from within the ARW decoder?

kmilos: > Another question: after the changes in this PR we do not need the cropping which applies the values from `cameras.xml`, where is this applied and can we easily bypass it from within the ARW decoder?

Yes, for now I'd go with ARW decoder overriding these internally rather than bloating `cameras.xml` with more modes straight away if we don't really have to.

kmilos: You probably just want to set `applyCrop` to false if you detect compression 7 [before this call](https://github.com/darktable-org/rawspeed/blob/1ea45d926df9dbe904eb4b37f118277fbf6a16ab/src/librawspeed/decoders/ArwDecoder.cpp#L365), but can't recall now if `decodeRawInternal()` comes before so you could do it there already, or you'd have to detect it again here...

da-phil: Thanks, will have a look later, when I've got time to finish this work package for final review ;)

da-phil: Just checked all the rawspeed entrypoints and `decodeRaw()` always comes before `decodeMetaData()` (both calling their inherited decoder specific function), so this should work. Will give it a try now...

da-phil: by the way, by resetting the history on one of my test images using the `Exif.SubImage1.0x7038` crop I saw that this gives an area which still contains garbage from the tiling operation on the right and a black bar on the bottom
Screenshot from 2023-06-06 22-53-40

If I set the offset to 0,0 I only get valid pixels and no borders, so this way you can get the max. possible pixel area it seems. What's weird though is the size from `Exif.SubImage1.0x7038 = 9568, 6376`, but darktable crop module shows me 9567, 6376.
The DNG converted file shows 9504, 6336 in the crop module :sweat_smile: 

Then I did another test with way bigger crop to see where the garbage & borders begins and this proves that the values in `Exif.SubImage1.0x7038` seem to be quite accurate...
Screenshot from 2023-06-06 23-51-22

kmilos: Rather than resetting for validation, I'd remove the old images from the database (and their xmp sidecars), and import again.

The 0x7038 tag values, as mentioned, should result in the same size as the uncompressed/lossy case w/ `cameras.xml` crop. For 7RM5 this is, for uncompressed/lossy (9600-32)x(6376-0) which matches ljpeg compressed 0x7038 values 9568x6376.

da-phil: Agreed, then the current state (of cropping) in this PR makes sense, right?

src/librawspeed/decoders/ArwDecoder.cpp
```diff
@@ -145,6 +147,13 @@ RawImage ArwDecoder::decodeRawInternal() {
     return mRaw;
   }
 
+  if (7 == compression) {
+    DecodeLJpeg(raw);
+    // cropping of lossless compressed L files already done in Ljpeg decoder
```
kmilos, line 152: Not only L, but also APS-C/S35...

Could also leave a note that M/S are not CFA, and support is TBD. We also probably want to detect this somehow, if not throwing an exception earlier already?

src/librawspeed/decoders/ArwDecoder.cpp
```diff
@@ -270,6 +279,80 @@ void ArwDecoder::DecodeUncompressed(const TiffIFD* raw) const {
...
+                                                Endianness::little)),
+                          mRaw, true);
+      decoder.decode(tileX * tilew, tileY * tileh, tilew, tileh, false);
+    }
+
+  const TiffEntry* size_entry = raw->hasEntry(TiffTag::SONYRAWIMAGESIZE)
+                                    ? raw->getEntry(TiffTag::SONYRAWIMAGESIZE)
+                                    : raw->getEntry(TiffTag::DEFAULTCROPSIZE);
+  iRectangle2D crop(0, 0,
+                    size_entry->getU32(0), size_entry->getU32(1));
```
kmilos, line 352: If you don't find SONYRAWIMAGESIZE (which implies crop origin of 0,0), you better get DEFAULTCROPORIGIN in addition to DEFAULTCROPSIZE and use it as well, you can't assume 0,0 then...

kmilos (changes requested):

src/librawspeed/decoders/ArwDecoder.cpp
da-phil: oh, good point, thanks!

LebedevRI:
Let me just take this over...

LebedevRI:
(Taking over in https://github.com/darktable-org/rawspeed/pull/483)

src/librawspeed/decoders/ArwDecoder.cpp
da-phil: Question is, where in the whole pipeline does the encoding fail, it is not even executing the ARW decoder in case  of M/S images.

da-phil:
> Let me just take this over...

would be great if you have some time to continue this work properly, but please let us know if you need help with some tasks, maybe with the easier ones ;)

LebedevRI:
c38746dae23d6f8fca594be50678bc55607da268

LebedevRI:
> > Let me just take this over...
> 
> would be great if you have some time to continue this work properly

Nah.

> but please let us know if you need help with some tasks, maybe with the easier ones ;)

Yeah i wish i knew. Some small self-contained things could include:
* Get rid of all `hints.has()` calls, there should be a way to tell from the raw itself, most of the time
* Some cameras store black/white levels in EXIF, while some decoders already handle that, many do not (e.g. Canon)
* Black area sampling is just completely broken
* Black level handling needs to be refactored to support arbitrary number (and layout) of bins
* ???

On a high level, i'm still stuck on the general idea to the likes of #325.

LebedevRI:
@da-phil thank you!

PR #483, opened 2023-06-07
ARW: Support new LJPEG compression on ILCE-7M4 and ILCE-7R5

LebedevRI:
Based on https://github.com/darktable-org/rawspeed/pull/482 & https://github.com/darktable-org/rawspeed/pull/386


src/librawspeed/decompressors/LJpegDecoder.cpp
```diff
@@ -101,6 +102,10 @@ void LJpegDecoder::decodeScan() {
 
   int N_COMP = frame.cps;
 
+  const iPoint2D MCUSize = !interleaveRows
+                               ? iPoint2D(frame.cps, 1)
+                               : iPoint2D(frame.cps / 2, frame.cps / 2);
+
```
kmilos, line 108: Is there still a plan to eliminate `interleaveRows` and keep generalizing this?

The reason I ask is, I don't see how this new MCUSize scheme can support the Blackmagic case, which would need something like `iPoint2D(frame.cps / 2, frame.cps * 2)` with `frame.cps / 2` having to be 0.5!?

LebedevRI: > Is there still a plan to eliminate `interleaveRows` and keep generalizing this?

`LJpegDecoder` should not take `interleaveRows`, so i guess this needs to be generalized further, yes.

> The reason I ask is, I don't see how this new MCUSize scheme can support the Blackmagic case, which would need something like `iPoint2D(frame.cps / 2, frame.cps * 2)` with `frame.cps / 2` having to be 0.5!?

No idea. That is a future me problem.

kmilos: > That is a future me problem.

Might warrant a thought though, so you don't have to potentially tear down what you're building up now? Not that there is anything wrong with that if that process works for you ;)

LebedevRI:
Ok, i can't be bothered.

PR #485, opened 2023-06-09
Sony ARW Ljpeg tweaks

kmilos:
Follow-up to #484

### DIFF
--- a/src/librawspeed/decoders/ArwDecoder.cpp
+++ b/src/librawspeed/decoders/ArwDecoder.cpp
@@ -352,9 +352,7 @@ void ArwDecoder::DecodeLJpeg(const TiffIFD* raw) {
 
   PostProcessLJpeg();
 
-  const TiffEntry* size_entry = raw->hasEntry(TiffTag::SONYRAWIMAGESIZE)
-                                    ? raw->getEntry(TiffTag::SONYRAWIMAGESIZE)
-                                    : raw->getEntry(TiffTag::DEFAULTCROPSIZE);
+  const TiffEntry* size_entry = raw->getEntry(TiffTag::SONYRAWIMAGESIZE);
   iRectangle2D crop(0, 0, size_entry->getU32(0), size_entry->getU32(1));
   mRaw->subFrame(crop);
 }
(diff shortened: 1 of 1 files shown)


src/librawspeed/decoders/ArwDecoder.cpp
```diff
@@ -352,9 +357,7 @@ void ArwDecoder::DecodeLJpeg(const TiffIFD* raw) {
 
   PostProcessLJpeg();
 
-  const TiffEntry* size_entry = raw->hasEntry(TiffTag::SONYRAWIMAGESIZE)
-                                    ? raw->getEntry(TiffTag::SONYRAWIMAGESIZE)
-                                    : raw->getEntry(TiffTag::DEFAULTCROPSIZE);
+  const TiffEntry* size_entry = raw->getEntry(TiffTag::SONYRAWIMAGESIZE);
```
da-phil, line 360: so no fallback is necessary, we're always sure to have this EXIF marker available?

kmilos: > we're always sure to have this EXIF marker available?

Seems to be the case so far, as mentioned.

In the worst case, I'd rather return the uncropped image w/ black borders than the DNG crop, I feel it's more consistent with the stated behaviour:

> RawSpeed does NOT…
> 
> crop the image to the same sizes as manufactures, but supplies biggest possible images.

LebedevRI: FWIW, `getEntry()` throws if you query for an entry that will not be found.

LebedevRI:
Makes sense, thank you!

kmilos:
Do we feel cheeky enough to drop this on dt? ;)

I suspect they'll [freeze and tag 4.4 very soon](https://github.com/darktable-org/darktable/wiki/Releases-cycle)... @TurboGit

LebedevRI:
Even though `LJpegDecompressor` is no longer affected, i'd still prefer to let it fuzz for a few more days.
What is the cut-off date?

TurboGit:
@LebedevRI : It is planed for June 14th, I can move it one day on 15th if necessary.

LebedevRI:
Ah, that's plenty of time then.
I'll wait some more, and update submodule on 13'th then.

PR #488, opened 2023-06-13
Sony ZV-1M2 support

kmilos:
This should be [the same sensor as ZV-1](https://petapixel.com/2023/05/26/sony-zv-1-vs-zv-1m2-an-upgrade-of-compromises-and-diminishing-returns/), and confirmed in ADC 15.4 the color matrix is indeed identical.


LebedevRI:
Hopefully closing&reopening will retrigger CI.

PR #490, opened 2023-06-29
Crop Panasonic G9 images using vendor metadata

### LINKED ISSUE #170: Panasonic DC-G9 high-res RAW: insufficient crop
When opening a High Resolution RAW (the large variant with ~125 MB file size) with darktable 2.6.0, artifacts can be seen on the right border of the picture (panorama) or on the top border (portrait).

This issue was approved by @LebedevRI. More details are to follow if needed.

jsmucr:
This is supposed to fix #170 and help prepare other related fixes.

Now the decoders can decide whether to supply their own crop data when the `Crop` tag in a `camera.xml` entry is empty or absent. `Rw2Decoder` has been updated to leverage this new ability, and the `Crop` tags for Panasonic G9 entries have been removed.


LebedevRI:
Github suggests this isn't rebased properly.

jsmucr:
> Github suggests this isn't rebased properly.

I know. My original intention has been to test it against DT 4.2.1 because that's what I use, hence the rebase. I'll put it back later.

jsmucr:
Ok, it appears to work as intended with DT 4.2.1. Rebased back onto develop.

LebedevRI:
Thanks for looking into this!

I do think something like this is needed, but:
this perhaps needs to be more fine-grained,
and perhaps less fine-grained, i.e.:
* Should this only apply to the problematic mode?
* is this really the only affected camera?
* Should we just always use this vendor crop for panasonic?

(I don't know the answer to these questions, this isn't a recommendation.)

Thoughts?

jsmucr:
The most problematic thing here is in fact to figure out the mode, as here are many combinations of how you can shoot. What I was addressing here was primarily my issue with hi-res pixel shift mode and *some* of the burst modes. I wasn't able to figure it out just by looking at the metadata. I would have to take the camera and literally try out every single setting to make a note about the crop. And then maybe a software update would come out and I would have to start over to make sure they didn't change anything related. And even then I could get it wrong without some serious pixel peeping. 

It's definitely not the only camera affected. There was this issue with a Panasonic camera, I think LX7, which seemed to crop raw files when using digital zoom.
But generally, I was very surprised to see there's a database of cameras with reverse engineered crops for each mode (whatever that means). I myself wouldn't dare to override vendor settings unless necessary. 

I can only provide knowledge regarding the G9 and the GX80 Panasonic cameras. And lend a hand in checking the available samples of the other models. Until then, falling back to vendor settings could provide a fast hotfix whenever someone comes up with a similar issue.

jsmucr:
Will this get merged, or shall I do something more with this fix?

LebedevRI (approved):
Thank you!
Sorry for the silence, i suppose this is better than what we do now :)

High-level, the particular approach feels somewhat wrong,
i mean, how do those with `applyCrop = false` deal with this?
But that is not a blocker i suppose.

PR #508, opened 2023-08-08
Removes references to the `rawspeed.org` domain that no longer exists.

xsdg:


CameronNemo:
Is it worth pointing to the Internet Archive? https://web.archive.org/web/20230202134713/https://rawspeed.org/doxygen/

Is this API documentation hosted elsewhere now (or in the future)?

Edit: the archive is pretty useless on further investigation, but maybe a good idea to point out that there is Doxygen-based documentation that you have to build yourself?

PR #510, opened 2023-08-09
Remove some unsupported Samsung DNG model flags

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/12160, https://github.com/darktable-org/darktable/issues/14094 and https://github.com/darktable-org/darktable/issues/14219

CC0 samples are indeed present on RPU...

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -17521,16 +17521,16 @@
 	<Camera make="SAMSUNG DIGITAL IMA" model="SAMSUNG GX20" mode="dng" supported="no-samples">
 		<ID make="Samsung" model="GX20">Samsung GX20</ID>
 	</Camera>
-	<Camera make="samsung" model="SM-G920F" supported="no-samples">
+	<Camera make="samsung" model="SM-G920F" mode="dng">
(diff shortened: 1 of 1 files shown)


LebedevRI:
I think the problem here is that these should be `mode="dng"`.
(this is likely true for some other `missing-samples` entries)

kmilos:
True, I found two more Samsung models like that.

LebedevRI:
Thank you!

jiemdev:
Hi,
Is it normal, with darktable version 4.8.1, to still get these error messages?

`RawSpeed:Camera support status is unknown: 'SAMSUNG DIGITAL IMA' 'SAMSUNG GX20' 'dng'
Please consider providing samples on <https://raw.pixls.us/> if you wish for the support to not be discontinued, thanks!`

kmilos:
@jiemdev Note that this PR was for the 'SAMSUNG TECHWIN Co.' camera make, not 'SAMSUNG DIGITAL IMA', and these are (for better or worse) currently listed as separate/different devices in the database. Unfortunately https://raw.pixls.us is still down so I can't check if we are still indeed missing the sample for the latter device in order for the warning to be finally removed...

PR #512, opened 2023-08-16
Update in-tree version of openmp from 7.0.0 to 14.0.6

mallman:
This PR updates the in-tree version of openmp from 7.0.0 to 14.0.6. It's still not the latest-and-greatest, but it's later and greater than version 7.

I've been using this version for a while without problems.

### DIFF
--- a/cmake/Modules/LLVMOpenMP.cmake.in
+++ b/cmake/Modules/LLVMOpenMP.cmake.in
@@ -33,8 +33,8 @@ else()
     message(WARNING "Did not find LLVM OpenMP runtime library sources! Fetching from web...")
     ExternalProject_Add(
       llvm-openmp
-      URL               https://releases.llvm.org/7.0.0/openmp-7.0.0.src.tar.xz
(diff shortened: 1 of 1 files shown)


LebedevRI:
Note that options were generally meant as an debug-time escape hatches,
not for any kind of production use, there you really should get said library with the compiler/system.

There's zero CI coverage for them, and i haven't really even test-built with them for some time.

mallman:
> Note that options were generally meant as an debug-time escape hatches, not for any kind of production use, there you really should get said library with the compiler/system.
> 
> There's zero CI coverage for them, and i haven't really even test-built with them for some time.

Understood. The in-tree openmp makes cross-compilation and deployment _much_ easier. Cross-compiling, especially across deployment platforms, using system library dependencies is an incredible PITA. It's usually easier to compile and link a source dep than link from some ensemble of system binaries.

mallman:
FWIW, I take the same approach with pugixml. The in-tree pugixml build makes cross-compiling RawSpeed a ton easier.

mallman:
I just published this this morning: https://github.com/mallman/RawDecoder.

That is where I'm doing my cross-platform build. It uses the in-tree openmp if the user configures it so, but by default it disables openmp support.

LebedevRI:
I'll merge this, but i have not done any due diligence on this change.

Thank you!

mallman:
I appreciate it, thank you. 🙏

PR #514, opened 2023-08-17
Sigma sd Quattro H (partial) support 

kmilos:
DNG mode name normalization only

Partly addresses https://github.com/darktable-org/darktable/issues/2879


LebedevRI:
@kmilos thank you.

PR #515, opened 2023-08-21
Improved calculation of DNG white balance from AsShotWhiteXY

kmilos:
Attempt to address https://github.com/darktable-org/darktable/issues/14805 @jenshannoschwalm @mcogoni

Not sure if an additional chromatic adaptation step is needed on the as shot XYZ values (relative to D65) before using the D65 matrix on them...

### DIFF
--- a/src/librawspeed/decoders/DngDecoder.cpp
+++ b/src/librawspeed/decoders/DngDecoder.cpp
@@ -662,6 +662,8 @@ void DngDecoder::decodeMetaDataInternal(const CameraMetaData* meta) {
     }
   }
 
+  parseColorMatrix();
+
   // Fetch the white balance
   if (mRootIFD->hasEntryRecursive(TiffTag::ASSHOTNEUTRAL)) {
     const TiffEntry* as_shot_neutral =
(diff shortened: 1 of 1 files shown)


src/librawspeed/decoders/DngDecoder.cpp
```diff
@@ -672,22 +674,25 @@ void DngDecoder::decodeMetaDataInternal(const CameraMetaData* meta) {
...
-      mRaw->metadata.wbCoeffs[1] = as_shot_white_xy->getFloat(1);
-      mRaw->metadata.wbCoeffs[2] =
-          1 - mRaw->metadata.wbCoeffs[0] - mRaw->metadata.wbCoeffs[1];
-
-      const std::array<float, 3> d65_white = {{0.950456, 1, 1.088754}};
-      for (uint32_t i = 0; i < 3; i++)
-        mRaw->metadata.wbCoeffs[i] /= d65_white[i];
+      const float x = as_shot_white_xy->getFloat(0);
+      const float y = as_shot_white_xy->getFloat(1);
+      const std::array<float, 3> as_shot_white = {{x / y, 1, (1 - x - y) / y}};
```
LebedevRI: What should happen when `y` is zero?

LebedevRI: Please add a link to http://www.brucelindbloom.com/index.html?Eqn_xyY_to_XYZ.html,
replacing hardcoded `1` with `constexpr float Y = 1.0;`

src/librawspeed/decoders/DngDecoder.cpp
```diff
@@ -672,22 +674,25 @@ void DngDecoder::decodeMetaDataInternal(const CameraMetaData* meta) {
...
+             mRootIFD->hasEntryRecursive(TiffTag::ASSHOTWHITEXY)) {
     const TiffEntry* as_shot_white_xy =
         mRootIFD->getEntryRecursive(TiffTag::ASSHOTWHITEXY);
     if (as_shot_white_xy->count == 2) {
-      mRaw->metadata.wbCoeffs[0] = as_shot_white_xy->getFloat(0);
-      mRaw->metadata.wbCoeffs[1] = as_shot_white_xy->getFloat(1);
-      mRaw->metadata.wbCoeffs[2] =
-          1 - mRaw->metadata.wbCoeffs[0] - mRaw->metadata.wbCoeffs[1];
-
-      const std::array<float, 3> d65_white = {{0.950456, 1, 1.088754}};
```
LebedevRI, line 684: Hmm yeah, hardcoding D65 here clearly looks suspicious.

src/librawspeed/decoders/DngDecoder.cpp
kmilos: Done

kmilos: Done, just ignored

LebedevRI:
Ok, i agree that this is an improvement, it very obviously fixes an existing `CRW_8531.DNG` sample.
This roughly matches what `Translating White Balance xy Coordinates to Camera Neutral
Coordinates` says in the DNG spec, other than the rest of advanced steps mentioned there.

This "regresses" things in case where `CalibrationIlluminant*` is missing (e.g. `Nikon LS-5000` sample),
or where we don't have a color matrix w/ `CalibrationIlluminant = D65`
(since that is what `DngDecoder::parseColorMatrix()` requires).

Thank you!

PR #517, opened 2023-09-06
Add missing schema ImBack camera maker

kmilos:
Cf. recent commit https://github.com/darktable-org/rawspeed/commit/59cf9f95c4c770f7161b7bb54c7f52d4b6c91922

Should it be "ImBack" or "Imback"? We left GoPro, OnePlus, etc. alone AFAICT?

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -16270,7 +16270,7 @@
 		</ColorMatrices>
 	</Camera>
 	<Camera make="ImBack" model="ImB35mm" mode="chdk">
-		<ID make="Imback" model="ImB35mm">ImBack</ID>
+		<ID make="ImBack" model="ImB35mm">ImBack</ID>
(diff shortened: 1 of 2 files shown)


LebedevRI:
I'm reasonably sure i just used what was in the DNG provided by the linked tool.
To be fully honest, i do not particularly understand this name normalization/sanitization,
it's rules really need to be documented.

LebedevRI:
Thank you!

kmilos:
> To be fully honest, i do not particularly understand this name normalization/sanitization,
> it's rules really need to be documented.

Yeah, I think the dngmeta.sh script does a blind sentence capitalization for the rawspeed internal ID maker attribute (this was to handle and prettify NIKON, RICOH, SONY, etc.), so there is still a need to double-check this manually...

PR #521, opened 2023-09-20
Relax DNG black level parsing

kmilos:
Use the default BlackLevelRepeatDim (1,1) if the single entry value is 1

This e.g. allows the black level of Pentax K-3 Mark III Monochrome DNG from RPU to be parsed instead of being ignored:
```
Exif.SubImage1.BlackLevelRepeatDim           Short       1  1
Exif.SubImage1.BlackLevel                    Long        1  2048
```

~~Alternatively, could read the one value if count is 1, and copy it?~~

### DIFF
--- a/src/librawspeed/decoders/DngDecoder.cpp
+++ b/src/librawspeed/decoders/DngDecoder.cpp
@@ -767,9 +767,13 @@ bool DngDecoder::decodeBlackLevels(const TiffIFD* raw) const {
   iPoint2D blackdim(1, 1);
   if (raw->hasEntry(TiffTag::BLACKLEVELREPEATDIM)) {
     const TiffEntry* bleveldim = raw->getEntry(TiffTag::BLACKLEVELREPEATDIM);
-    if (bleveldim->count != 2)
+    if (bleveldim->count == 2)
+      blackdim = iPoint2D(bleveldim->getU32(0), bleveldim->getU32(1));
+    else if (bleveldim->count != 1 || 1 != bleveldim->getU32(0)) {
+      writeLog(DEBUG_PRIO::EXTRA,
+               "BLACKLEVELREPEATDIM entry is invalid, not allowed to guess");
       return false;
-    blackdim = iPoint2D(bleveldim->getU32(0), bleveldim->getU32(1));
+    }
   }
 
   if (!blackdim.hasPositiveArea())


LebedevRI:
What were they thinking? :/
Was this reported to Pentax?

In this case, i think i'm, at most, okay with

> Alternatively, could read the one value if count is 1, and copy it?

kmilos:
> Was this reported to Pentax?

Don't think so. You know where/how?

src/librawspeed/decoders/DngDecoder.cpp
```diff
@@ -767,9 +767,12 @@ bool DngDecoder::decodeBlackLevels(const TiffIFD* raw) const {
   iPoint2D blackdim(1, 1);
   if (raw->hasEntry(TiffTag::BLACKLEVELREPEATDIM)) {
     const TiffEntry* bleveldim = raw->getEntry(TiffTag::BLACKLEVELREPEATDIM);
-    if (bleveldim->count != 2)
+    if (bleveldim->count == 1)
+      blackdim = iPoint2D(bleveldim->getU32(0), bleveldim->getU32(0));
+    else if (bleveldim->count == 2)
+      blackdim = iPoint2D(bleveldim->getU32(0), bleveldim->getU32(1));
+    else
       return false;
```
LebedevRI:
```
if (bleveldim->count == 2)
  blackdim = iPoint2D(bleveldim->getU32(0), bleveldim->getU32(1));
else if (bleveldim->count == 1) {
  blackdim = iPoint2D(bleveldim->getU32(0), bleveldim->getU32(0));
  if (bleveldim != {1,1})
    return false;
  writeLog("???");
else
  return false;
}
```

kmilos: Why do you assume a single value must denote (1,1)? Just because we have the one corner case in the wild? What's the difference then to the previous version falling back to default (1,1)?

LebedevRI: > Why do you assume a single value must denote (1,1)?

Why do you assume a single value must denote (N,N)?
Maybe it means (N/2,N/2) or something?

> Why do you assume a single value must denote (1,1)?

I'm not assuming that. I'm just saying that this workaround
is just completely horrible, and i'm not okay with
blindly accepting random crap values, only `(1)`,
because it's rather unambigious.

> What's the difference then to the previous version falling back to default (1,1)?

There we'd also accept obviously-broken 3-component size.

kmilos: Ok, let's cover this particular case then only...

LebedevRI:
> > Was this reported to Pentax?
> 
> Don't think so. You know where/how?

I haven't managed to find any kind of non-user-facing side,
so i guess the usual "straight-to-spam" https://www.ricoh.com/contact

LebedevRI:
@milos thank you.

But this really needs to be reported to Ricoh/Pentax.

kmilos:
Will do. I guess it doesn't hurt and improves the chances if you do it as well?

image

PR #522, opened 2023-09-25
Add Canon EOS Kiss F alias

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/15299

The "Canon EOS Kiss Digital F" was probably wrong to begin with, but probably doesn't hurt to keep it...

Edit: Added a few more while at it, mainly from random sampling https://global.canon/en/c-museum/camera.html

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -1433,6 +1433,7 @@
 		</BlackAreas>
 		<Aliases>
 			<Alias id="EOS Rebel T3">Canon EOS REBEL T3</Alias>
+			<Alias id="EOS Kiss X50">Canon EOS Kiss X50</Alias>
 		</Aliases>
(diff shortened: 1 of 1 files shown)


LebedevRI:
I think basically all Canon CR2 three or four-digit cameras (XXXXD/XXXD) have a few missing aliases.

LebedevRI:
@kmilos thank you.

kmilos:
> I think basically all Canon CR2 three or four-digit cameras (XXXXD/XXXD) have a few missing aliases.

I just spent a bit more time on that museum page and [wikipedia](https://en.wikipedia.org/wiki/Template:Canon_EOS_digital_cameras), I _think_ rawspeed should cover most now w/ Europe/US/Japan models...

PR #523, opened 2023-09-27
Nikon Coolpix A1000 support

kmilos:
This is the one remaining Coolpix model that RPU has a sample for, however:

* Missing crop - heavy vignetting present, but no clues in Exif or makernotes, nor DNG
* Incorrect white level - default 12-bit 4095 max is too high, but no info in Exif or makernotes

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -6429,6 +6429,27 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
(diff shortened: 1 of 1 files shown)


LebedevRI:
@kmilos thank you.
White level of ~3890 seems somewhat correct-ish.

kmilos:
Sounds about right.

PR #525, opened 2023-10-05
Explicitly allow fallback mode for Panasonic cameras

kmilos:
This is to avoid throwing exceptions when e.g. not all crop modes are present, but the basic one works.

Alternative fix for https://github.com/darktable-org/darktable/issues/15368 as all crops are identical and covered by the basic mode.

### DIFF
--- a/src/librawspeed/decoders/Rw2Decoder.cpp
+++ b/src/librawspeed/decoders/Rw2Decoder.cpp
@@ -176,8 +176,13 @@ RawImage Rw2Decoder::decodeRawInternal() {
 
 void Rw2Decoder::checkSupportInternal(const CameraMetaData* meta) {
   auto id = mRootIFD->getID();
-  if (!checkCameraSupported(meta, id, guessMode()))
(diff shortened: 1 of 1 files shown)


LebedevRI:
Does this work in the situation of the camera where the crops are not identical?

kmilos:
I think so.

This just actually enables the "empty mode" fallback regardless of the `failOnUnknown` setting, the previous non-empty mode detection should be unaffected, and has priority.

LebedevRI:
This is working as intended. I don't think we can fall back to the default entry iff we were explicitly told not to do that.
That being said, as i have previously said, this whole `cameras.xml` business is a messy mess.
Perhaps those mode-less modes should not be there in the first place?

kmilos:
> This is working as intended.

Without a spec, it's difficult to say what was actually intended, especially because of existing
```
if (!checkCameraSupported(meta, id, guessMode()))
    checkCameraSupported(meta, id, "");
```

> Perhaps those mode-less modes should not be there in the first place?

Case in point.

PR #526, opened 2023-10-05
Drop erroneous newline in `data/README.md` (spurious diff to stable branch)

kmilos:
Cosmetic only to minimize the diff output between the two branches.


LebedevRI:
Huh, i don't quite understand how that difference ended up existing in the first place.

LebedevRI:
@kmilos Thank you.

kmilos:
> Huh, i don't quite understand how that difference ended up existing in the first place.

Dunno. But I did see similar situations when merging branches on other projects as well...

PR #527, opened 2023-10-05
Sync differences to develop branch

kmilos:
Cosmetic only to minimize the diff output between the two branches.


LebedevRI:
Hm indeed. I've fixed that in 5eb7e2685c0e53d9c6e113f8b5c6d3f2c66398cd, but github is being petty/picky as usual.
Thank you.

PR #529, opened 2023-10-11
Panasonic DC-G9M2 support (data only)

kmilos:
Partly addresses https://github.com/darktable-org/darktable/issues/15679

Used ADC 16.0. Crop is unverified.

This depends on https://github.com/darktable-org/rawspeed/issues/367

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -11410,8 +11410,29 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="Panasonic" model="DC-G9M2" supported="no">
+	<Camera make="Panasonic" model="DC-G9M2">
 		<ID make="Panasonic" model="DC-G9M2">Panasonic DC-G9M2</ID>
+		<Crop x="0" y="0" width="0" height="0"/>
+		<Sensor black="2048" white="65535"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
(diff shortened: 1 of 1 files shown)


LebedevRI:
@kmilos hmm you know what we could do?

Since we are polluting `cameras.xml` with garbage anyways,
why don't we add "dormant" entries for cameras that exist in the wild,
(of which we are ignorant of), e.g. for the purposes of RPU sample asking
(RPU does that automatically more or less).

We will need `supported="unknown"` though, so rawspeed does not
start considered them supported since it's present in `cameras.xml`.

Does that sound interesting to you?

kmilos:
Sorry Roman, I might be slow these days - how is this exactly different/what's the benefit compared to what we have today: error and ask for samples if camera is not supported, or mode='no-samples'?

LebedevRI:
`mode='no-samples'` means "this camera was supposedly explicitly supported, but we don't really know".
Note the word "explicitly". Currently we can't add a camera to `cameras.xml` without
deeming it either supported or unsupported.

`supported="unknown-no-samples"` would allow us to list (in `cameras.xml`) more known
(to exist in the real world) cameras, without deeming them either supported or unsupported.
That could be helpful e.g. for RPU sample request lists.

This, of course, hinges on the access to such list of real-world cameras.

kmilos:
> This, of course, hinges on the access to such list of real-world cameras.

So far, I've just been using https://www.dpreview.com/products/timeline?year=all&brand=&category=cameras for manually checking some coverage and to keep on top of new cameras...

PR #530, opened 2023-10-12
Fujifilm GFX100 II support

kmilos:
Using ADC 16.0

Currently uncompressed and lossless compressed only, as with other Fujifilm cameras.

https://github.com/darktable-org/darktable/issues/16130

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -15033,6 +15033,42 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="FUJIFILM" model="GFX100 II">
+		<ID make="Fujifilm" model="GFX100 II">Fujifilm GFX 100 II</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
+		</CFA>
+		<Crop x="0" y="0" width="-146" height="-6"/>
+		<Sensor black="0" white="0"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">12806 -5779 -1110</ColorMatrixRow>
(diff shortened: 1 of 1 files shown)


kmilos:
@LebedevRI [I have a few open PRs](https://github.com/darktable-org/rawspeed/pulls?q=is%3Aopen+author%3Akmilos+sort%3Aupdated-desc) where codecov seems to be running periodically and updating the status, you have any idea what might be triggering it?

LebedevRI:
> @LebedevRI [I have a few open PRs](https://github.com/darktable-org/rawspeed/pulls?q=is%3Aopen+author%3Akmilos+sort%3Aupdated-desc) where codecov seems to be running periodically and updating the status, you have any idea what might be triggering it?

I don't have an answer but i could guess that it either reacted to the new changes in the `develop` branch,
or some github CI job completed/timeouted and it updated the status accordingly. Does not seem erroneous to me.

kmilos:
Updated the crop slightly, should be good to go now - seems to be working ok on those RPU samples (except the lossy ones of course).

P.S. I think you can remove the 32-bit MSYS environments from CI, [they're being phased out](https://www.msys2.org/news/#2023-12-13-starting-to-drop-some-32-bit-packages).

LebedevRI:
> P.S. I think you can remove the 32-bit MSYS environments from CI, [they're being phased out](https://www.msys2.org/news/#2023-12-13-starting-to-drop-some-32-bit-packages).

Right. At this point i'm not really sure if this library _needs_ to support 32-bit environment
(in general, not just for some specific OS), so while i've tried to keep 32-bit CI working
(openSUSE_Tumbleweed @ i586 is essentially dead too),
i'm not sure how much effort is worth spending on that.

But once said CI can not be used anymore, 32-bit env would also be blocked via a cmake check like compiler versions.

LebedevRI:
I can not verify crop at all here, samples are really bad.

LebedevRI:
@kmilos thank you!

kmilos:
Pretty sure it's meant to be the same as Mark I...

PR #531, opened 2023-10-22
Basic Sinarback eVolution decoder

LebedevRI:
Seems to kind-of work, but `CFARepeatPatternDim` is `2,0` (??)

Refs. https://github.com/darktable-org/darktable/issues/15466

### DIFF
--- a/src/librawspeed/decoders/StiDecoder.cpp
+++ b/src/librawspeed/decoders/StiDecoder.cpp
@@ -1,0 +1,97 @@
+/*
+    RawSpeed - RAW file decoder.
+
+    Copyright (C) 2023 Roman Lebedev
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2 of the License, or (at your option) any later version.
+
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public
+    License along with this library; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+*/
+
+#include "decoders/StiDecoder.h"
+#include "adt/Point.h"                              // for iPoint2D
+#include "decoders/RawDecoderException.h"           // for ThrowException
+#include "decompressors/HasselbladLJpegDecoder.h"   // for HasselbladLJpegDec...
+#include "decompressors/UncompressedDecompressor.h" // for UncompressedDeco...
+#include "io/Buffer.h"                              // for Buffer, DataBuffer
+#include "io/ByteStream.h"                          // for ByteStream
+#include "io/Endianness.h"                          // for Endianness, Endian...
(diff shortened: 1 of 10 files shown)


src/librawspeed/decoders/StiDecoder.cpp
```diff
@@ -0,0 +1,97 @@
...
+void StiDecoder::DecodeUncompressed(const TiffIFD* raw) const {
+  if (mRaw->getDataType() != RawImageType::UINT16)
+    ThrowRDE("Unexpected data type");
+
+  if (mRaw->getCpp() != 1 || mRaw->getBpp() != sizeof(uint16_t))
+    ThrowRDE("Unexpected cpp: %u", mRaw->getCpp());
+
+  // FIXME: could be wrong.
+  if (!mRaw->dim.hasPositiveArea() || mRaw->dim.x % 2 != 0 ||
+      mRaw->dim.y % 2 != 0 || mRaw->dim.x > 4992 || mRaw->dim.y > 6668) {
```
kmilos, line 76: Is the RPU sample in portrait by any chance? Is it possible that the camera outputs an already oriented raw raster? The [spec](https://sinar.swiss/assets/Uploads/Manual-Sinarback-eVolution-EN.pdf) says 6668 x 4992 4:3 ([landscape](https://peartreephoto.com/used-sinar-digital-backs/used-sinar-evolution-75h-digital-back)), not the other way around... So I guess one needs to allow for both possibilities here?

LebedevRI: > Is the RPU sample in portrait by any chance?

@robert-olding ?

kmilos: Note also that the raw IFD has:
```
[IFD1]          Orientation                     : Rotate 180
```
vs. the no-op orientation of the root IFD:
```
[IFD0]          Orientation                     : Horizontal (normal)
```

It would be good to check a landscape shot as well...

robert-olding: Yes. The shot was taken with the digital back oriented in the vertical (portrait) position.

kmilos: @robert-olding Can you share (upload to RPU) a landscape sample as well please? If not (you mentioned privacy issues), can you at least share the output of e.g. `exiftool -a -u -s -G1 -w txt <landscape_file>.sti`?

LebedevRI: `Addy_and_Odin_Image_0095.sti` is also portrait-oriented apparently.
I'm guessing the camera always writes images that way.

robert-olding: The Addy_and_Odin file were shot with the digital back oriented in the landscape position

LebedevRI: That's what i said, yes.

kmilos: So the sensor really is scanned that way it would seem 

However, I still need the various Orientation tags from the landscape case.

LebedevRI: `[IFD1]          Orientation                     : Rotate 270 CW`
[Addy_and_Odin_Image_0095.txt](https://github.com/darktable-org/rawspeed/files/13078013/Addy_and_Odin_Image_0095.txt)

kmilos: Ta. I guess that's the one we'd need to specially read in dt then in case of Sinar.

PR #540, opened 2023-10-27
Leica M10-D, M10-P and M10-R name normalisation

kmilos:
They keep popping up...


LebedevRI:
We really should either be doing that automatically (how are these derived manually?),
or not doing it at all. It is intrinsically wrong that DNG files require such manual adjustments.

LebedevRI:
Thank you.

PR #542, opened 2023-10-27
Leica SL2 and SL2-S name normalisation

kmilos:


LebedevRI:
Maybe these kind of changes should be bunched up in a single large PR?

Thank you.

kmilos:
You mean just the DNG naming ones? Sure.

Anyway, I think I'm done covering the Leicas for the time being, all the major models going back to 2019 are there now.

PR #543, opened 2023-10-30
Nikon Z f support (lossless only)

kmilos:
Using latest ADC 16


LebedevRI:
@kmilos thank you.

In all of the PR's in this batch, i didn't spot wrong crop in any of them,
so unless something is yet again broken on dt's side, nice.

PR #544, opened 2023-10-30
Samsung EK-GN120: re-introduce support

ge0rg:
I've uploaded a sample to RPU (but can't see it now on the website?!), but also took the occasion to just fix the cameras.xml - the EK-GN120 is using the same sensor as the NX300, and now the entries are identical.

SRW from the Galaxy NX (EK-GN120) look good in darktable now as well.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -12742,12 +12742,12 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="SAMSUNG" model="EK-GN120" decoder_version="3" supported="no-samples">
+	<Camera make="SAMSUNG" model="EK-GN120" decoder_version="3">
 		<ID make="Samsung" model="EK-GN120">Samsung EK-GN120</ID>
 		<CFA width="2" height="2">
 			<Color x="0" y="0">GREEN</Color>
-			<Color x="1" y="0">BLUE</Color>
-			<Color x="0" y="1">RED</Color>
+			<Color x="1" y="0">RED</Color>
+			<Color x="0" y="1">BLUE</Color>
 			<Color x="1" y="1">GREEN</Color>
 		</CFA>
 		<Crop x="45" y="25" width="-11" height="-29"/>
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -12743,7 +12743,7 @@
 		</ColorMatrices>
 	</Camera>
 	<Camera make="SAMSUNG" model="EK-GN120" decoder_version="3">
-		<ID make="Samsung" model="EK-GN120">Samsung EK-GN120</ID>
+		<ID make="Samsung" model="EK-GN120">Samsung Galaxy NX</ID>
 		<CFA width="2" height="2">
 			<Color x="0" y="0">GREEN</Color>
 			<Color x="1" y="0">RED</Color>


data/cameras.xml
```diff
@@ -12742,12 +12742,12 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="SAMSUNG" model="EK-GN120" decoder_version="3" supported="no-samples">
-		<ID make="Samsung" model="EK-GN120">Samsung EK-GN120</ID>
+	<Camera make="SAMSUNG" model="EK-GN120" decoder_version="3">
+		<ID make="Samsung" model="EK-GN120">Samsung Galaxy NX</ID>
```
LebedevRI, line 12746: I'm not really sure this kind of normalization is common here.
I think let's not do that.

ge0rg: Well, it's officially called the Galaxy NX, but this is not a hill for me to die on :)

LebedevRI: Oh wait, i misread the change as-if the new maker was non-`Samsung`. Let's leave this in. Sorry.

kmilos, line 12746:
```suggestion
		<ID make="Samsung" model="EK-GN120">Samsung NX U</ID>
```

This actually comes from UniqueCameraModel after Adobe DNG Conversion.

ge0rg: I've spent quite some time studying Samsung NX and the models created by them, and I never encountered a "Samsung NX U". There is an [April's Fool "NX Ultra" joke](https://www.digitalcameraworld.com/news/samsung-nx-ultra-samsung-resurrects-camera-line-with-108mp-s20-ultra-sensor), but the "NX U" only seems to exist in Adobe's DNG and websites that use it for listing the cameras they support. The official name is [Samsung Galaxy NX](https://www.samsungmobilepress.com/press-releases/introducing-galaxy-nx-the-first-interchangeable-lens-camera-with-3g-4g-lte-wi-fi-connectivity) and there were two internal model numbers: EK-GN100 and EK-GN120.

LebedevRI: I don't understand this `<ID>` name remangling, and there's no docs.
My best guess is that indeed, the idea is to really be 1:1 with ADC.

ge0rg: I understand the urge to be consistent, but quite obviously Adobe is in a dead-end here. I'd really appreciate showing the name that's actually printed on the device ("Galaxy NX"), or at least the internal model number ("EK-GN120", I'm pretty sure it's part of the fineprint on the body label, will check on Saturday). The ADC name will most probably only confuse the users.

kmilos: TBH, I haven't seen the [actual value](https://github.com/darktable-org/rawspeed/blob/39625068a2a4397a0169d84ffab5c40fe4809bad/src/librawspeed/metadata/Camera.cpp#L299) being used for anything meaningful in rawspeed nor darktable, whereas the `make` and `model` attributes are what seems to really count for the normalisation...

The [dngmeta script](https://github.com/darktable-org/darktable/blob/95c8e4176d603608eb3b6ac6e0b37b8eb4299a9b/tools/dngmeta.sh#L41) populates the ID value with the UniqueCameraModel of the DNG (if not native for the camera, then one produced by ADC), but I guess it's not the end of the world if we override it...

ge0rg: Maybe I'm mistaking the ID field, and it's not supposed to contain the human readable name of the camera?

kmilos: > Maybe I'm mistaking the ID field, and it's not supposed to contain the human readable name of the camera?

Yes, but at least in rawspeed tools and darktable, it is constructed from `make` + `model` ID attributes, not the ID value. 🤷 

In this case it only helps converting "SAMSUNG" -> "Samsung", but you'll find more useful example pairs in the file.

ge0rg:
Okay, given the controversy about the `<ID>` field, feel free to just cherry-pick the first commit and close this as done. Alternatively, I can remove the second commit from the PR to get this resolved. Thanks! :)

LebedevRI:
@ge0rg thank you!

PR #565, opened 2023-11-20
Canon IXY 220F support

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/15598

See https://global.canon/en/c-museum/product/dcc619.html for regional aliases.

Edit: went for the max crop for the CHDK mode, different/larger than the DNG.


LebedevRI:
@kmilos thank you!

Issue #234, opened 2020-06-06
TIFF IFD: 'give me the makernote IFD'


LebedevRI:
So instead of "give me the IFD that constains the magical 0x8c" is it not just as simple as "give me the IFD which is a MakerNote" since we know it'll be a Nikon MakerNote anyway (the file is already identifed as a Nikon raw)? And then you check for 0x96, 0x8c, and eventually fail.

_Originally posted by @kmilos in https://github.com/darktable-org/darktable/issues/5149#issuecomment-640066157_

LebedevRI:
Yes, from the cursory examination this may work.
(thought we won't know until it works on one sample and we run it on all the nikon raws on RPU)

Right now it's not entirely straight-forward to do, because we don't actually record
which IFD originates from which Tiff tag.

Perhaps we shouldn't eagerly parse makernotes,
but leave that to each decoder do to if it wants to.

PR #566, opened 2023-11-23
NefDecoder: refactor linearization curve detection

kmilos:
Avoids throwing exception when only the tag 0x96 exists.

For an example see https://github.com/darktable-org/darktable/issues/15562#issuecomment-1824526754

This is again related to https://github.com/darktable-org/darktable/issues/5149

### DIFF
--- a/src/librawspeed/decoders/NefDecoder.cpp
+++ b/src/librawspeed/decoders/NefDecoder.cpp
@@ -115,7 +115,12 @@ RawImage NefDecoder::decodeRawInternal() {
 
   mRaw->dim = iPoint2D(width, height);
 
-  raw = mRootIFD->getIFDWithTag(static_cast<TiffTag>(0x8c));
+  auto ifds = mRootIFD->getIFDsWithTag(static_cast<TiffTag>(0x96));
+  if (ifds.empty()) {
+    raw = mRootIFD->getIFDWithTag(static_cast<TiffTag>(0x8c)); // Fall back
+  } else {
+    raw = ifds.front();
+  }
 
   const TiffEntry* meta;
   if (raw->hasEntry(static_cast<TiffTag>(0x96))) {
--- a/src/librawspeed/decoders/NefDecoder.cpp
+++ b/src/librawspeed/decoders/NefDecoder.cpp
@@ -115,18 +115,13 @@ RawImage NefDecoder::decodeRawInternal() {
 
   mRaw->dim = iPoint2D(width, height);
 
-  auto ifds = mRootIFD->getIFDsWithTag(static_cast<TiffTag>(0x96));
-  if (ifds.empty()) {
-    raw = mRootIFD->getIFDWithTag(static_cast<TiffTag>(0x8c)); // Fall back
-  } else {
-    raw = ifds.front();
-  }
-
-  const TiffEntry* meta;
-  if (raw->hasEntry(static_cast<TiffTag>(0x96))) {
-    meta = raw->getEntry(static_cast<TiffTag>(0x96));
-  } else {
-    meta = raw->getEntry(static_cast<TiffTag>(0x8c)); // Fall back
+  const TiffEntry* meta =
+      mRootIFD->getEntryRecursive(static_cast<TiffTag>(0x96));
+  if (!meta) {
+    meta = mRootIFD->getEntryRecursive(static_cast<TiffTag>(0x8c)); // Fall back
+    if (!meta) {
+      ThrowRDE("Missing linearization table.");
+    }
   }
 
   ByteStream rawData(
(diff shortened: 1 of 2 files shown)


src/librawspeed/decoders/NefDecoder.cpp
```diff
@@ -115,13 +115,14 @@ RawImage NefDecoder::decodeRawInternal() {
...
-
   const TiffEntry* meta;
-  if (raw->hasEntry(static_cast<TiffTag>(0x96))) {
+  try {
+    raw = mRootIFD->getIFDWithTag(static_cast<TiffTag>(0x96));
     meta = raw->getEntry(static_cast<TiffTag>(0x96));
-  } else {
-    meta = raw->getEntry(static_cast<TiffTag>(0x8c)); // Fall back
+  } catch (const TiffParserException&) {
+    // Fall back
```
LebedevRI: Please don't do this kind of stuff.
There's `hasEntry()`.

kmilos: The problem was that the IFD couldn't be found. I can't run `hasEntry()` if I don't have an IFD and the exception was thrown in `getIFDWithTag()` previously, no? Chicken & egg kind of...

Feel free to refactor in a better way, but the point is that we need to fetch an IFD with either 0x96 or 0x8c.

LebedevRI: Then see what `getIFDWithTag()` does, and call `getIFDsWithTag()` directly,
and fall back to the other tag if `getIFDsWithTag()` returned nothing.
My point: don't just catch exceptions if the code can be written to handle it more gracefully.

kmilos: Done.

kmilos: I guess one could go a step further and remove `hasEntry()` redundancy, it should be implied by `getIFDsWithTag()`/`getIFDWithTag()`...

kmilos: Or skip looking for the IFD altogether as it's not used, we just need the tag entry.

src/librawspeed/decoders/NefDecoder.cpp
```diff
@@ -115,13 +115,13 @@ RawImage NefDecoder::decodeRawInternal() {
 
   mRaw->dim = iPoint2D(width, height);
 
-  raw = mRootIFD->getIFDWithTag(static_cast<TiffTag>(0x8c));
-
-  const TiffEntry* meta;
-  if (raw->hasEntry(static_cast<TiffTag>(0x96))) {
-    meta = raw->getEntry(static_cast<TiffTag>(0x96));
```
LebedevRI, line 122: Uh, what was the old code even trying to do?
Looking for an IFD with one specific tag, but then opting for completely another tag...

kmilos: Yep, definitely not entirely correct. [The assumption was that 0x8c was always there (which is the case for OOC NEFs), and this was just a hack to get the MakerNote IFD.](https://github.com/darktable-org/rawspeed/issues/234)

kmilos:
@LebedevRI Can we please address this for the upcoming dt 4.6 please?

src/librawspeed/decoders/NefDecoder.cpp
kmilos: Turns out, we don't need the IFD at all, so maybe better to just recursively search for the desired 0x96 and 0x8c tags (`getIFDsWithTag`/`getIFDWithTag` don't do anything different to that recursive tag search anyway).

LebedevRI:
> @LebedevRI Can we please address this for the upcoming dt 4.6 please?

Reminder that i've yet again no idea what the actual schedule is.
I do not understand this secrecy at all. Why can it not be posted to discourse.

kmilos:
> Reminder that i've yet again no idea what the actual schedule is.
> I do not understand this secrecy at all. Why can it not be posted to discourse.

It's public on the wiki, and the process hasn't changed in a while: https://github.com/darktable-org/darktable/wiki/Releases-cycle

Pascal also sends out a reminder on the [mailing list](https://www.mail-archive.com/darktable-dev@lists.darktable.org/msg06828.html), don't know about other channels...

LebedevRI:
@kmilos thank you.
We'll see when this gets to dt, i kinda want to let things settle just a bit before updating `stable`.

kmilos:
Thanks. Sure, there are 3 more weeks by the looks of it.

kmilos:
@LebedevRI FYI, [another dt schedule and code freeze confirmation](https://www.mail-archive.com/darktable-dev@lists.darktable.org/msg06829.html).

LebedevRI:
> @LebedevRI FYI, [another dt schedule and code freeze confirmation](https://www.mail-archive.com/darktable-dev@lists.darktable.org/msg06829.html).

Assuming things didn't get tagged weeks in advance, propagated.

PR #578, opened 2023-12-03
CR2: read black/white levels from Makernotes

### LINKED ISSUE #102: Canon: WB correction not handled.
At least some cameras, when the "White balance correction" is enabled, bake white balance into the raw data. So far i'm aware than Canon 6D does that. In such a case, the white balance info is still present.
```
[MakerNotes]    0x040a ViewfinderWarnings              : Monochrome, WB corrected
```
(from exiftool)
So we would have to detect that "WB corrected" flag is set, and then either say that wb is `(1.0f, 1.0f, 1.0f)`, or, for consistency, un-apply WB...

Thank you, Canon!

### LINKED ISSUE #417: Get black level and white point from EXIF for Canon
Something similar to #385 should be done for Canon cameras to replace the per-model, per-ISO table in [data/cameras.xml](https://github.com/darktable-org/rawspeed/blob/develop/data/cameras.xml). https://exiftool.org/TagNames/Canon.html#ColorData4 has the relevant info (`SpecularWhiteLevel` and `AverageBlackLevel` are what we are after).

Happy to help with testing or providing data from my library of photographs, but I can’t submit a PR.

For what it’s worth, a search on “canon” currently yields 16 open issues in this project, and it looks like most if not all would be solved by implementing this.

(This is a followup from darktable-org/darktable#13320)

LebedevRI:
This handles basically everything CR2, except a few of the oldest cameras,
it is not obvious if they even store `SpecularWhiteLevel` in the `MakerNotes`.

Fixes https://github.com/darktable-org/rawspeed/issues/102
Fixes https://github.com/darktable-org/rawspeed/issues/417

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -380,7 +380,6 @@
 			<Color x="1" y="1">BLUE</Color>
 		</CFA>
 		<Crop x="22" y="18" width="4290" height="2856"/>
-		<Sensor black="1020" white="14500"/>
 		<Aliases>
 			<Alias id="Digital Rebel XSi">Canon EOS DIGITAL REBEL XSi</Alias>
 			<Alias id="Kiss Digital X2">Canon EOS Kiss Digital X2</Alias>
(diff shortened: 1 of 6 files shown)


LebedevRI:
Gone through all of the RPU regression tests once more,
all seems to be in order, aside from the `Canon EOS 6D Mark II`'s sRaw/mRaw,
where we go from final whitelevel of 16383 to ~58232. It *seems* like those images
are just -2EV underexposed, but i'm not really sure...

I'm just going to merge this now.

PR #579, opened 2023-12-05
Fujifilm FinePix SL1000 support

victoryforce:
After checking [the sample taken from the RPU](https://raw.pixls.us/getfile.php/5760/nice/Fujifilm%20-%20FinePix%20SL1000%20-%2012bit%20uncompressed%20(4:3).RAF) I settled on specifying the entire active area (which is larger than what the ADC reports) as I don't see any garbage pixels.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -14681,6 +14681,25 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Hmm yes, i don't see hints of garbage pixels either on that sample.
Does not mean there are aren't, but impossible to tell on a single sample.

LebedevRI:
@victoryforce thank you!

PR #608, opened 2024-01-18
Add support for `unknown` and `unknown-no-samples` `support` modes in `cameras.xml`

LebedevRI:
The idea is that we aren't likely to ever explicitly either support
or unsupport every single camera ever released,
and we must make that determination to enter a camera
into `cameras.xml`, yet it is generally beneficial to list a camera
in `cameras.xml` without making that determination,
for example then RPU can ask for samples for that camera.

Therefore, we need new types of `support=` values in `cameras.xml`,
ones that do not affect the support-ness, but allow to differentiate
the existence of RPU samples.

Thus, patch.

CC @kmilos

### DIFF
--- a/data/cameras.xsd
+++ b/data/cameras.xsd
@@ -514,11 +514,13 @@
   </xs:simpleType>
   <xs:simpleType name="CameraTypeSupportedType">
     <xs:restriction base="xs:string">
-      <xs:pattern value="(no|no-samples)"/>
+      <xs:pattern value="(no|no-samples|unknown|unknown-no-samples)"/>
       <xs:minLength value="2"/>
-      <xs:maxLength value="10"/>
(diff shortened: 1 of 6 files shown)


data/cameras.xsd
```diff
@@ -514,11 +514,13 @@
   </xs:simpleType>
   <xs:simpleType name="CameraTypeSupportedType">
     <xs:restriction base="xs:string">
-      <xs:pattern value="(no|no-samples)"/>
+      <xs:pattern value="(no|no-samples|unknown|unknown-no-samples)"/>
```
LebedevRI, line 517: Maybe this should be two separate fields (`unsupported-no-samples`?),
but i suppose that can be done later.

kmilos:
Thanks, will be interesting to see how this plays out.

PR #610, opened 2024-01-19
Fujifilm X-A20 support

kmilos:
Haven't actually used the new "unknown-no-samples" tag (as first planned) because [it is basically the same as the supported X-A10](https://www.photographyblog.com/reviews/fujifilm_x_a20_review).

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -15614,6 +15614,27 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="FUJIFILM" model="X-A20" supported="unknown-no-samples">
+		<ID make="Fujifilm" model="X-A20">Fujifilm X-A20</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
+		</CFA>
+		<Crop x="0" y="0" width="0" height="0"/>
+		<Sensor black="256" white="4094"/>
+		<Hints>
+			<Hint name="jpeg32_bitorder" value=""/>
+		</Hints>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">11540 -4999 -991</ColorMatrixRow>
+				<ColorMatrixRow plane="1">-2949 10963 2278</ColorMatrixRow>
+				<ColorMatrixRow plane="2">-382 1049 5605</ColorMatrixRow>
+			</ColorMatrix>
+		</ColorMatrices>
+	</Camera>
 	<Camera make="FUJIFILM" model="XQ1">
 		<ID make="Fujifilm" model="XQ1">Fujifilm XQ1</ID>
 		<CFA2 width="6" height="6">


LebedevRI:
Thank you for looking into this.

The thing is, i really don't want to add explicit support for something,
when it fails to meet the minimal requirements, namely: a sample on RPU.
On the other hand, i'd gladly take an entry similar to that of https://github.com/darktable-org/rawspeed/commit/8b50aa4fc4b2676938cf54d0c0df461129569558,
which, as discussed, may actually help us get said sample.

kmilos:
> On the other hand, i'd gladly take an entry similar to that of [8b50aa4](https://github.com/darktable-org/rawspeed/commit/8b50aa4fc4b2676938cf54d0c0df461129569558),
> which, as discussed, may actually help us get said sample.

Well, that's how I started... ;)

LebedevRI:
> > On the other hand, i'd gladly take an entry similar to that of [8b50aa4](https://github.com/darktable-org/rawspeed/commit/8b50aa4fc4b2676938cf54d0c0df461129569558),
> > which, as discussed, may actually help us get said sample.
> 
> Well, that's how I started... ;)

I'm sorry, i'm failing to comprehend the meaning.
What i'm saying is that i don't want to add new `cameras.xml` entries with
`mode` not being one of `yes`, `no`, `unknown`, `unknown-no-samples`.

kmilos:
> I'm sorry, i'm failing to comprehend the meaning.

I _was_ going to add this one as "unknown-no-samples", but then I figured it is sort of supported (not entirely unknown).

As I mentioned in the discussion way before, I don't really see any difference between "no-samples" and "unknown-no-samples" in the darktable UI, the warning message is the same in the current implementation. Perhaps it'll come in the future...

I'll change the PR anyway.

LebedevRI:
> I _was_ going to add this one as "unknown-no-samples", but then I figured it is sort of supported (not entirely unknown).

Right, i'm not arguing with that. I'm sure it works with the entry added.
My point is that i don't want to add said entry (other than `<ID>`)
with no sample :S (== no test coverage)

> As I mentioned in the discussion way before, I don't really see any difference between "no-samples" and "unknown-no-samples" in the darktable UI, the warning message is the same in the current implementation. Perhaps it'll come in the future...

Is that with submodule updated, or just edited `cameras.xml`?
With `unknown-no-samples` it will completely refuse to load in darktable.

kmilos:
> Is that with submodule updated, or just edited `cameras.xml`?

Good point, I'll retest when I get a chance.

LebedevRI:
> > Is that with submodule updated, or just edited `cameras.xml`?
> 
> Good point, I'll retest when I get a chance.

Note that i suspect you'll need to update black level copying code in `imageio_rawspeed`.

kmilos:
> Note that i suspect you'll need to update black level copying code in `imageio_rawspeed`.

Hm, it fails building as a dt subproject even before getting to `imageio_rawspeed`. It build fine standalone though?!
```
C:/msys64/home/kmilos/darktable/src/external/rawspeed/src/librawspeed/common/Cpuid.cpp:37:8: error: '__get_cpuid' was not declared in this scope
   37 |   if (!__get_cpuid(1, &eax, &ebx, &ecx, &edx))
      |        ^~~~~~~~~~~
C:/msys64/home/kmilos/darktable/src/external/rawspeed/src/librawspeed/common/Cpuid.cpp:40:16: error: 'bit_SSE2' was not declared in this scope; did you mean 'WITH_SSE2'?
   40 |   return edx & bit_SSE2;
      |                ^~~~~~~~
      |                WITH_SSE2
```

I guess I need to sync up some of the dt cmake modules as well?

LebedevRI:
Huh, that makes no sense. You have removed the build directory first, right?

kmilos:
> Huh, that makes no sense. You have removed the build directory first, right?

Yep... It's almost as if the system `<cpuid.h>` doesn't get included, i.e. `common/Cpuid.h` gets priority and is included twice?!

LebedevRI:
> Hm, it fails building as a dt subproject even before getting to `imageio_rawspeed`.

I just tried locally, and it works fine for me.

kmilos:
> I just tried locally, and it works fine for me.

Windows? Could be because it can't tel between "Cpuid.h" and <cpuid.h>? But why when it's only as subproject and not standalone?

LebedevRI:
> > I just tried locally, and it works fine for me.
> 
> Windows?
No.

> Could be because it can't tel between "Cpuid.h" and <cpuid.h>? But why when it's only as subproject and not standalone?

None of that stuff has been touched for a while now.
Does dt master compile for you as-is?

LebedevRI:
FWIW, if i rename `common/Cpuid.h` file to `common/cpuid.h` (and fix includes),
it does fail just the way you've described. So the checkout must be on a case-sensitive filesystem.
But why does it fail for you now? Sounds like you changed something on your side.

kmilos:
> Does dt master compile for you as-is?

Yep, no problem if I go back on dt master w/
```
cd src/external/rawspeed
git checkout 1e505de
cd ../../..
rm -rf build
```

LebedevRI:
Well, can you `git bisect` it then please?

kmilos:
> Well, can you git bisect it then please?

Can try when I get another chance.

kmilos:
> Well, can you git bisect it then please?

Supposedly Windows dt build didn't like c2a68bb4124265eaf3b3dbac910b9cf32213943c

`target_include_directories(rawspeed_common SYSTEM PUBLIC "${RAWSPEED_SOURCE_DIR}/src/external")` perhaps makes `common/Cpuid.h` come before `<cpuid.h>` on case-insensitive platforms?

LebedevRI:
Right. That kind of makes sense i suppose.
(I really should find energy to kill last remnant of intrinsics and remove those files.)

kmilos:
With your `imageio_rawspeed.cc` patch, this now indeed fails to load as any other unknown camera, so good to go I guess...

data/cameras.xml
```diff
@@ -15614,6 +15614,27 @@
...
+			<Hint name="jpeg32_bitorder" value=""/>
+		</Hints>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">11540 -4999 -991</ColorMatrixRow>
+				<ColorMatrixRow plane="1">-2949 10963 2278</ColorMatrixRow>
+				<ColorMatrixRow plane="2">-382 1049 5605</ColorMatrixRow>
+			</ColorMatrix>
+		</ColorMatrices>
+	</Camera>
```
LebedevRI, line 15637: To be most precise and repeat myself: `supported="unknown-no-samples"` should only include `<ID>`.
```suggestion
	<Camera make="FUJIFILM" model="X-A20" supported="unknown-no-samples">
		<ID make="Fujifilm" model="X-A20">Fujifilm X-A20</ID>
	</Camera>
```

kmilos: As it is really the same camera and same coeffs as X-A10, I'm jus gonna drop this PR then until a sample appears.

LebedevRI: But that's just it, the sample is more likely to appear with such placeholder entry in place.
I don't understand what the mis-communication here is.

kmilos: I guess I still don't understand what's the utility of all these different labels...

> But that's just it, the sample is more likely to appear

How? I guess we'd have to provide some new messages/user feedback in dt? As things stand currently, there is no difference to plain unknown/no entry...

> I don't understand what the mis-communication here is.

I also don't see what's the harm in keeping all the known to be good coeffs for this particular model only, while having it labeled as "unknown-no-samples" to solicit those samples or keep track of what's out there. In my mind it would've been just more convenient to flick the switch to support it in the future...

LebedevRI: > I guess I still don't understand what's the utility of all these different labels...
> 
> > But that's just it, the sample is more likely to appear
> 
> How? I guess we'd have to provide some new messages/user feedback in dt? As things stand currently, there is no difference to plain unknown/no entry...

https://raw.pixls.us/ "[View](https://raw.pixls.us/?missing#repo) all the cameras with no samples at all.",
which parses `cameras.xml`.

kmilos:
Note to self: reopen PR **_before_** making changes to the branch. 🤦

LebedevRI:
Yup, github is best, isn't it?

kmilos:
> Yup, github is best, isn't it?

It's not just a permission issue, I really have to create a new one?

LebedevRI:
It's not a permission issue, i regularly hit that too.

PR #617, opened 2024-01-23
Add some isolated unknown camera models

kmilos:


data/cameras.xml
```diff
@@ -12480,6 +12489,9 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="SAMSUNG" model="NX3300" supported="unknown-no-samples">
+		<ID make="Samsung" model="NX3300">Samsung NX3300</ID>
+	</Camera>
```
LebedevRI, line 12494: Just verified sample for this one.

LebedevRI:
Awesome, thank you!

data/cameras.xml
kmilos: Could be a straight copy (or even an alias?) of NX3000...

PR #618, opened 2024-01-23
Clean up some Canon aliases

kmilos:
Follow the normal scheme including "EOS", cf. https://global.canon/en/c-museum/camera.html?s=dslr


LebedevRI:
I have not verified that this is what ADC calls them, but this looks more consistent.
Thank you!

PR #619, opened 2024-01-23
Samsung NX3300 support

kmilos:


LebedevRI:
I suppose it does look right ish.

Amazing, each time i start darktable, it becomes visibly worse since the last time.

LebedevRI:
@kmilos thank you!

PR #635, opened 2024-01-31
DngDecoder: use uninitialized white point value for float DNGs

kmilos:
E.g. HDRMerge will use 65535 for real 16b cameras, don't hijack it.

This is to go w/ https://github.com/darktable-org/darktable/pull/16206

~~TODO: WhiteLevel should probably become float to support RATIONAL type and default value 1.0 (same goes for BlackLevel).~~


LebedevRI:
Yeah i don't know, seems broken either way.
Thank you.

kmilos:
> Yeah i don't know, seems broken either way.

Yep, "damned if you do, damned if you don't" sort of thing... ~~(until full support for rationals is implemented.)~~

kmilos:
Actually, after taking a look around the SDK, this should literally be "1"...

PR #640, opened 2024-02-02
oss-fuzz: OpenMP-aware fuzzing

LebedevRI:


LebedevRI:
Alright, that is good enough for now.

PR #641, opened 2024-02-02
Fujifilm X-A20 support

kmilos:
https://github.com/darktable-org/rawspeed/pull/610 redux

[It is basically the same as the supported X-A10](https://www.photographyblog.com/reviews/fujifilm_x_a20_review).


LebedevRI:
Still no samples :(

PR #647, opened 2024-02-06
Leica V-Lux 5 support

kmilos:
Confirmed equivalent to DMC-FZ1000M2


LebedevRI:
RPU sample sucks, i can not verify crop at all.

@kmilos thank you.

PR #650, opened 2024-02-08
Hasselblad CFV-50c alias cleanup

kmilos:
The `id` attribute is what gets shown in dt UI (if present and different from the actual value), we only want one model for all body combos.

Hasselblad Phocus software also normalizes to just "Hasselblad CFV-50c" when converting any of these from 3FR to FFF.


LebedevRI:
I guess this makes sense. This is so horrible :)

@kmilos thank you!

PR #651, opened 2024-02-08
Pentax Q-S1 normalization

kmilos:
We could drop "no-samples" on both this one and the Q10, as this is only about naming scheme for regular DNGs...

Alternatively, if we really want to solicit even more OOC DNG samples, we might want to add "no-samples" to the Leicas added recently?


LebedevRI:
> We could drop "no-samples" on both this one and the Q10, as this is only about naming scheme for regular DNGs...

Yes, please.
It was an error to put `(unknown-)?no-samples` on DNG-only cameras.

LebedevRI:
@kmilos thank you!

PR #663, opened 2024-02-13
`compiler-versions.cmake`: 2024 edition

LebedevRI:
LLVM16 migrated into debian stable.
macOS 12 will become EOL in summer.

Therefore, for the summer dt release,
we can require macOS 13 + XCode 15.2,
which means we'll be able to bump required LLVM version up to 15 or even 16.

@TurboGit @zisoft FYI ^
I'm really looking forward dropping at least the LLVM14 support.

PR #666, opened 2024-02-15
Require XCode 15.2 / libc++-16 / clang-15

LebedevRI:
See https://github.com/darktable-org/rawspeed/pull/663

> LLVM16 migrated into debian stable. macOS 12 will become EOL in summer.
> 
> Therefore, for the summer dt release, we can require macOS 13 + XCode 15.2, which means we'll be able to bump required LLVM version up to 15 or even 16.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -56,7 +56,6 @@ jobs:
           - { distro: "debian:bookworm-slim", family: GNU,  version: 12, CC: gcc-12,   CXX: g++-12 }
           - { distro: "debian:bookworm-slim", family: LLVM, version: 16, CC: clang-16, CXX: clang++-16 }
           - { distro: "debian:bookworm-slim", family: LLVM, version: 15, CC: clang-15, CXX: clang++-15 }
-          - { distro: "debian:bookworm-slim", family: LLVM, version: 14, CC: clang-14, CXX: clang++-14 }
         flavor: [ ReleaseWithAsserts, Release ]
     uses: ./.github/workflows/CI-linux.yml
     with:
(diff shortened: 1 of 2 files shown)


kmilos:
I would be afraid to merge this one! 😨 

And not passing at that... should be telling us something!

LebedevRI:
:imp: :imp: :imp:

LebedevRI:
(this is waiting at least on the dt 4.6.1 release)

LebedevRI:
Ha!

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.22) # FIXME: 3.24.2 / 3.25
```
kmilos: You probably checked already, but > 3.22 would exclude Ubuntu 22.04 (unless they backport), which is still supported for a while...

LebedevRI: That's why i've lowered it from `3.24.2` to `3.22`...
https://build.opensuse.org/public/build/graphics:darktable:master/xUbuntu_22.04/x86_64/darktable/_log says:
`[   49s] [622/766] installing cmake-3.22.1-1ubuntu1`
https://packages.ubuntu.com/jammy/cmake also says `3.22`.

Besides, as per `cmake/compiler-versions.cmake` wording,
we _could_ just drop support for `Ubuntu 22.04` right away.
But that does not really help us since that would only help with CMake version.

kmilos: > we _could_ just drop support for `Ubuntu 22.04` right away

Don't think that'll be a realistic option for the summer cycle... IIRC the last time around it took several months after the release for GitHub runners to become available...

LebedevRI: I don't quite follow, why on earth would it ever matter what the _default_ is on github runners?
The only thing the github linux runners should be used for, is to run docker on them,
and from there on you are free to pick your baseline image.
See https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container
That's what i'd always done in this repo. Is that not what happens for darktable? o_O

kmilos: Note that I said _availability_, without mentioning any default.

And no, not in a container, dt uses straight runners.

LebedevRI: > And no, not in a container, dt uses straight runners.

Well, "you" might want to get that fixed then :)

LebedevRI:
dt 4.6.1 has been released, and oss-fuzz appears happy with #667.
Let's merge this into darktable and see what rest of CI says.

PR #677, opened 2024-02-21
OM System OM-1 Mark II support (12-bit only)

kmilos:
https://github.com/darktable-org/darktable/issues/16297

Using latest ADC 16.2

Crops are not verified, waiting for RPU samples.


LebedevRI:
Checked, still no sample uploads.

LebedevRI:
@kmilos thank you!

PR #680, opened 2024-02-22
Add placeholders for known unsupported cameras

kmilos:
* Canon models using CR3 codec
* Panasonic models using v8 codec

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -10844,6 +10844,12 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="Panasonic" model="DC-S5M2" supported="no">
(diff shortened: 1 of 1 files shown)


LebedevRI:
(I have not verified that each change here does exactly what it says it does.)

Hm i don't know. I guess this makes sense, and seeds the names for RPU sample request thing.

LebedevRI:
@kmilos thank you.

kmilos:
> I guess this makes sense, and seeds the names for RPU sample request thing.

One of the ideas behind it indeed. Another I had in mind is to still have `getCamera()` return at least something (e.g. sanitized maker/model names and aliases) for as many known cameras as possible.

PR #690, opened 2024-03-01
RafDecoder: support vendor crop from metadata

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/5729

Needs some testing though...

It could make sense to remove the explicit crop from database for [Gen 4](https://en.wikipedia.org/wiki/Fujifilm_X-Trans_sensor#List_of_X-Trans_sensors) (first w/ this sports finder mode?) and onwards, maybe GFX series as well?


LebedevRI:
Thanks for looking into it!

What i think needs to happen, is we need to actually store this vendor crop field in `RawImage`,
just like we do `colormatrix`, and then `rstest`/`rs-identify` need to be updated to dump that field.
Then this effectively has test coverage. Do you want to look into that?
If yes, let's do that as a separate PR, since we already read vendor crop in some other decoder.

kmilos:
I could look into it, but might be a while until I get around to it...

LebedevRI:
I'm quite conflicted here.
On one hand, this breaks a very basic promise of providing all usable pixels.
On the other hand, it's obvious that such hardcoding was never going to work
in the long run, precisely because of these raw crops.
On another hand, i don't like `cameras.xml` in the first place, and this reduces that file.

kmilos:
Damned if you do and damned if you don't. 😕

LebedevRI:
I'll merge this, but i think we need to come to a more general decision on this kind of thing,
do we just kill this concept of custom crop when there's vendor crop avaliable?

LebedevRI:
@kmilos thank you!

kmilos:
> do we just kill this concept of custom crop when there's vendor crop avaliable?

That is one consistent option.

The other is more messy I think - somehow detecting and introducing more modes to cameras.xml

LebedevRI:
> The other is more messy I think - somehow detecting and introducing more modes to cameras.xml

Yup, that is the other option i hadn't mentioned, and indeed it's messy.

Issue #520, opened 2023-09-18
Probable overscanning of sony ILCE-7M4 APS-C cropped RAWs


koolfy:
Hello, I have noticed 100% of my APS-C cropped images getting out of darktable show lines of corrupted pixels on the right of the images (before rotation corrections)

I have found this interesting thread: https://www.dpreview.com/forums/thread/4637906
This seems to suggest that at least some generations of proprietary tools have struggled in a very similar fashion with this camera body, specifically with APS-C crops.

Visually it looks like we are simply overscaning the image beyond where there is actually information.

What's interesting is in any case we read more information than Sony presents in its in-body JPGs

Here is the in-body JPG for an uncropped picture (I disabled in-body lens distortion correction):

https://ibb.co/qxz5SvP

Same picture, out of darktable 4.4.2 built against rawspeed commit 2f1e3143875b7834322daa0b7bdbd5aec0ded932 (17 sept develop HEAD) (I went to the original step in the history stack to remove as much processing as possible)

https://ibb.co/128p7M5

Now the cropped ones:

in-body JPG for APS-C cropped image:

https://ibb.co/LQ9jtxL

And the exact same picture straight out of darktable (same build):

https://ibb.co/BgMpcYt

In all cases, we read much more data around the edges than Sony does, and in the APS-C crop format, we appear to be actively overscanning the right bound on the image (but are fine vertically)

One interesting note, using digikam quickly (libraw), I noticed they suffer from a similar overscan on both the cropped and uncropped file (rawspeed reads correctly uncropped files but libraw overscans them).

I was going to compare pixel counts on the in-body camera and the raw file exports, but I don't trust how different settings in the body might affect pixel counts on the in-body JPGs.

Here is the link to a google drive containing all versions including actual raw .ARW files:

https://drive.google.com/drive/folders/1sSbFRuQuxJi14ZYcVAzrMeM2At-hOtFV?usp=sharing

One even more interesting note is that the google drive preview for APS-C crops seems to suffer from the exact same symptom.
Maybe my camera is causing this at a very low level? But it would be extremely strange that the dpreview thread would outline the exact same issue.

If it's not an incorrect RAW decoding, might it be a commonplace firmware bug that nobody noticed before?

UPDATE: in case it ends up being relevant, I'm using firmware 2.00 from Sony

LebedevRI:
Yup, not really surprised that this happens. 
Given that the crop information is hardcoded
(generally, as offsets to the image size) in an XML,
[...]

PR #699, opened 2024-03-26
Sony ILCE-9M3 support

### LINKED ISSUE #16523: Sony Alpha 9 III (ILCE-9M3)
I have recently acquired a Sony Alpha 9M3, which appears to be at least partially supported by RawTherapee (uncompressed mode) but not by DarkTable yet. It would be good enough if I could read the same format for a start.

I have submitted a set of 5 ARWs  (one per compression mode) to raw.pixls.us.

kmilos:
Resolves https://github.com/darktable-org/darktable/issues/16523

Luckily the APS-C crop is smaller than the FF one (unlike w/ the [7M4](https://github.com/darktable-org/rawspeed/issues/520)), so this is good to go IMHO. Otherwise, we switch all "modern" Sonys to vendor metadata crop...

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -14127,8 +14127,23 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="SONY" model="ILCE-9M3" supported="unknown-no-samples">
+	<Camera make="SONY" model="ILCE-9M3">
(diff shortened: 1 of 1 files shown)


LebedevRI:
X-Ref: https://github.com/pixlsus/raw/issues/83

LebedevRI:
@kmilos thank you!
Checked on the samples, looks about right.

PR #707, opened 2024-04-06
Apple MakerNotes parsing support

kmilos:
Saves significant load time by skipping the erroneous scanning of 16752 ("Ap" = 0x4170) entries otherwise.

WIP: there is a 0x2e entry in the Apple makernotes that now clashes w/ the (Panasonic?) MAKERNOTE_ALT. Don't seem to have access to `make` at that point, so not sure how to best handle this? @LebedevRI 

It might in general be a good idea to add a mechanism to bail out of trying to parse some known unsupported makernotes earlier? Another example: Google's are not in the IFD format, but we try parsing 17480 ("DH" = 0x4448) entries anyway.

### DIFF
--- a/src/librawspeed/tiff/TiffIFD.cpp
+++ b/src/librawspeed/tiff/TiffIFD.cpp
@@ -80,7 +80,7 @@ void TiffIFD::parseIFDEntry(NORangesSet<Buffer>* ifds, ByteStream& bs) {
       break;
 
     case MAKERNOTE:
-    case MAKERNOTE_ALT:
+    // case MAKERNOTE_ALT:
       add(parseMakerNote(ifds, t.get()));
       break;
 
@@ -194,6 +194,8 @@ TiffRootIFDOwner TiffIFD::parseMakerNote(NORangesSet<Buffer>* ifds,
     setup(true, 16);
   } else if (bs.hasPrefix("EPSON")) {
     setup(false, 8);
+  } else if (bs.hasPrefix("Apple iOS")) {
+    setup(true, 14, 12, "Apple makernote");
   } else if (bs.hasPatternAt("Exif", 6)) {
     // TODO: for none of the rawsamples.ch files from Panasonic is this true,
     // instead their MakerNote start with "Panasonic" Panasonic has the word


LebedevRI:
I think i'm missing a bit more trivia here in order to comment on this.

* Why do we need to do this?
* What goes wrong if we don't do this?
* What do we achieve by doing this?

> WIP: there is a 0x2e entry in the Apple makernotes that now clashes w/ the (Panasonic?) MAKERNOTE_ALT. Don't seem to have access to `make` at that point, so not sure how to best handle this? @LebedevRI

Clashes in what way?
 
> It might in general be a good idea to add a mechanism to bail out of trying to parse some known unsupported makernotes earlier? Another example: Google's are not in the IFD format, but we try parsing 17480 ("DH" = 0x4448) entries anyway.

Maybe, i don't know. I do suspect that the parsing should be less up-front but more on the lazy, as-needed side.

kmilos:
Nothing goes wrong, as in it doesn't break functionality. But there's a lot of cycles wasted and eventually an exception thrown for random OOB access.

LebedevRI:
Ok, and what about the clashing part?

kmilos:
> what about the clashing part?

Apple makernotes actaully have a regular (unknown to us) 0x2e entry, but RawSpeed thinks that's a Panasonic MakerNote IFD, so it ends up calling `TiffIFD::parseMakerNote()` recursively (which actually segfaults).

LebedevRI:
By `segfaults`, i'm guessing you mean "throws an exception"?
I see no segfaults here, and i'm reasonably sure there shouldn't be any.

@kmilos this sounds reasonable, although the comments here are a bit cryptic.
Thank you.

kmilos:
> By `segfaults`, i'm guessing you mean "throws an exception"?

Hm... I thought I saw something different in the debugger, but an exception it is, so I guess no need for additional workarounds.

michelerenzullo:
Hi @kmilos, thank you for this project, I am not sure if is something off on my side or something is going on when trying to parsing the IFD on an Apple iOS DNG.
I found this closed MR, and I believe is highly related to my issue.
I got this exception when trying to decoding an Apple ProRAW image.
```
ThrowException<rawspeed::TiffParserException>(char const*, ...)
TiffEntry::TiffEntry(rawspeed::TiffIFD*, rawspeed::ByteStream&)
TiffIFD::parseIFDEntry(rawspeed::NORangesSet<rawspeed::Buffer>*, rawspeed::ByteStream&)
```
I am trying to debugging and figure out what's going on, in the meantime this is the image to reproduce the issue:
[_iPhone14Pro_ProRAW48MP.DNG](https://drive.proton.me/urls/6TREFGABQG#TW69qwt51jly)

Thanks for helping

LebedevRI:
//Which// exception?
https://github.com/darktable-org/rawspeed/blob/b646af081183d97ae3b9ea421d2b4afa9961fe76/src/librawspeed/tiff/TiffEntry.cpp#L106C43-L113
Also, more globally, what is the bigger overall problem that you are having with that file?

michelerenzullo:
> //Which// exception? https://github.com/darktable-org/rawspeed/blob/b646af081183d97ae3b9ea421d2b4afa9961fe76/src/librawspeed/tiff/TiffEntry.cpp#L106C43-L113 Also, more globally, what is the bigger overall problem that you are having with that file?

This one to be more precise "Error reading TIFF structure. Unknown Type 0x%x encountered."
```cpp
TiffEntry::TiffEntry(TiffIFD* parent_, ByteStream& bs)
    : parent(parent_), tag(static_cast<TiffTag>(bs.getU16())) {
  const uint16_t numType = bs.getU16();
  if (numType > static_cast<uint16_t>(TiffDataType::OFFSET))
    ThrowTPE("Error reading TIFF structure. Unknown Type 0x%x encountered.",
```

From debugger the locals says
`tag = PANASONIC_ISO_SPEED, numType = 16`

The exception make crash the implementation I have in my library. I could workaround probably, but I wanted to report so maybe can be fixed

LebedevRI:
I'm not sure what you mean by "crash", it decodes "successfully":
```
rawspeed/build-Clang18-SANITIZE$ ./src/utilities/rstest/rstest -c -f -d /tmp/_iPhone14Pro_ProRAW48MP.DNG 
/tmp/_iPhone14Pro_ProRAW48MP.DNG                       : starting decoding ... 
/tmp/_iPhone14Pro_ProRAW48MP.DNG failed: /home/lebedevri/rawspeed/src/librawspeed/decompressors/AbstractDngDecompressor.cpp:249: void rawspeed::AbstractDngDecompressor::decompress() const: Too many errors encountered. Giving up. First Error:
/home/lebedevri/rawspeed/src/librawspeed/decompressors/LJpegDecoder.cpp:108: virtual Buffer::size_type rawspeed::LJpegDecoder::decodeScan(): Unsupported predictor mode: 7
Total decoding time: 0s
```

kmilos:
The lack of ljpeg predictor 7 support for pixel decoding is a separate, bigger (and known) issue here, to the unsupported parsing of Apple MakerNotes.

In any case, you should probable be handling the exceptions in your app.

michelerenzullo:
Thank you, I got later that even with the correct tiff parser I wouldn't be able because the predictor 7 is in progress. Yes, I will handle this more properly on my side.

PR #713, opened 2024-04-22
Remove "no-samples" for GoPro FUSION

kmilos:
Samples now provided in https://github.com/darktable-org/darktable/issues/16646


kmilos:
In fact, we could potentially remove all GoPro entries for a slimmer `cameras.xml`? They don't really do anything (like make/model normalization), and these are just like ordinary/generic DNGs... E.g., we didn't bother adding new ones after HERO8 anyway.

LebedevRI:
@kmilos thank you!

Yeah, i do not know. We could, but then it's not like it represents a significant chunk of the file.

kmilos:
Yep, just a thought... Never mind since this is now merged.

Issue #717, opened 2024-05-13
Bad columns in left optical black area Canon EOS 2000D


PeterWem:
A darktable user had an issue with greenish raw files due to wrong black level. The culprit was 4 bad columns in the left optical black area. 
![black](https://github.com/darktable-org/rawspeed/assets/34200649/f108d5f8-accf-4458-8f7c-c92c90dcfc1e)
Changing from 0 to 5 will solve the issue.
![Skärmbild från 2024-05-13 22-42-35](https://github.com/darktable-org/rawspeed/assets/34200649/ba03eb7e-8314-4f81-b8eb-ff9c08f7a4c1)

I also downloaded a Canon EOS 2000D raw file from raw.pixls.us. All the columns there are fine so this may only be an issue with a few batches.

Issue #389, opened 2022-10-21
`RawImageDataU16::calculateBlackAreas()` only looks at a single pixel o_O


LebedevRI:
https://github.com/darktable-org/rawspeed/blob/990cec748fbb98ef5d5f9347fde013ce34dff2b8/src/librawspeed/common/RawImageDataU16.cpp#L76-L82
https://github.com/darktable-org/rawspeed/blob/990cec748fbb98ef5d5f9347fde013ce34dff2b8/src/librawspeed/common/RawImageDataU16.cpp#L93-L99
^ nowhere in these loop do we advance the `pixel` pointer...

PR #718, opened 2024-05-13
Updated EOS 2000D BlackAreas cameras.xml

PeterWem:
This change solves https://github.com/darktable-org/rawspeed/issues/717 and is cropping out the first 4 columns from left optical black area when the first 4 columns are not black.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -1388,7 +1388,7 @@
 		</CFA>
 		<Crop x="78" y="42" width="-4" height="-6"/>
 		<BlackAreas>
-			<Vertical x="0" width="70"/>
+			<Vertical x="5" width="70"/>
 			<Horizontal y="4" height="30"/>
 		</BlackAreas>
 		<Aliases>


LebedevRI:
for RPU sample, exiftool says:
```
[MakerNotes]    SensorWidth                     : 6096
[MakerNotes]    SensorHeight                    : 4051
[MakerNotes]    SensorLeftBorder                : 88
[MakerNotes]    SensorTopBorder                 : 47
[MakerNotes]    SensorRightBorder               : 6087
[MakerNotes]    SensorBottomBorder              : 4046
[MakerNotes]    BlackMaskLeftBorder             : 0
[MakerNotes]    BlackMaskTopBorder              : 0
[MakerNotes]    BlackMaskRightBorder            : 0
[MakerNotes]    BlackMaskBottomBorder           : 0
```
Is the output identical for that raw too?

https://github.com/darktable-org/rawspeed/issues/389 is the main culprit here, a few bad rows/cols shouldn't normally cause so dramatic effect.

PeterWem:
Same.
[IMG_2637_CR2.txt](https://github.com/darktable-org/rawspeed/files/15304749/IMG_2637_CR2.txt)

The sample from RPU says firmware 1.0.0. The one I got has 1.1.0

To me it seems like Canon didn't mask those 4 columns at all with this camera. Never seen that before with Canon.

Skärmbild från 2024-05-14 09-46-15

LebedevRI:
Huh, that's bizarre. And the horizontal black area in `cameras.xml` also (already) does not match the makernotes...

PeterWem:
> Huh, that's bizarre. And the horizontal black area in `cameras.xml` also (already) does not match the makernotes...

<Horizontal y="4" height="30"/>
That should be good, shouldn't it? 4 lines down to crop out nonsens pixels and ending 30 lines down instead of 36 not getting too close to the active image pixels.


Or do you mean the part that is 0?
BlackMaskLeftBorder          
BlackMaskTopBorder

PeterWem:
With permission to share this raw file here for everyone, https://drive.google.com/file/d/1fAzOCo2dGLM6yJfyJB-yG34KyFTRmfgg/view?usp=drive_link

I asked the photographer about high ISO shot because the different parts of the optical black area are more visible. 

The red part is the missing masked pixels.

I don't know what the blue part is but many raw files from Canon look like that. Is it how things work in sensor when the image first is captured? It begins to read the sensor from that part?

Purple part is the masked pixels we use to set the black level.

What about the green part? I see that part at higher ISO from different Canon models.

ob

kmilos:
> To me it seems like Canon didn't mask those 4 columns at all with this camera. Never seen that before with Canon.

Are these maybe similar in value to the pixels at the right border? Could be a timing issue w/ the sensor readout, and the last 4 pixels on the right actually end up as the first 4 pixels on the next row (and the whole image ends up shifted by 4)...

In any case, the x value can be 4, as it is zero based.

LebedevRI:
> In any case, the x value can be 4, as it is zero based.

Please change the pos to be a multiple of 2 (in either direction), i'll merge this then i guess.

PR #721, opened 2024-05-21
Fujifilm GFX100S II support

kmilos:
Confirmed color matrix is identical to GFX100 II using ADC 16.3

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -15421,8 +15421,39 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Thank you. I've checked, and there's no samples [in the review queue too] on RPU for these two cameras :/

kmilos:
> Thank you. I've checked, and there's no samples [in the review queue too] on RPU for these two cameras :/

They've just been announced last week, so no wonder 😉 Doubt we'll get some [before the dt 4.8 freeze](https://www.mail-archive.com/darktable-dev@lists.darktable.org/msg06870.html)...

kmilos:
Closing in favor of #800 where samples are located.

PR #722, opened 2024-05-21
Ricoh GR III HDF and GR IIIx HDF normalization

kmilos:
Or simply alias instead? https://www.dpreview.com/news/0569067557/ricoh-gr-iii-gr-iix-hdr-highlight-diffusion-filter


kmilos:
@LebedevRI What about converting to alias instead in order to benefit from existing noise profiles for example?

PR #723, opened 2024-05-21
Prefer aliases for Ricoh GR III HDF and GR IIIx HDF

kmilos:
Enables leveraging of existing noise profiles, lens calibration, etc.

Might be more practical than simply https://github.com/darktable-org/rawspeed/pull/722


LebedevRI:
Ok, i guess that makes sense for that model, based on light googling.

@kmilos thank you.

kmilos:
Indeed. Thanks!

PR #724, opened 2024-05-22
Correct Phase One P25+ white level

kmilos:
Following up from https://github.com/darktable-org/darktable/issues/16539#issuecomment-2041143375

Use 14b as with other Phase One models.


LebedevRI:
@kmilos thank you!

PR #726, opened 2024-05-24
Samsung EX2F: correct white point

ge0rg:
The old white-point of 4095 yields pink spots in the sky in bright pictures (sample submitted to RPU). This PR reduces the value to 4000, similar to the NX mini.

Slightly higher values may work as well, I don't know how to correctly calibrate that value from an over-exposed image.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -12398,7 +12398,7 @@
 			<Color x="1" y="1">BLUE</Color>
 		</CFA>
 		<Crop x="22" y="12" width="-148" height="-18"/>
-		<Sensor black="0" white="4095"/>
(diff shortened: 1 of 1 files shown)


LebedevRI:
@ge0rg i have not verified this, i'm going to assume that you as an affected user checked that this makes sense.

LebedevRI:
@ge0rg thank you!

ge0rg:
Thanks very much. I'm trying hard to qualify as the Samsung-camera-weirdo, and I've done only a bit of RAW shooting with the EX2F so far, and 4000 looks like it works for all shots. I'll post more PRs / issues if I encounter more issues :)

PR #731, opened 2024-06-13
Remove "no-samples" for Fujifilm FinePix S9600fd

victoryforce:
S9600fd is just S9600 with face detection (and for S9600 we have a sample in RPU). In the Fujifilm model line, there were several cases of models being released without and with the `fd` suffix. Such models are identical in everything else except for the presence of face detection code in the camera firmware.

So we can avoid annoying the user with "no-samples" warnings for S9600fd.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -14995,25 +14995,9 @@
 		</CFA2>
 		<Crop x="32" y="0" width="-32" height="0"/>
 		<Sensor black="0" white="15872"/>
-		<Hints>
-			<Hint name="fuji_rotate" value=""/>
-		</Hints>
-		<ColorMatrices>
-			<ColorMatrix planes="3">
-				<ColorMatrixRow plane="0">12343 -4515 -1285</ColorMatrixRow>
-				<ColorMatrixRow plane="1">-7165 14899 2435</ColorMatrixRow>
-				<ColorMatrixRow plane="2">-1895 2496 8800</ColorMatrixRow>
-			</ColorMatrix>
-		</ColorMatrices>
-	</Camera>
-	<Camera make="FUJIFILM" model="FinePix S9600fd" supported="no-samples">
-		<ID make="Fujifilm" model="FinePix S9600fd">Fujifilm FinePix S9600fd</ID>
-		<CFA2 width="2" height="2">
-			<ColorRow y="0">GB</ColorRow>
-			<ColorRow y="1">RG</ColorRow>
-		</CFA2>
-		<Crop x="64" y="0" width="-64" height="0"/>
-		<Sensor black="0" white="15872"/>
+		<Aliases>
+			<Alias id="FinePix S9600fd">Fujifilm FinePix S9600fd</Alias>
+		</Aliases>
 		<Hints>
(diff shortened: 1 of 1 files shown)


LebedevRI (changes requested):
First, if that is a standalone camera entry, it must have samples.

Second, clearly we aren't bothering them enough since there's still no RPU sample for S9600fd in review queue. /s

Thirdly, if this is the exact same camera, it's whole `<Camera>` should be 
```
		<Aliases>
			<Alias id="FinePix S9600fd">Fujifilm FinePix S9600fd</Alias>
		</Aliases>
```
in `<Camera>` for `S9600`. I think that will also silence that begging too.

victoryforce:
> First, if that is a standalone camera entry, it must have samples.
> 
> Second, clearly we aren't bothering them enough since there's still no RPU sample for S9600fd in review queue. /s

This camera is almost two decades old, so we shouldn't expect many darktable users to process images from this camera. And this makes it very unlikely that we will ever see samples from this particular camera model. :(


> Thirdly, if this is the exact same camera, it's whole `<Camera>` should be
> 
> ```
> 		<Aliases>
> 			<Alias id="FinePix S9600fd">Fujifilm FinePix S9600fd</Alias>
> 		</Aliases>
> ```
> 
> in `<Camera>` for `S9600`. I think that will also silence that begging too.

I agree. The xml entries for these two cameras are identical, so the replacement to Alias was a natural choice. I simply chose the fix that made the least change to the current data. Changed.

LebedevRI:
@victoryforce thank you!

data/cameras.xml
```diff
@@ -14995,25 +14995,9 @@
...
-	<Camera make="FUJIFILM" model="FinePix S9600fd" supported="no-samples">
-		<ID make="Fujifilm" model="FinePix S9600fd">Fujifilm FinePix S9600fd</ID>
-		<CFA2 width="2" height="2">
-			<ColorRow y="0">GB</ColorRow>
-			<ColorRow y="1">RG</ColorRow>
-		</CFA2>
-		<Crop x="64" y="0" width="-64" height="0"/>
-		<Sensor black="0" white="15872"/>
+		<Aliases>
+			<Alias id="FinePix S9600fd">Fujifilm FinePix S9600fd</Alias>
```
kmilos, line 14999: This is wrong and won't work - should just say `<Alias>FinePix S9600fd</Alias>` (i.e., [the actual value and the `id` attribute should be identical in this case, so the attribute can be omitted](https://github.com/darktable-org/rawspeed/tree/develop/data#camera-model-aliases)).

Please fix before merging to stable and uploading to dt submodule. @victoryforce @LebedevRI

victoryforce: Oh, drat! I just copy-pasted the suggested string, although I could have noticed that a little higher in the file there is Alias ​​written differently... :( Thanks for noticing!

LebedevRI: Yeah, as i've said previously, i don't fully understand/remember this `<Alias>` syntax.
I thought that is what the right line is, based on what happens for Canon entries.
@kmilos thanks!

PR #732, opened 2024-06-14
Fix alias for Fujifilm FinePix S9600fd

victoryforce:
Fix for correct syntax of Alias, thanks to @kmilos.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -14996,7 +14996,7 @@
 		<Crop x="32" y="0" width="-32" height="0"/>
 		<Sensor black="0" white="15872"/>
 		<Aliases>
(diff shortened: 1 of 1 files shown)


kmilos:
To clarify: you use both real value and clean `id` attribute when you want to normalize/prettify the actual string from the Exif Model tag, like we usually do for Canons to avoid the duplicate "Canon Canon" when concatenating Make + Model: https://github.com/darktable-org/rawspeed/blob/8d343c0eaa820e0f69e2f8a36ac8744807665ad1/data/cameras.xml#L110-L113

Since the Exif Model string is already "clean" for Fujifilms, there is no need for both.

PR #734, opened 2024-06-25
Do ask for recent Panasonic samples

kmilos:
Turns out I was wrong - we do want to ask for samples even if there is no known codec (fail is silent otherwise).


LebedevRI:
Right, i missed that.
@kmilos thank you!

PR #737, opened 2024-06-27
Leica D-Lux 8 normalization

kmilos:
Looks like from the D-Lux 8 Leica is switching this series to DNG output option only: https://www.dpreview.com/news/3471020867/leica-continues-compacts-with-d-lux-8-featuring-four-thirds-type-sensor


LebedevRI:
Nice.
@kmilos thank you!

PR #738, opened 2024-06-27
CI: slim down MSYS2 install

kmilos:
Also switch coverage to now recommended/default UCRT64 environment

### DIFF
--- a/.github/workflows/CI-windows-msys2.yml
+++ b/.github/workflows/CI-windows-msys2.yml
@@ -51,9 +51,6 @@ jobs:
           cache: false
           msystem: ${{ inputs.msys2-msystem }}
           update: false
-          install: >-
-            git
-            base-devel
           pacboy: >-
             cc:p
             gcc-libs:p
@@ -67,12 +64,8 @@ jobs:
       - name: Install Additional Dependencies (Coverage)
         if: inputs.flavor == 'Coverage'
         timeout-minutes: 1
-        uses: msys2/setup-msys2@v2
-        with:
-          cache: false
-          msystem: ${{ inputs.msys2-msystem }}
-          pacboy: >-
-            zstd:p
+        run: |
+          pacboy --noconfirm -S --needed zstd:p
       - name: Configure
         timeout-minutes: 3
         env:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -178,7 +178,7 @@ jobs:
       matrix:
         include:
           - os: windows-latest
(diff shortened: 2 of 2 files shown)


kmilos:
The missing setuptools issue seems to be inside the CodeChecker setup code, `python3-setuptools-whl` is already pulled in previously...

.github/workflows/CI-windows-msys2.yml
```diff
@@ -71,6 +68,7 @@ jobs:
         with:
           cache: false
           msystem: ${{ inputs.msys2-msystem }}
+          update: false
```
kmilos: In fact, this whole step could potentially be dropped as `zstd` is already installed previously, as a dependency of `cc`...

kmilos: Or at least converted to just the `pacboy ...` command instead of going through the entire `setup-msys2` action again...

LebedevRI: It is not a great idea to rely on known dependencies
to be happenstancely being installed due to them also being
transitively dependent by something else.

kmilos: Ok, then I'll change it to just the package install instead of setting up MSYS2 from scratch again...

LebedevRI:
> The missing setuptools issue seems to be inside the CodeChecker setup code, `python3-setuptools-whl` is already pulled in previously...

Yeah, i don't understand what is going on there.

.github/workflows/CI-windows-msys2.yml
kmilos: This is now good to go, should save some trees hopefully. 🌍

.github/workflows/CI-windows-msys2.yml
```diff
@@ -51,9 +51,6 @@ jobs:
           cache: false
           msystem: ${{ inputs.msys2-msystem }}
           update: false
```
LebedevRI, line 53: As per https://github.com/msys2/setup-msys2?tab=readme-ov-file#update,
that means we don't update the package index either? That seems unfortunate.

LebedevRI:
@kmilos thank you.
Does not seem to break CI further than it already is, so i guess this is fine.

.github/workflows/CI-windows-msys2.yml
kmilos: Note that this was off before my change.

Not really an issue IMHO as you're not interested in testing the latest MSYS2 packages but your code. The [msys2-installer and the setup-msys2 action](https://github.com/msys2/setup-msys2/commit/ff00a68cbc306bd3e1a99ead92b9a773b8a3b951) get an update every 3 months or so anyway.

PR #739, opened 2024-06-30
Nikon Z6_3 support

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/17075 using ADC 16.4.

Covers traditional lossless only, obviously.


ertwroc:
@LebedevRI Would it be possible to merge this PR?

PR #741, opened 2024-07-11
Add Konica Minolta Alpha Sweet Digital alias

kmilos:
See https://en.wikipedia.org/wiki/Konica_Minolta_Maxxum_5D and https://discuss.pixls.us/t/cannot-load-raw-from-minolta-dynax-maxxum-5d/44314


LebedevRI:
@kmilos thank you.

kmilos:
@LebedevRI Looks like this sample could also be potentially salvaged, seems [it was uploaded just before the crash](https://discuss.pixls.us/t/cannot-load-raw-from-minolta-dynax-maxxum-5d/44314/24?u=kmilos)?

LebedevRI:
Ack

PR #742, opened 2024-07-17
Add Canon EOS R1 placeholder

kmilos:
Btw, R5 Mark II was also released simultaneously, but won't have access to my machine for a while to check if the real Exif model string is "Canon EOS R5m2" or something else...


LebedevRI:
@kmilos thank you.

PR #744, opened 2024-08-07
Samsung WB2000: correct white point

ge0rg:
The old white-point of 4095 yields pink spots in the sky in bright pictures. This reduces the value to 4000, similarly to the NX mini and the EX2F (537953a71a9284353ae8830e24a4f01b4cdb453f).

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -12847,7 +12847,7 @@
 			<Color x="1" y="1">BLUE</Color>
 		</CFA>
 		<Crop x="0" y="2" width="-10" height="-2"/>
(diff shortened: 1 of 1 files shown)


LebedevRI:
I have not verified this, i trust that you did need this and tested on your own raws :)

LebedevRI:
@ge0rg thank you!

ge0rg:
Thank you too! Yeah, I just stumbled over it when editing the photo for https://photog.social/@ge0rg/112921408734423747 which had pink sky, and so applied the correction. I'll double-check the other Samsungs that I have to see if they also need to tune the white level.

PR #747, opened 2024-09-03
Add Panasonic DC-FZ80D/FZ82D/FZ85D aliases

kmilos:
Seems to be display & USB port refresh like with other recent "D" models:
https://www.dpreview.com/news/0698534590/panasonic-lumix-fz80d-brings-usb-c-and-new-displays-to-60x-superzoom
https://petapixel.com/2024/07/02/panasonics-new-lumix-fz80d-is-a-480-60x-superzoom-camera/

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -11311,6 +11311,9 @@
 			<Alias>DC-FZ80</Alias>
 			<Alias>DMC-FZ80</Alias>
 			<Alias>DMC-FZ85</Alias>
+			<Alias>DC-FZ80D</Alias>
+			<Alias>DC-FZ82D</Alias>
+			<Alias>DC-FZ85D</Alias>
 		</Aliases>
 		<ColorMatrices>
 			<ColorMatrix planes="3">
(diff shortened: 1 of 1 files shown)


LebedevRI:
@kmilos thank you!

kmilos:
@LebedevRI It also looks like there are some confusing legacy model designations in there as well, like "DMC-FZ80" and "DMC-FZ85". Do you think we should clean those up? I can't verify w/ what's on RPU, but the DMC variants don't seem to appear anywhere in reality, and [Adobe also only uses DC-FZ8x](https://helpx.adobe.com/camera-raw/kb/camera-raw-plug-supported-cameras.html)...

LebedevRI:
Best i can tell is that the `DMC-FZ85` name was used interchangeably in some announcements/reviews.
Unless they've changed the camera name in a firmware version, i guess there is no `DMC-FZ85`.

kmilos:
> Best i can tell is that the `DMC-FZ85` name was used interchangeably in some announcements/reviews.

Yep, I think some reviewers made a proverbial slip of the tongue, this model never existed as such AFAICT... Done in https://github.com/darktable-org/rawspeed/pull/748

PR #770, opened 2024-12-30
RafDecoder: set CFA pattern before cropping correction

kmilos:
The CFA is correctly adjusted according to crop in `applyCorrections()`, and this was resetting it.


dllu:
@LebedevRI can you kindly review and merge this? In https://github.com/darktable-org/darktable/issues/18073 we have several GFX 100S users who would be very grateful :) As a GFX 100S owner, let me know if there is anything else I can do to help test and/or support this.

EDIT: I guess kmilos has write access too so can merge directly, which would be nice.

LebedevRI:
Like i have asked numerous times before, where's the commit message? Where's at least the PR blurb? :)

PR #775, opened 2025-02-13
Update Sony ILCE-1M2 color matrix

kmilos:
Changed in ADC 17.2 since https://github.com/darktable-org/rawspeed/pull/768

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -13915,9 +13915,9 @@
 		<Sensor black="512" white="16383"/>
 		<ColorMatrices>
 			<ColorMatrix planes="3">
-				<ColorMatrixRow plane="0">8161 -2947 -739</ColorMatrixRow>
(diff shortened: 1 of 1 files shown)


LebedevRI:
I don't recall what's our stance on updating these things, since it breaks all the image edits downstream?
That we are free to do such updates, and consumers should cache the color matrix used?

kmilos:
> consumers should cache the color matrix used?

I _think_ this is the case, so existing edits are preserved and only newly imported ones are affected, right? @TurboGit @jenshannoschwalm

LebedevRI:
I mean more generally, since dt isn't, perhaps, the only user.

LebedevRI:
Looks like we've updated color matrix at least once before: #379
So i'm just going to pretend that it is fine to do.

kmilos:
> So i'm just going to pretend that it is fine to do.

Thanks, that's sensible, especially given that it didn't end up on stable branch and only dt 5.0 is using it so far most likely.

PR #780, opened 2025-02-28
Sigma BF normalization

kmilos:


LebedevRI:
Looks like this doesn't pass xmllint

kmilos:
> Looks like this doesn't pass xmllint

Argh... I guess we now have to also add "Sigma" as another vendor in contrast to existing all caps "SIGMA"... 😞

LebedevRI:
@kmilos thank you!

PR #786, opened 2025-03-25
Panasonic DC-GH7 support

kmilos:
Used ADC 16.4. Crop is unverified.


LebedevRI:
Looks like this is the only(?) v8 camera with still no samples.

kmilos:
Yes, and the two announced today.

PR #787, opened 2025-03-25
Panasonic DC-S9 support (data only)

kmilos:
Used ADC 16.4. Crop is unverified.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -10800,8 +10800,29 @@
 	<Camera make="Panasonic" model="DC-S1RM2" supported="no">
 		<ID make="Panasonic" model="DC-S1RM2">Panasonic DC-S1RM2</ID>
 	</Camera>
-	<Camera make="Panasonic" model="DC-S9" supported="no">
+	<Camera make="Panasonic" model="DC-S9">
 		<ID make="Panasonic" model="DC-S9">Panasonic DC-S9</ID>
+		<Crop x="0" y="0" width="0" height="0"/>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Hello. RPU is finally up. Please contribute the full sample set.

ascertrobw:
> Hello. RPU is finally up. Please contribute the full sample set.

I have an S9 - i can upload some I have, but it isn't a full sample set. Just the normal res, no crops or HHHR. I can make some time to do others but may be a few days before I get a chance. 

_**Edit:**_ have uploaded S9 3:2 raw and an HHHR raw. I'm not sure there's any point uploading other crops. Inspecting them on my PC, they only seem to crop the generated JPG. The raw file appears always to be 3:2 i.e. full sensor, so regardless of crop I seem to get the same size and aspect ratio raw. If others are needed, let me know.

LebedevRI:
> > Hello. RPU is finally up. Please contribute the full sample set.
> 
> I have an S9 - i can upload some I have, but it isn't a full sample set. Just the normal res, no crops or HHHR. I can make some time to do others but may be a few days before I get a chance.
> 
> _**Edit:**_ have uploaded S9 3:2 raw and an HHHR raw. I'm not sure there's any point uploading other crops. Inspecting them on my PC, they only seem to crop the generated JPG. The raw file appears always to be 3:2 i.e. full sensor, so regardless of crop I seem to get the same size and aspect ratio raw. If others are needed, let me know.

Thank you for contributing samples!

PR #798, opened 2025-04-09
Try to produce SARIF from linux clang

LebedevRI:
This is just horrible.

There really should be a way to still output the normal textual diagnostics to console,
but additionally place SARIF into a file named after the object file being produced,
into a specified prefix directory.


LebedevRI:
(Please consider upvoting:)
X-Ref: https://github.com/orgs/community/discussions/157136

PR #801, opened 2025-04-21
Add a Github Action to automatically build docs and deploy to Github Pages

### LINKED ISSUE #489: Website died
The rawspeed.org domain has not been renewed and now points towards random commercial websites. It was useful at very least to get https://rawspeed.org/CameraSupport.html aka the list of supported cameras.

There is a [Github action](https://github.com/marketplace/actions/sphinx-pages) that can build the HTML doc from Sphinx and deploy to Github pages upon new pushes, if paying for a domain name and maintaining a server is not an option, it seems low-overhead enough to keep an updated list of supported cameras.

aurelienpierre:
- Update Doxyfile.in for Doxygen 1.12, using `doxygen -u` (auto fix warnings regarding deprecations),
- Fix Sphinx -> Doxygen doc relative link in website,
- Add a workflow/Github Action to automatically build docs (user-side Sphinx and dev API Doxygen) every Sunday at midnight UTC and deploy to Github pages. You will need to enable pages from actions in the repo setting,
-  Fix #489 

Demo : https://aurelienpierreeng.github.io/rawspeed/

Note: the whole build process takes 1 min, deploying to Github Pages takes another 1 min 15. I have been careful to install only the minimum amount of dependencies, checkout only the last commit in repo, and use CMake to only build the docs targets.


LebedevRI (commented):
Please
1. Follow the pattern of the existing linux jobs:
```
    container:
      image: "debian:unstable-slim"
      options: --tmpfs /tmp:exec --tmpfs /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}:exec
```
2. Both of the things that get installed from PIP here are packaged in debian, let's use those instead.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,81 @@
...
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+      - name: Install Python deps for Sphinx
+        run: |
+          pip install -r docs/requirements.txt
+      - name: Setup Pages
+        id: pages
+        uses: actions/configure-pages@v2
```
LebedevRI, line 56: I don't think this (& `Upload page artifact`) can't be in the dependant job

aurelienpierre: I don't understand what you mean here. I have reused the workflow template for Hugo static website that I have used to update Github pages for 2 years now, this works just fine

LebedevRI: Let me reword.
Can the `- name: Setup Pages` and `- name: Upload page artifact` steps be moved into the `deploy:` job?

aurelienpierre: For `- name: Upload page artifact`, as far as I understand it, it can't be moved to another job because we upload the HTML content of the `build/docs` folder that will not be available anymore outside of this job. Then, `- name: Setup Pages` is needed before we do `- name: Upload page artifact` to be able to run it at all.

aurelienpierre:
I don't understand the purpose of 
```
 container:
      image: "debian:unstable-slim"
      options: --tmpfs /tmp:exec --tmpfs /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}:exec
```
and it makes all `apt install` commands fail

.github/workflows/docs.yml
LebedevRI: > For `- name: Upload page artifact`, as far as I understand it, it can't be moved to another job because we upload the HTML content of the `build/docs` folder that will not be available anymore outside of this job.

Right, ok, we can exfiltrate that data via upload-artifact, but that's kinda pointless.

> Then, `- name: Setup Pages` is needed before we do `- name: Upload page artifact` to be able to run it at all.

I'm not following, but i don't see where `actions/configure-pages` writes anything to `build/docs`,
especially since it's only input that takes a path, `generator_config_file`, isn't specified.

aurelienpierre:
My last change:

- uses Ubuntu packages for Sphinx/Sphinx RTD theme instead of PIP,
- collapses both jobs (build and upload to pages/deploy pages) since it seems there is no valid reason to reboot another container,
- changes the supported cameras page to use a table (see: https://aurelienpierreeng.github.io/rawspeed/CameraSupport.html), as it seems more readable to me than a list,
- does not use the Debian Unstable Slim image since I can't make it work with `apt-get` packages and anyway the current action doesn't build any piece of binary software.

docs/CameraSupport.rst
```diff
@@ -61,31 +62,38 @@ Camera Support
     print(
         "Any support is impossible without the samples.\nCurrently, |rpu-button-cameras| cameras have samples, with total count of |rpu-button-samples| unique samples. **Please contribute samples**!\n\n")
 
-    for make, models in unique_makes.items():
-        print(make)
-        print('=' * len(make), "\n")
+    with open('docs/cameras.csv', 'w', newline='') as csvfile:
+        csvwriter = csv.writer(csvfile, delimiter=',',
+                               quotechar='\"', quoting=csv.QUOTE_ALL)
```
LebedevRI, line 67: I agree that table is probably better, but i'm not sure where that file will be placed,
so can we just keep the `print()`s please? I think this should work:
```
        csvwriter = csv.writer(sys.stdout, delimiter=',',
                               quotechar='\"', quoting=csv.QUOTE_ALL)
```

.github/workflows/docs.yml
aurelienpierre: It seems `actions/configure-pages` prepares the metadata of the Github Page to which we will upload the artifacts. All actions for static website generators call it before uploading anything: https://github.com/actions/starter-workflows/tree/main/pages.

docs/requirements.txt
```diff
@@ -0,0 +1,2 @@
+sphinx
+sphinx_rtd_theme
```
LebedevRI, line 2: Seems unneeded now

docs/CameraSupport.rst
aurelienpierre: Printing a table in `.rst` is a PITA. The point of the CSV file (that is not commited, as per `.gitignore` rule, by the way) is to be able to use 

```
.. csv-table:: Supported cameras
   :file: cameras.csv
   :header-rows: 1
```

Previously, I tried using native RST table with prints (https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html) but I couldn't make it work because apparently it needs the `-` marks to be perfectly aligned and Python prints make it very hard.

The CSV file is placed in the `/build/docs` folder when Sphinx is called from `cmake`.

docs/requirements.txt
aurelienpierre: `pip install -r requirements.txt` is standard in Python world, it's a lazy and efficient way to let people on Win/Mac install everything at once and run `sphinx-build -M html docs build/docs` on their OS, should they need it. I can remove it if you really want, but I see the benefit (aside from Github auto-stuff) and not the harm.

docs/CameraSupport.rst
LebedevRI: Aha.
Can you not just call `sys.stdout.write(, "   ")` before each `csvwriter.writerow()`?

aurelienpierre: I'm not sure we understand each other. Either I print each line of the table with proper RST syntax so it gets (hopefully) parsed the right way (tried and failed, so far), or I write the rows to a CSV file and let the RST parser use it in one shot. Why would I print and write to CSV ?

LebedevRI:
@aurelienpierre thank you, i've merged after applying the changes i proposed.

PR #812, opened 2025-04-24
Fujifilm GFX100RF support

kmilos:
Confirmed color matrix is identical to GFX100 II using ADC 17.3

Addresses https://github.com/darktable-org/darktable/issues/18944


victoryforce:
It would be great if this was accepted and made it into darktable before the source cut (release tagging), which will be later today.

LebedevRI:
Lucky you.

Let's see what happens for the next major dt version...

PR #790, opened 2025-04-03
First revision of Panasonic RW2 version 8 Decompressor

kjyager:
Adapted from the Rust implementation in @cytrinox's [dnglab](https://github.com/dnglab/dnglab). Should make it possible to provide support for several newer Panasonic cameras: G9M2, S5II, S1RII, GH6, GH7, and S9.

In this branch, cameras.xml has only been updated for the G9M2.

PR #813, opened 2025-04-24
Panasonic DC-S1RM2 support (data only)

kmilos:
Used ADC 17.3. Crop is unverified.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -10797,8 +10797,41 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="Panasonic" model="DC-S1RM2" supported="no">
+	<Camera make="Panasonic" model="DC-S1RM2">
 		<ID make="Panasonic" model="DC-S1RM2">Panasonic DC-S1RM2</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
+		</CFA>
+		<Crop x="0" y="0" width="0" height="0"/>
+		<Sensor black="512" white="16380"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">9941 -4056 -871</ColorMatrixRow>
+				<ColorMatrixRow plane="1">-3745 11604 2433</ColorMatrixRow>
+				<ColorMatrixRow plane="2">-254 1132 5526</ColorMatrixRow>
+			</ColorMatrix>
+		</ColorMatrices>
+	</Camera>
+	<Camera make="Panasonic" model="DC-S1RM2" mode="3:2">
+		<ID make="Panasonic" model="DC-S1RM2">Panasonic DC-S1RM2</ID>
+		<CFA width="2" height="2">
+			<Color x="0" y="0">RED</Color>
+			<Color x="1" y="0">GREEN</Color>
+			<Color x="0" y="1">GREEN</Color>
+			<Color x="1" y="1">BLUE</Color>
+		</CFA>
+		<Crop x="0" y="0" width="0" height="0"/>
+		<Sensor black="512" white="16380"/>
+		<ColorMatrices>
+			<ColorMatrix planes="3">
+				<ColorMatrixRow plane="0">9941 -4056 -871</ColorMatrixRow>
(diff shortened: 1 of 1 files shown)


jdneumeyer77:
@kmilos  You beat me to it. I just received this camera Tuesday. I'll create some samples for RPU this weekend. I tested a few DPReview samples this past weekend with @kjyager's #790 PR and pulled the matrix from a DNG like this PR. It does work except hi-res mode. More testing and samples this weekend.

kjyager:
Thanks @jdneumeyer77. Looking forward to testing your samples. What kind of issues did you see with the hi-res?

jdneumeyer77:
@kjyager not on my machine ATM, but I think it was this exception was thrown. This is likely the case for Live Composite and Multiple-Exposure modes as well. My Panasonic S5 used a different raw version (7) for those and hi-res instead of version 6 for standard mode.

"Multiple Strips found: 2"

https://github.com/darktable-org/rawspeed/blob/7d583d6a1ce961ac57074450d7f1653b8d1d136d/src/librawspeed/decoders/Rw2Decoder.cpp#L126

kmilos:
Also note that the black & white levels here are for 14-bit raws, and the camera seems to also output 12-bit raws in burst mode, so this needs to be tested as well. If read from metadata these then need to be set to 0, like we do for 14/16-bit Fujifilms, or completely remove the `<Sensor>` element?

sergaderg:
The 12-bit mode I believe is used whenever the electronic shutter is used (both in high speed bursts as well as when silent mode is turned on)
There's also an APS-C crop mode when APS-C lenses are used which appears to crop the RAW file itself, as the files are about half the size. High res mode doesn't seem to work with APS-C lenses, at least with the current firmware.
I'm not sure if handheld high-res and the motion blur compensation settings in the high res mode effect the RAW at all, but those settings could be something to be aware of. I've managed to take at least one high-res shot that CaptureOne barfs on so there may be some kinks for Panasonic to work out still (or my camera is bad)

jdneumeyer77:
Electronic Shutter does not necessarily mean it's 12 bit. Electronic Shutter with burst is 12 bit. (I need to test burst + mechanical shutter).

All the crop modes are there, but I think they may apply to the jpeg only and some exif data (crop). The full raw is recorded and usable. 

@sergaderg I don't own any APS-C crop lenses so I'm unable to test that. That's interesting how some hi-res modes are supported and some are not even with Capture One with whom Panasonic actually worked with for the initial release. I've only recorded a handheld shot, which was actually on a tripod, because I forgot it about that feature. Good call I'll test normal hi-res mode. Maybe normal will work as-is and it'll only be handheld not supported. :crossed_fingers: 

I can edit multiple-exposure and the 12 bit burst shots just fine. I'll add mode 1/2 hi-res modes, live composite (later tonight) and mechanical shutter.

Any others: @kmilos  @kjyager ?

I'll try to make a some noise profiles for it and submit a PR for those.

LebedevRI:
Hello. RPU is finally up. Please contribute the full sample set.

jdneumeyer77:
Uploaded all the samples to RPU.

Live Composite mode works. Everything seems to work except hi-res mode. Based on the error it looks like there's two sub images in hi-res mode. Maybe it's as easy as picking the right one?

LebedevRI:
> Uploaded all the samples to RPU.

Thank you for contributing the samples!
I do note, it is very unfortunate that the images have different contents.
It would be really good if you could re-take them all, from a tripod, of the same static landsca-y scene.

kjyager:
Looks like the S1RM2 breaks hi-res shots into both horizontal and vertical strips, the first for RW2 V8 as far as I know. Allowing the decoder to run, despite the presence of vertical strips, results in a successfully decoded top half. I'll look at updating the decoder to handle the extra strips. Hopefully not too much work, as long as Panasonic arranged things the way one would expect :crossed_fingers:. 
image

kjyager:
> .... I'll look at updating the decoder to handle the extra strips. Hopefully not too much work, as long as Panasonic arranged things the way one would expect 🤞.

Already got it! I'll push the code in the morning (UTC-7).
image

jdneumeyer77:
Great! That image looks correct! Happy to do some more testing around hi-res variations to see if any fail. I'm testing noise profiles for the camera and will put up a PR for those by the end of the weekend.

kjyager:
Ok, I didn't end up having time to do it this morning, but it is done now. Just pushed the fix to support vertical strips to https://github.com/darktable-org/rawspeed/pull/790.

jdneumeyer77:
Looks like it's working for hi-res files I've tried. I've created several various hi-res RAWs which all seem to work. I'll test more this weekend if I can think of any.

PR #828, opened 2025-04-28
Remove uncompressed mode from Nikon entry level DSLRs

kmilos:
These cameras only output compressed. Checked the reference manuals for spec and options.

~~TBD: maybe check [models before D60](https://en.wikipedia.org/wiki/Template:Nikon_DSLR_cameras) as well...~~

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -3572,24 +3572,6 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="NIKON CORPORATION" model="NIKON D3300" mode="12bit-uncompressed" supported="no-samples">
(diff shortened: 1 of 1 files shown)


LebedevRI:
> Checked the reference manuals for spec and options.

Of every camera removed by this PR, correct?

kmilos:
> Of every camera removed by this PR, correct?

Indeed.

I'll push another change for D40 and D40X shortly, just checked those as well...

LebedevRI:
Do tell when this is ready.

kmilos:
Good to go. Only the 12-bit compressed sample for D5300 is actually missing.

It looks like there are phantom modes for other (non-"entry") Nikons as well, but I'll leave that for another PR/contributor.

LebedevRI:
@kmilos thank you.

PR #829, opened 2025-04-28
Remove uncompressed mode from more Nikon models

kmilos:
Covers the 1 series and enthusiast APS-C bodies (no uncompressed, only lossy and eventually lossless).

All related reference manuals checked.

Professional and flagship APS-C bodies do seem to support uncompressed + compressed, as do apparently FF bodies.


LebedevRI:
> All related reference manuals checked.

Based on that...

LebedevRI:
@kmilos thank you!

Issue #686, opened 2024-02-28
OM SYSTEM OM1 Mark II "14-bit" raw support


LebedevRI:
These are not normal Olympus compression,
and are not plain uncompressed,
and not a proper LJpeg (no `SOI` marker).

kmilos:
https://www.rawdigger.com/news/rawdigger-1-4-9-beta

https://www.fastrawviewer.com/comment/6056#comment-6056

kmilos:
One implementation is available in LibRaw since snapshot 202502: https://github.com/LibRaw/LibRaw/blob/master/src/decoders/olympus14.cpp

astmuc:
I discovered, that the OrfDecoder can decompress 14 bit files (mine is from pixl.us for OM-3) if one changes OlympusDecompressor.cpp at line 209 

from
```
  input.skipBytes(7);
```

to
```
  input.skipBytes(8);
```
Problem is, that the picture looks as if the colors are distorted. My question to the experts: Is it worth to follow this route ? If yes, what could be the reason for this color shift ?

![Image](https://github.com/user-attachments/assets/506efb80-f2f1-4c8d-911d-e504303512b9)

kmilos:
> If yes, what could be the reason for this color shift ?

Possibly a shift in the CFA pattern (RGGB to GBRG or similar).

astmuc:
Would method decodeMetaDataInternal (line 355+) in OrfDecoder.cpp be the right place to fiddle with the CFA pattern ?

kmilos:
> Would method decodeMetaDataInternal (line 355+) in OrfDecoder.cpp be the right place to fiddle with the CFA pattern ?

One way of doing it, yes... You can do something like [this](https://github.com/darktable-org/rawspeed/blob/e4d99acebd8d7fcf2922198e478e37a3988d3316/src/librawspeed/decoders/DngDecoder.cpp#L259) right after `parseCFA()` w/ different combos for (right.down) = (-1,0) / (0,-1) / (-1,-1) just to confirm this is indeed it.

astmuc:
First i want to thank you for your fast help and patience !
 Second, as you might have guesses already, i have another question: The changes with the CFA shift after parseCFA and setMetaData had no effect at all. Probably i did it not correct. Then i observed the values in the "raw black/white point" module (see first picture). The white point is at 2^12 - 1. So patched this line

```
          blackLevelSeparate1D(i) = blackEntry->getU16(j) * 4;
```
[...]

PR #833, opened 2025-05-04
extend Olympus/OM-System raw-decoder to allow 14-bit ORF files also

astmuc:
OrfDecoder and OrfDecompressor are extended to handle 14 bit files, by

* skip 8 instead of 7 bytes before decoding
* multiply white-, black-level and the detailed black-levels by 4

Tests were done with an installed darktable based on a Release-build (no asserts). The test pictures are described in #686

### DIFF
--- a/src/librawspeed/decoders/OrfDecoder.cpp
+++ b/src/librawspeed/decoders/OrfDecoder.cpp
@@ -167,7 +167,6 @@ int OrfDecoder::getBitsPerPixel() const {
   return result;
 }
 
-
 void OrfDecoder::decodeUncompressedInterleaved(ByteStream s, uint32_t w,
                                                uint32_t h,
                                                uint32_t size) const {
@@ -381,7 +380,7 @@ void OrfDecoder::decodeMetaDataInternal(const CameraMetaData* meta) {
         // range is the same)
         mRaw->whitePoint =
             *mRaw->whitePoint - (mRaw->blackLevel - blackLevelSeparate1D(0));
-        if ( getBitsPerPixel() == 14 ) {
+        if (getBitsPerPixel() == 14) {
           mRaw->whitePoint = *mRaw->whitePoint * 4;
           mRaw->blackLevel = mRaw->blackLevel * 4;
           for (int i = 0; i < 4; i++) {
--- a/src/librawspeed/decoders/OrfDecoder.cpp
+++ b/src/librawspeed/decoders/OrfDecoder.cpp
@@ -131,6 +131,23 @@ RawImage OrfDecoder::decodeRawInternal() {
     ThrowRDE("%u stripes, and not uncompressed. Unsupported.",
              raw->getEntry(TiffTag::STRIPOFFSETS)->count);
 
+  if (const int bitsPerPixel = getBitsPerPixel(); bitsPerPixel == 12) {
+    input.skipBytes(7);
+  } else if (bitsPerPixel == 14) {
+    input.skipBytes(8);
+  } else {
+    ThrowRDE("%i-bit images are not supported currently.", bitsPerPixel);
+  }
+
+  const OlympusDecompressor o(mRaw);
+  mRaw->createData();
+  o.decompress(input);
+
+  return mRaw;
+}
+
+int OrfDecoder::getBitsPerPixel() const {
+  int result = 12;
   if (mRootIFD->hasEntryRecursive(TiffTag::OLYMPUSIMAGEPROCESSING)) {
     // Newer cameras process the Image Processing SubIFD in the makernote
     const TiffEntry* img_entry =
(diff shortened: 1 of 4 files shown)


src/librawspeed/decoders/OrfDecoder.cpp
```diff
@@ -42,7 +42,6 @@
 #include "tiff/TiffTag.h"
 #include <array>
 #include <cassert>
-#include <cstdint>
```
kmilos: AFAIK this header is [needed explicitly since GCC 13](https://www.gnu.org/software/gcc/gcc-13/porting_to.html#header-dep-changes) (and is also inserted there by IWYU), please restore it everywhere.

astmuc: Done. I beg your pardon for the fault and multiple fix commits. I did not remember making these changes and it took me a little bit to recognize the scope of the import cleanup action.

LebedevRI:
@astmuc thank you!
@krid thank you for the extra verification samples!

I'm not sure the white level handling is right,
i don't get this whole thing:
```
        // Adjust whitelevel based on the read black (we assume the dynamic
        // range is the same)
        mRaw->whitePoint =
            *mRaw->whitePoint - (mRaw->blackLevel - blackLevelSeparate1D(0));
```
Is that the right thing to do for //all// ORF raws, or just some older ones?
Also, `mRaw->whitePoint = (1U << 12) - 1;` looks suspect for 14-bit,
but i don't know if making it dynamic on the actual bit depth would be more correct.

Best would be to read white level from metadata, but i don't see it there.

astmuc:
@LebedevRI The ORF decompression was difficult for me to understand. Better yet, i did not... My theory for the patch was: The 14 bit files are created from several 12 bit shots with an algorithm implemented in the body firmware. And the company probably decided to store black and white levels from the original shots. So, i added the multiply by 4. During my experiments i changed `mRaw->whitePoint = (1U << 12) - 1`  to use a 14 bit shift and it had no effect. It seems to be a starting value only. I studied the libraw olympus decoder `olympus14.cpp` but didn't find references to white or black level.

TLDR i do not know where the bw levels are in the metadata and i did my changes assuming that the existing code is fine.

LebedevRI:
Wrong black and/or white levels aren't always glaringly obvious,
if they are just slightly off they manifest as e.g. magenta highlights,
or gray-ish black areas.
I don't know what the right values are (both for 14-bit, and 12-bit).

astmuc:
Yes, this was my situation during after discovering the 'skip 8 instead of 7 bytes' feature.  After applying the  'multiply bw-levels by 4 trick' they displayed "correct" values for me. Obviously, i cannot prove this. Please note, the PR did not change the decoding of 12 bit files.

kmilos:
> I studied the libraw olympus decoder `olympus14.cpp` but didn't find references to white or black level.

@astmuc I think metadata parsing is also elsewhere in LibRaw, IIRC in open/identify routines early on...

astmuc:
I found the reading of black-levels in libraw file `olympus.ccp` at line 607. The logic seems identical to the rawspeed code. But I found nothing in this file regarding white-level ... I think its is the best to keep the existing heuristic like code and the magic value `mRaw->whitePoint = (1U << 12) - 1` as this is adjusted later with a factor of 4 for 14 bit files.

kmilos:
@astmuc FYI: https://github.com/LibRaw/LibRaw/blob/36fa8bb9c7e009a46bc293ba9e5fb174bf9eb0fb/src/metadata/olympus.cpp#L487 and https://github.com/LibRaw/LibRaw/blob/36fa8bb9c7e009a46bc293ba9e5fb174bf9eb0fb/src/metadata/identify.cpp#L2263-L2270

astmuc:
@kmilos i checked all olympus shots of pixl.us for OM-3 and OM-1 MarkII with `rstest`. They contain the tag (0x805) value  4095 always. Means, we could replace the magic constant (`  mRaw->whitePoint = (1U << 12) - 1;`) with this value from the metadata. But the computation
```
        // Adjust whitelevel based on the read black (we assume the dynamic
        // range is the same)
        mRaw->whitePoint =
            *mRaw->whitePoint - (mRaw->blackLevel - blackLevelSeparate1D(0));
```
cannot be replaced due to this discovery. Do you think this change is worthwile ?

kmilos:
@astmuc I guess you could try parsing [0x0805 ](https://exiftool.org/TagNames/Olympus.html#ImageProcessing) if it exists, otherwise fall back to `(1U <<12) - 1`. As for the fudge, I guess we just keep the status quo for now - while it is confusing, as @LebedevRI  hints, it's probably there for some legacy (and forgotten?) reason, so best not to mess with it unless someone wants to check _**all**_ Olympus/OM System RPU samples _**in darktable**_ for magenta highlights when removed...

LebedevRI:
> As for the fudge, I guess we just keep the status quo for now - while it is confusing, as @LebedevRI hints, it's probably there for some legacy (and forgotten?) reason, so best not to mess with it unless someone wants to check _**all**_ Olympus/OM System RPU samples _**in darktable**_ for magenta highlights when removed...

Yup...
Essentially, my understanding/guess is that there now has been,
what, 4 different Panasonic compressed raw codecs?
that fudge was added way back when, for the first generation,
so perhaps the newer generations changed things up,
and that fudge should only be applied to some older generations only.

This line of thinking applies to every new model/firmware version of every manufacturer.

PR #834, opened 2025-05-06
Clean up some Pentax data entries

kmilos:
* K100D (Super) and K110D don't output DNG
* K20D and K200D do have both PEF+DNG samples on RPU

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -11608,42 +11608,6 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="PENTAX" model="PENTAX K100D" supported="no-samples">
-		<ID make="Pentax" model="K100D">Pentax K100D</ID>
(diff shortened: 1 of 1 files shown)


data/cameras.xml
```diff
@@ -11608,42 +11608,6 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="PENTAX" model="PENTAX K100D" supported="no-samples">
```
LebedevRI, line 11611: I think in cases like these ,it would be best to still keep the entry, but with `mode="dng"`,

kmilos: There is no DNG output for these two models, so these entries are totally superfluous. The PEF ones are covered by similar existing ones, but w/ make="PENTAX Corporation".

LebedevRI:
@kmilos thank you!

PR #835, opened 2025-05-11
Mark Panasonic DC-S1RM2 and DC-S9 as not supported

kmilos:
ATM there is no codec for v8 Panasonic cameras.


LebedevRI:
I guess the lesson is: there should be `mode=no-no-samples`.
I should have just added a new field `samples="no"` i guess.

PR #853, opened 2025-05-20
Add Panasonic DC-G97 alias

kmilos:
Confirmed ADC 17.3.1 identifies it as DC-G99

See also https://www.dpreview.com/news/8943078360/panasonic-g97-zs99-tz99-refresh-usb-c

Addresses https://github.com/darktable-org/darktable/issues/18836

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -11393,6 +11393,7 @@
 			<Alias>DC-G90</Alias>
 			<Alias>DC-G91</Alias>
 			<Alias>DC-G95D</Alias>
+			<Alias>DC-G97</Alias>
 			<Alias>DC-G99</Alias>
 			<Alias>DC-G99D</Alias>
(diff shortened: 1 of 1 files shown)


LebedevRI:
This adds DC-G97, but that issue is about DC-G97M.
RPU has a sample for DC-G97M but not DC-G97.

kmilos:
The RPU sample _is_ DC-G97, as its exif data confirms... No idea where the "M" by the OP comes from?!

kmilos:
> No idea where the "M" by the OP comes from?!

Ah, it's just how they market different kits: https://www.panasonic.com/fr/consumer/appareils-photo-et-camescopes/lumix-g-hybrides/lumix-g.html

(In general I find Panasonic's marketing and proliferation of model numbers for different markets very confusing, to put it mildly.)

PR #858, opened 2025-05-22
Panasonic V8: completely replace prefix code decoder

LebedevRI:


LebedevRI:
Ok so TLDR is that this isn't going to work, that isn't a Huffman code, and it's not even a Prefix code.

PR #884, opened 2025-08-21
Add Ricoh GR IV placeholder

kmilos:


data/cameras.xml
```diff
@@ -18193,6 +18193,9 @@
 			<Alias id="GR IIIx HDF">RICOH GR IIIx HDF</Alias>
 		</Aliases>
 	</Camera>
+	<Camera make="RICOH IMAGING COMPANY, LTD." model="RICOH GR IV" mode="dng" supported="unknown-no-samples">
```
LebedevRI: I think this just shouldn't have a `supported` attr,
as per precedent because it is a DNG.

kmilos: Ok, I'll remove and squash in a while.

LebedevRI:
Unmark as draft?

kmilos:
Can't squash though, away from my dev env...

PR #885, opened 2025-08-22
Add support for Fujifilm X-E5

horazont:
The Fujifilm X-E5 uses almost the same hardware build as the X-T50, so I copy/pasted the X-T50 block. Raw samples have been submitted to raw.pixls.us and I locally tested this change to produce sensible results.

Let me know if this is a sensible approach.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -16327,8 +16327,43 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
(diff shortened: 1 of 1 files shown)


kmilos:
Just waiting for the samples to get approved then.

I have verified using ADC 17.5 the color matrix is indeed identical to other models, and had the change ready: https://github.com/darktable-org/rawspeed/commit/ae24a192750166245b8d6b78db5aac966137da1a

LebedevRI:
Thank you for contributing samples!

PR #892, opened 2025-09-11
PefDecoder: support rebranded Samsung cameras

kmilos:
Addresses https://github.com/darktable-org/darktable/issues/19228 and complements #877.

Hope this is trivial enough to make an exception for touching the code, apologies for not catching it earlier... ;)


LebedevRI:
I did not verify that this works, but yes, this is rather trivial and is fine :)

kmilos:
Thanks, I did test it this time! ;)

PR #905, opened 2025-10-28
Nikon modes cleanup

kmilos:
* Coolpix B700 only has 12bit-uncompressed (verified in manual)
* Coolpix E8400 and E8800 only have 12bit-uncompressed (verified in manual)
~~* D800 and D800E "extended" modes are not needed~~
~~* D100 12bit-uncompressed sample is available, fix mode detection~~

@LebedevRI ~~Feel free to skip code changes and cherry pick only the XML ones - B700, E8400/E8800 & D800(E) only; D100 does require the code change though.~~

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -6084,27 +6084,6 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="NIKON" model="COOLPIX B700" supported="no-samples">
-		<ID make="Nikon" model="COOLPIX B700">Nikon Coolpix B700</ID>
-		<CFA width="2" height="2">
-			<Color x="0" y="0">RED</Color>
-			<Color x="1" y="0">GREEN</Color>
-			<Color x="0" y="1">GREEN</Color>
-			<Color x="1" y="1">BLUE</Color>
-		</CFA>
-		<Crop x="0" y="0" width="0" height="0"/>
-		<Sensor black="200" white="4000"/>
-		<Hints>
-			<Hint name="coolpixmangled" value=""/>
-		</Hints>
-		<ColorMatrices>
-			<ColorMatrix planes="3">
-				<ColorMatrixRow plane="0">14387 -6014 -1299</ColorMatrixRow>
-				<ColorMatrixRow plane="1">-1357 9975 1616</ColorMatrixRow>
-				<ColorMatrixRow plane="2">467 1047 4744</ColorMatrixRow>
-			</ColorMatrix>
-		</ColorMatrices>
-	</Camera>
 	<Camera make="NIKON" model="COOLPIX B700" mode="12bit-uncompressed">
 		<ID make="Nikon" model="COOLPIX B700">Nikon Coolpix B700</ID>
 		<CFA width="2" height="2">
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -5938,29 +5938,8 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
-	<Camera make="NIKON" model="E8400" mode="12bit-compressed" decoder_version="3" supported="no">
-		<ID make="Nikon" model="E8400">Nikon E8400</ID>
-		<CFA width="2" height="2">
-			<Color x="0" y="0">BLUE</Color>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Would it please be possible to split this into two separate stand-alone parts,
first one without those codechanges and without the XML cnages that require them,
and the rest?

kmilos:
Makes sense, will do a bit later (might even be three parts). I also want to check why E8400/E8800 are "supported=no", the samples are there...

kmilos:
This PR now contains only the XML changes, I'll open other ones separately.

data/cameras.xml
```diff
@@ -6284,29 +6206,8 @@
...
-		</Hints>
-		<ColorMatrices>
-			<ColorMatrix planes="3">
-				<ColorMatrixRow plane="0">7971 -2314 -913</ColorMatrixRow>
-				<ColorMatrixRow plane="1">-8451 15762 2894</ColorMatrixRow>
-				<ColorMatrixRow plane="2">-1442 1520 7610</ColorMatrixRow>
-			</ColorMatrix>
-		</ColorMatrices>
-	</Camera>
-	<Camera make="NIKON" model="E8800" mode="12bit-uncompressed" decoder_version="3" supported="no">
```
LebedevRI, line 6308: Git is a wonderful tool. Learn to use it, and it will answer many questions that should not be left to the reviewer, but be explained in the commit message. Really.

This was due to
```
commit 6eb86833b29bfb7972db2a0f3f8d3d9f024f8dac
Author:	Roman Lebedev <lebedev.ri@gmail.com>  Mon Sep 12 21:39:01 2016
Committer:	Roman Lebedev <lebedev.ri@gmail.com>  Fri Nov  4 17:57:41 2016

Cameras.xml: disable all nikon cameras with questionable whitelevel.

Ugh.

If you are affected by this, you *need* to report this to the
https://redmine.darktable.org/projects/darktable/issues/new
and provide one raw file per each camera's raw mode.
```

LebedevRI:
> * 
> * D800 and D800E "extended" modes are not needed

And this is why nobody should be merging crap commits with no commit messages.
The reason why those modes are there is in the git log. Can you tell me why they are there?

data/cameras.xml
kmilos: Ah, should've checked the history for those indeed, sorry. I assumed it was because the non-existing compressed samples were "missing". I'll revert the support for those then, but keep the compressed mode removal.

I did check the history for D800/D800E extended mode entries, and couldn't find a reason for their introduction...

LebedevRI:
> > * D800 and D800E "extended" modes are not needed
> 
> And this is why nobody should be merging crap commits with no commit messages. The reason why those modes are there is in the git log. Can you tell me why they are there?

... The caveat being, the commit that added them didn't _really_ explain why they are needed in the first place:
https://github.com/darktable-org/rawspeed/commit/52294de2f962a94d856fa100ad90a2898a510e70

kmilos:
Hm, back draft it Is... I'll check if those output anything other than full frame.

LebedevRI:
I'm going to guess it might be about raw images taken during Live View / video recording, but i didn't dig up the relevant information.

kmilos:
D800(E) indeed has several crop modes (no samples at RPU), so dropping those changes.

data/cameras.xml
```diff
@@ -6284,29 +6242,8 @@
...
-		<ColorMatrices>
-			<ColorMatrix planes="3">
-				<ColorMatrixRow plane="0">7971 -2314 -913</ColorMatrixRow>
-				<ColorMatrixRow plane="1">-8451 15762 2894</ColorMatrixRow>
-				<ColorMatrixRow plane="2">-1442 1520 7610</ColorMatrixRow>
-			</ColorMatrix>
-		</ColorMatrices>
-	</Camera>
 	<Camera make="NIKON" model="E8800" mode="12bit-uncompressed" decoder_version="3" supported="no">
-		<ID make="Nikon" model="Coolpix E8800">Nikon Coolpix E8800</ID>
```
LebedevRI, line 6309: See 8eacc642b070fce7a096e772a3ab002e040a0f64.

LebedevRI:
@kmilos thank you!

PR #908, opened 2025-11-07
Fix some Sony aliases

kmilos:


LebedevRI:
This is missing some words. Why is this change correct?

kmilos:
Sorry.

The `id` attribute is used to sub/normalize Exif model strings for alias models, like
```
<Alias id="EOS Rebel SL2">Canon EOS REBEL SL2</Alias>
```
In these cases, we actually don't want to substitute anything, there's no need (and is in fact wrong) to make it look exactly like the base model string.

LebedevRI:
So IOW these are actual standalone cameras, not just marketing rebranding, and we don't want to pretend that they aren't?

kmilos:
Yep.

PR #915, opened 2025-11-13
Hasselblad H4D-60 and H6D-50c placeholders

kmilos:
[This H4D sample](https://raw.pixls.us/getfile.php/2631/nice/Hasselblad%20-%20Hasselblad%20H4D%20-%2016bit%20(4:3).3FR) is actually the H4D-60, but it cannot be properly supported as it is actually mistagged in Exif as H3D (perhaps later firmware fixed this?).

ADC identifies it when converting ("Hasselblad H4D-60" in `LocalizedCameraModel`), and so does Phocus SW when importing, but neither doesn't seem to fix up the `Model` tag up when saving to DNG/FFF...

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -17983,6 +17983,9 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="Hasselblad" model="Hasselblad H4D-60" supported="unknown-no-samples">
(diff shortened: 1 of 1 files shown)


LebedevRI:
CI seems rather unhappy.

kmilos:
Sorry. Fixed. Groan.

PR #923, opened 2025-11-26
CI: fix for deprecated macOS Intel job

kmilos:
Replace with the only available macos-15-intel runner substitute.

Also bump to default macOS 15 runner XCode version and deployment target to match.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -195,8 +195,8 @@ jobs:
       fail-fast: false
       matrix:
         compiler:
-          - { os: macos-15, family: XCode, version: 16.3, deployment_target: 15.2, CC: cc, CXX: c++ } # AArch64, LLVM19, native
-          - { os: macos-13, family: XCode, version: 15.2, deployment_target: 13.5, CC: cc, CXX: c++ } # x86_64,  LLVM16, native
+          - { os: macos-15,       family: XCode, version: 16.4, deployment_target: 15.3, CC: cc, CXX: c++ } # AArch64, LLVM19, native
+          - { os: macos-15-intel, family: XCode, version: 16.4, deployment_target: 15.3, CC: cc, CXX: c++ } # x86_64,  LLVM19, native
         flavor: [ ReleaseWithAsserts, Release ]
     uses: ./.github/workflows/CI-macOS.yml
     with:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -195,8 +195,8 @@ jobs:
       fail-fast: false
       matrix:
         compiler:
-          - { os: macos-15, family: XCode, version: 16.3, deployment_target: 15.2, CC: cc, CXX: c++ } # AArch64, LLVM19, native
-          - { os: macos-13, family: XCode, version: 15.2, deployment_target: 13.5, CC: cc, CXX: c++ } # x86_64,  LLVM16, native
+          - { os: macos-15,       family: XCode, version: 16.4, deployment_target: 15.3, CC: cc, CXX: c++ } # AArch64, LLVM19, native
+          - { os: macos-15-intel, family: XCode, version: 16.4, deployment_target: 15.3, CC: cc, CXX: c++ } # x86_64,  LLVM19, native
```
LebedevRI, line 199: Specifically here, why is `deployment_target` also being bumped?
In other two jobs, it's correct to bump it.

kmilos: Can I in turn ask how was XCode 16.3 and deployment target of 15.2 picked previously?

I was kinda guided by the matrix [here](https://en.wikipedia.org/wiki/Xcode#Xcode_15.0_-_16.x_(since_visionOS_support))...

The macos-14 job seems to have the logic of "latest XCode for that major macOS version and minimum minor macOS version that XCode runs on"? It seemed to me to be true for the exisiting 16.3/15.2 combo as well at the time... (Caveat emptor: I don't own or develop for macOS.)

LebedevRI: > Can I in turn ask how was XCode 16.3 and deployment target of 15.2 picked previously?
> 
> I was kinda guided by the matrix [here](https://en.wikipedia.org/wiki/Xcode#Xcode_15.0_-_16.x_(since_visionOS_support))...
> 
> The macos-14 job seems to have the logic of "latest XCode for that major macOS version and minimum minor macOS version that XCode runs on"? It seemed to me to be true for the exisiting 16.3/15.2 combo as well at the time... (Caveat emptor: I don't own or develop for macOS.)

Yes. There's basically just two configs here: newest XCode targeting it's native OSX version + some (newest?) xcode targeting OSX version that is claimed as being the oldest supported (13.5)

I'm just saying: keep the `deployment_target: 13.5` bit.

kmilos: > targeting OSX version that is claimed as being the oldest supported (13.5)

This is the part I don't get, claimed where? Shouldn't it be now 14.0 then, as 13 is EOL'd?

So how about 16.2/14.0? (I.e. the combo that'd be used on the macos-14 runner had it been possible for Intel.)

LebedevRI: You know what, i really don't care about this os, so i'm just not going to bother.

kmilos: Fine by me 😉

PR #926, opened 2025-12-13
Fix build with Clang on Window

stevenlr:
This PR fixes two issues that made it impossible to build rawspeed on Windows with Clang (and probably MSVC too, though I have not tested that).

Note that I'm currently building the library without any of the external dependencies, and hence there could exist other problematic code paths that I'm not hitting.


LebedevRI:
Please split commit in two and mention the actual error in each commit message.

stevenlr:
Done in latest revision.

LebedevRI:
@stevenlr please rebase

stevenlr:
Done!

PR #928, opened 2025-12-16
Sony ILCE-7M5 support

kmilos:
Partly addresses [#19917](https://github.com/darktable-org/darktable/issues/19917) _for lossless only_.

ADC 18.1 color matrix is not marked as final.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -14029,9 +14029,9 @@
 		<Sensor black="512" white="16383"/>
 		<ColorMatrices>
 			<ColorMatrix planes="3">
-				<ColorMatrixRow plane="0">7460 -2365 -588</ColorMatrixRow>
-				<ColorMatrixRow plane="1">-5687 13442 2474</ColorMatrixRow>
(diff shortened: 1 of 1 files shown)


tomzorz:
One of the failed tests seems to be a temporary networking issue, the other one seems to be failing on issues unrelated to these changes - I wonder if we could re-run those. Otherwise, is there anything else we can do to help? (Asking as a new 7m5 owner and developer.)

kmilos:
Seems the color matrix has been updated/finalized in ADC 18.2.

kmilos:
@LebedevRI Thanks! How about RPU, it seems several cameras should be in the upload queue?

kmilos:
@LebedevRI Thanks for putting up all the other recent samples, but these still seem to be missing?

LebedevRI:
> @LebedevRI Thanks for putting up all the other recent samples, but these still seem to be missing?

I don't remember why i didn't validate some set at that time. Did that now.

PR #932, opened 2026-01-30
Canon EOS Rebel T7i alias normalization

kmilos:


kmilos:
@LebedevRI These (Nikon Coolpix B700/P950/P1000) now show up as missing in https://raw.pixls.us/?missing#repo - the script might be case sensitive? Alternatively rename [on disk](https://raw.pixls.us/data/Nikon/)?

PR #938, opened 2026-02-25
Panasonic alias cleanup

kmilos:
I believe FZ92/FZ93 was a typo - no reference to these models is to be found outside of rawspeed/darktable...

Cf. https://av.jpn.support.panasonic.com/support/global/cs/dsc/download/TZ90_TZ91_TZ92_TZ93_ZS70/index2.html

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -11687,10 +11687,9 @@
 		<Sensor black="139" white="4095"/>
 		<Aliases>
 			<Alias>DC-TZ91</Alias>
+			<Alias>DC-TZ92</Alias>
+			<Alias>DC-TZ93</Alias>
 			<Alias>DC-ZS70</Alias>
-			<Alias>DC-FZ91</Alias>
-			<Alias>DC-FZ92</Alias>
-			<Alias>DC-FZ93</Alias>
 		</Aliases>
 		<ColorMatrices>
 			<ColorMatrix planes="3">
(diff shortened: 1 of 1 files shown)


data/cameras.xml
```diff
@@ -7202,6 +7202,9 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
+	<Camera make="Panasonic" model="DMC-FX180" supported="unknown-no-samples">
+		<ID make="Panasonic" model="DMC-FX180">Panasonic DMC-FX150</ID>
```
LebedevRI, line 7206: Is `DMC-FX180` supposed to be an alias in `DMC-FX150`? I guess not?

data/cameras.xml
```diff
@@ -11625,7 +11628,7 @@
 		<Sensor black="143" white="4095"/>
 		<Aliases>
 			<Alias>DC-GF10</Alias>
-			<Alias>GF90</Alias>
```
LebedevRI, line 11628: https://www.dpreview.com/products/panasonic/slrs/panasonic_dcgf10
> The Panasonic DC-GF10 (also known as the GF90 in Japan)

LebedevRI (changes requested):

data/cameras.xml
kmilos: Most likely, but web seacrh doesn't turn up much...

data/cameras.xml
kmilos: That doesn't say much to me, who's to say they didn't omit 'DC-' for readibility? Cf. https://helpx.adobe.com/camera-raw/kb/camera-raw-plug-supported-cameras.html and https://github.com/LibRaw/LibRaw/blob/777f20ae21c611a78021bd051fbbf1e71eae78f2/src/metadata/normalize_model.cpp#L565

LebedevRI:
Digged around a bit, maybe this is fine..

LebedevRI:
@kmilos thank you!

PR #939, opened 2026-03-06
Remove ID remapping nops

kmilos:
Retain/fix ones where UniqueCameraModel differs.

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -7152,7 +7152,6 @@
 		</ColorMatrices>
 	</Camera>
 	<Camera make="Panasonic" model="DMC-CM1">
-		<ID make="Panasonic" model="DMC-CM1">Panasonic DMC-CM1</ID>
 		<Crop x="0" y="0" width="0" height="0"/>
 		<Sensor black="142" white="4095"/>
 		<ColorMatrices>
@@ -7164,7 +7163,6 @@
 		</ColorMatrices>
 	</Camera>
 	<Camera make="Panasonic" model="DMC-CM1" mode="3:2">
-		<ID make="Panasonic" model="DMC-CM1">Panasonic DMC-CM1</ID>
 		<Crop x="0" y="0" width="0" height="0"/>
 		<Sensor black="142" white="4095"/>
 		<ColorMatrices>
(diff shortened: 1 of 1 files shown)


LebedevRI:
So what's the testing story with this, is this hand-made, how do we know the changes are correct?
Regardless, can you please split this into smaller PR's:
* Removal of entire `<camera>` entries
* Change of `<ID>`
* Removal of redundant `<ID>`
?

kmilos:
Makes sense, will split those.

In terms of testing, how about if I check `identify` output before and after once for each of  those cases?

LebedevRI:
Hm, thinking about it, let's reorder these then:
> * Removal of entire `<camera>` entries
> * Removal of redundant `<ID>`
> * Change of `<ID>`

... with the last one being the only one i don't know how to validate.

Thank you!

kmilos:
This is now only the second step above,

LebedevRI:
> This is now only the second step above,

Thanks! I'll see how i can check this.

data/cameras.xml
```diff
@@ -17002,7 +16752,6 @@
 		<ID make="Contax" model="N Digital">Contax N Digital</ID>
 	</Camera>
 	<Camera make="ImBack" model="ImB35mm" mode="chdk">
-		<ID make="ImBack" model="ImB35mm">ImBack</ID>
```
LebedevRI, line 17005: ???

LebedevRI:
Ok, so the idea here is to match
```
<Camera\smake="([^"]+)"\smodel="([^"]+)"([^>]*)>\n\s*<ID\smake="\1"\smodel="\2">\1 \2<\/ID>
```
and just drop the ID tag for these i guess.

LebedevRI:
@kmilos thank you.
I merged locally, then checked out the file from the previous commit,
done the regex thing, and amend-merged the resulting changes.

data/cameras.xml
kmilos: Yeah, this one can be dropped as well - it's very niche HW and doubt there is much SW support for it out there to try and preserve something... The original ID entry w/ only the maker name in the UniqueCameraModel was incorrect to begin with IMHO.

PR #943, opened 2026-03-10
Remove unnecessary DNG entries

kmilos:
Make, Model (and UniqueCameraModel) tags for these are already "clean" so there is no ID remapping done, as for most other devices outputting DNG.

For example, we stopped bothering adding GoPro models at some point.


kmilos:
LGE Nexus 5X is somewhat of a special case: the noise profile in darktable uses the original "LGE" make string, so this remapping to "LG" is actually breaking things there. I really doubt it is relied on elsewhere...

(I'll squash this to a single commit later.)

PR #947, opened 2026-03-12
docs: support table generation tweaks

kmilos:
Actually do as the comment suggests for Leaf cameras only, and only up to the first delimiter.

Prevents e.g. Kodak DCS Pro SLR/n from being split.

We also don't want duplicated aliases for e.g. Hasselblad X1D II 50C, X2D 100C, X2D II 100C.


kmilos:
There is one more issue w/ the table IMHO: "unknown-no-samples" entries are shown w/ a check mark as supported... I can maybe take stab in this PR as well to mark them w/ a "?" unless you have a preference on how to tackle this?

PR #949, opened 2026-03-14
Adjust Panasonic UniqueCameraModel mappings

kmilos:
Change to canonical model as set by ADC in resulting DNG's UniqueCameraModel tags.

For all other Panasonic cameras the canonical model is used already in ID mapping, or [ID mappings are to be removed](https://github.com/darktable-org/rawspeed/pull/939) since the camera entry is already matching the canonical model.

AFAICT this is used in

1. [Setting `canonical_id` from DNGs](https://github.com/darktable-org/rawspeed/blob/f2f50db1b2c7b1dfcdb20de8c69fcfd6b58eef94/src/librawspeed/decoders/DngDecoder.cpp#L727) (not clear who uses this; darktable doesn't)

2. [dngmeta.sh darktable tool for making new camera.xml entries](https://github.com/darktable-org/darktable/blob/d61bb2f12c0efabc51cb58a55afd6b70bab1bfb6/tools/dngmeta.sh#L106)

I can split into individual PRs per camera if preferred...

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -11617,7 +11617,7 @@
 		</ColorMatrices>
 	</Camera>
 	<Camera make="Panasonic" model="DC-GX880">
-		<ID make="Panasonic" model="DC-GX880">Panasonic DC-GX880</ID>
+		<ID make="Panasonic" model="DC-GX880">Panasonic DC-GF10</ID>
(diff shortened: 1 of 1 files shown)


LebedevRI:
@kmilos thank you!

PR #970, opened 2026-06-12
Nikon ZR support

kmilos:
Basically a (ADC confirmed) [copy of Z6_3](https://www.dpreview.com/reviews/nikon-zr-in-depth-review-compact-full-frame-video-camera-6k-60)...

Public domain samples at https://rawdb.dnglab.org/sets/Nikon/ZR

### DIFF
--- a/data/cameras.xml
+++ b/data/cameras.xml
@@ -4241,8 +4241,23 @@
 			</ColorMatrix>
 		</ColorMatrices>
 	</Camera>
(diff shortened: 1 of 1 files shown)


kmilos:
Looks like the super shiny Clang 22.1.7 in MSYS2 started disliking something...

P.S. And the macOS runners have been crying for a timeout increase for a while...

PR #963, opened 2026-04-18
LJpeg: support predictor modes 2–7

da-phil:
This PR is supposed to address https://github.com/darktable-org/rawspeed/issues/189, https://github.com/darktable-org/rawspeed/issues/191 and https://github.com/darktable-org/rawspeed/issues/258 and was created with the help of Claude.

**Summary**

rawspeed's LJpeg decoder previously supported only predictor mode 1 (left neighbor). ITU-T T.81 defines seven predictor modes, and several camera manufacturers — notably DJI (Mavic 3S, Mavic 3 Pro, etc.) and Blackmagic — use mode 6 in their DNG files. This change adds support for all seven modes and handles the associated tile geometry those cameras use.

**Predictor modes 2–7**

* Added a computePrediction(mode, Ra, Rb, Rc) helper that computes the seven ITU-T T.81 predictions. Arithmetic is done in int32_t to avoid overflow in modes 4–6, where Ra + Rb - Rc can transiently exceed the 16-bit range; the final result wraps via uint16_t cast per the standard.
* decodeRowN() is extended to accept predMode and prevStripe. For mode 1, the existing fast path (no memory access to the previous row) is preserved. For modes 2–7, the three 2D neighbors (Ra = left, Rb = above, Rc = above-left) are looked up from prevStripe; at the start of a row the above-left neighbor is not available, so Rc = Rb per the JPEG specification.
* decodeN() now tracks isFirstRow per restart interval. The first row of each interval always uses predictor mode 1 (per T.81: the standard mandates horizontal prediction for the first row). For subsequent rows, prevStripe is a CroppedArray2DRef into the already-decoded portion of the output image. No extra allocation is needed. Inverted tile reshape (LJpegDecoder)

**Inverted tile reshape**

DJI DNGs present tiles in a non-standard geometry: the JPEG SOF3 frame is wider than the declared tile (e.g. 8000×1500 for a 4000×3000 tile). Each JPEG row encodes two tile rows concatenated side-by-side — a widthPack = jpegFrameDim.x / tileW packing. This is the inverse of the Adobe-style reshape rawspeed already handles (MCU > 1×1).

**Design trade-offs**

[...]

PR #971, opened 2026-06-21
Add JPEG XL (DNG 1.7 / Compression 52546) decompressor

MaykThewessen:
## What

Adds a libjxl-backed `JpegXlDecompressor` so rawspeed can decode DNG 1.7 tiles compressed with **JPEG XL** (TIFF `Compression` tag **52546**). It's wired into `DngDecoder` chunk acceptance and `AbstractDngDecompressor` dispatch, modelled on the existing lossy-JPEG path (`JpegDecompressor`). Gated behind a new `WITH_JPEGXL` CMake option (default **ON**, mirroring `WITH_JPEG`/`WITH_ZLIB`); libjxl is discovered via `pkg-config`.

## Why

This is what **Apple ProRAW** uses on the 48 MP main camera of recent iPhones (16 Pro, 17 Pro): `PhotometricInterpretation = LinearRaw (34892)`, 3 channels, 10-bit, tiled. Those files currently fail with `No RAW chunks found`, because compression `52546` is dropped as unsupported.

## Testing

Verified by decoding **real iPhone 16 Pro Max and iPhone 17 Pro** ProRAW DNGs end-to-end (all tiles, 8064×6048, clean artifact-free output; 10-bit samples scale to the full 16-bit range, `whitePoint` 65535). Built with `-DRAWSPEED_ENABLE_WERROR=ON` and the `WITH_JPEGXL` default (no extra flags); clang-format clean.

Honest scope: this verifies decode-without-error and visually-correct output, **not** pixel-exact output vs a reference decoder. The 2×2 interleaved-CFA JPEG XL variant is **not** exercised (my samples are LinearRaw, 1×1).

## Relationship to #755 / #516

There's an existing JPEG XL effort: **#755** by @kmilos (and issue **#516**). #755 is a proof-of-concept its author noted was stalled and open for adoption, and it currently hangs on real files due to a `JxlDecoderSetImageOutBuffer` byte-count-vs-element-count bug (reported on #755). This PR is an independent, self-contained, CI-passing implementation offered in that spirit — **not** trying to compete. Happy to consolidate however the maintainers prefer: fold this into #755, co-author, or close this in favour of #755. The goal is simply to get JPEG XL ProRAW support landed. Thanks to @kmilos for the original prototype.

## Scope

Only the JPEG XL piece. The *other* Apple ProRAW path — lossless JPEG with predictor mode 7 (iPhone ≤ 15 Pro and the telephoto cameras) — is a separate concern handled by **#963**.


dholth:
[toucan.zip](https://github.com/user-attachments/files/29411238/toucan.zip) is a sample 2×2 interleaved-CFA JPEG XL DNG compressed to be small for convenience. I took this image. Decompression should work the same whether it's lossy or lossless; then reshape to get the original Bayer pattern.
My camera produces 19M Panasonic .RW2 files, becoming 14M lossless LJPEG-92 DNGs, 12M lossless JPEG XL DNGs.
For some reason Darktable doesn't do the same dead-sensor-pixel processing when I convert to DNG compared to camera-native RW2.

I've learned that my camera is one of a few that marks its bad pixels by setting them to 0 in the RAW instead of e.g. the black threshold of 143. It is possible to translate this list to opcodes in a DNG but Adobe DNG converter doesn't. Many cameras interpolate bad pixels before writing the RAW.

For lossless I've found that the JPEG-XL modular predictor "9=leftleft" is 93% as good as 4-up for CFA data.

MaykThewessen:
## Follow-up verification: CFA variant + a second real file

The PR description flagged that the **CFA JPEG XL variant was not exercised** (my original samples were all LinearRaw 1×1). I've now closed that gap by testing against a real **single-channel CFA** DNG 1.7 / JPEG XL file (Panasonic DMC-GX85, `toucan.lossy-d2.00-3ev.dng`) alongside the iPhone 17 Pro LinearRaw file, using this branch built standalone with libjxl 0.11.2.

### End-to-end decode (full `decodeRaw()` via `rstest -c -d`)

| File | Photometric | Framing | Decode | Output |
|------|-------------|---------|--------|--------|
| Panasonic GX85 | **Color Filter Array** (1ch, 16-bit, `CFAPattern2 = 2 1 1 0`) | **bare codestream** (`FF 0A`) | OK | P5 4620×3464 |
| iPhone 17 Pro | LinearRaw (3ch) | **ISOBMFF container** (`00 00 00 0C 4A 58 4C 20`) | OK | P6 8064×6048 |

Both decode cleanly through the single `JxlDecoderSetInput` path. Worth highlighting: **two JPEG XL framings occur in the wild** — Apple wraps each tile in the ISOBMFF container box, while this Panasonic/Adobe-DNG-Converter file is a bare codestream. `JxlDecoder` auto-detects both, so the current `JxlDecoderSetInput`-based approach is correct, but it means a CFA-only test sample (bare codestream) and a container sample exercise different framing branches — both belong in any test corpus.

### CFA layout note

This GX85 file stores the CFA as a **single-channel, full-resolution mosaic** (decodes to 4620×3464, 1 channel), not a 4-channel half-resolution interleaved superpixel image. So there's no de-interleave step for this variant — the decoded plane is the Bayer mosaic directly, indexed by `CFAPattern2`. A "2×2 interleaved" half-res encoding would be a separate representation; if a converter emits that form too, both layouts are worth covering.

### Bit-depth / scaling check

The decoder requests `JXL_TYPE_UINT16`, which makes libjxl map the codestream's normalized `[0,1]` range to `[0,65535]`. I checked this against both files' actual codestream bit depth and DNG metadata:

| File | `BitsPerSample` | codestream `bits_per_sample` | `WhiteLevel` | decoded white |
|------|-----------------|------------------------------|--------------|---------------|
| GX85 | 16 | 12 | 63232 | ~63232 (3952 × 16.004) |
| iPhone 17 | 10 | 16 | 65535 | full-range |

The key point: **`BitsPerSample` is not the value-range indicator — `WhiteLevel` is.** Apple declares 10-bit but stores full-range values (`WhiteLevel` 65535); Adobe declares 16-bit (`WhiteLevel` 63232). In both cases the encoder pre-scales the codestream so libjxl's `[0,1]→[0,65535]` mapping lands exactly on `WhiteLevel`. So the **unconditional `JXL_TYPE_UINT16` request is correct as-is** — decoded values match `WhiteLevel` for both conventions, no decoder-side bit-depth handling needed.

(Conversely, keying the output depth off `BitsPerSample` would be wrong: scaling the iPhone file to its declared 10-bit would put white at ~1023 against a `WhiteLevel` of 65535, i.e. ~64× too dark. The current code avoids that by always producing full-range output.)

### Re: the toucan dead-pixel question

For anyone following the dead-sensor-pixel thread: the GX85 DNG carries **only `OpcodeList3 = WarpRectilinear`** — no `OpcodeList1`, no `FixBadPixels*`. So the missing defect correction vs. the camera-native RW2 is **not** a decode issue here; the RW2→DNG converter simply didn't emit bad-pixel opcodes. (rawspeed applies `OpcodeList1` Stage-1 opcodes in `decodeRaw()`; darktable's host side implements only `OpcodeList2`/`3`. Nothing to apply if the converter wrote none.)

Net: this branch decodes both real-world DNG 1.7 JPEG XL framings (container + bare codestream) and both channel layouts (1ch CFA + 3ch linear) correctly, with values matching the declared `WhiteLevel`.

---

*Edited: corrected the bit-depth section. An earlier version of this comment suggested the decoder might need to "honor `bits_per_sample`" to scale output to the DNG's declared depth. That was wrong — I verified that doing so would regress the Apple files (BitsPerSample=10 but WhiteLevel=65535). The current unconditional full-range `UINT16` output is correct for both Apple and Adobe DNGs.*

dholth:
FYI [digikam](https://www.digikam.org/) has a RW2 converter that does an excellent job copying metadata but doesn't do what I want re jxl. For me a perfect converter would either store the CFA data as lossless JXL; or interpolate out the dead pixels, arrange planes into a 2x2 layout, before doing (lossy, distance=0.5) compression.
rawspeed's RW4 record bad pixel [code is here](https://github.com/darktable-org/rawspeed/blob/develop/src/librawspeed/decompressors/PanasonicV4Decompressor.cpp#L263). 
Where is the handling for [Exif.Image.RowInterleaveFactor](https://exiv2.org/tags.html#:~:text=0xc71f) and [Exif.Image.ColumnInterleaveFactor](https://exiv2.org/tags.html#:~:text=0xcd43) which are `[2, 2]` if each of [B G, G R] are stored in quadrants of the image?
It is also possible to [use threads when decoding](https://libjxl.readthedocs.io/en/latest/api_threads.html) a single JXL tile.

dholth:
FYI Adobe's DNG SDK includes sample images, [mirror here.](https://github.com/Autoenhance-ai/adobe-dng-sdk/tree/main/sample_files)

MaykThewessen:
Thanks for catching that, @dholth. Let me go through your points in order, starting with a correction to my own earlier comment.

---

### Correction: the toucan sample is 2x2 interleaved, and my earlier "OK" decode was wrong

I owe a correction to what I wrote earlier in this thread. I claimed the toucan sample (`toucan.lossy-d2.00-3ev.dng`) was "a single-channel full-resolution mosaic" with "no interleave factors" and "no de-interleave step for this variant", and I called the decode "successful."

That was wrong. `exiftool` on the file shows `RowInterleaveFactor = 2` and `ColumnInterleaveFactor = 2` (CFA pattern GBRG). What I was actually observing was that rawspeed decoded without crashing, which is not the same thing as decoding correctly. The decoder silently ignored the interleave tags, leaving the four half-resolution color planes stacked in quadrant layout instead of being woven back into the Bayer mosaic. The output was a scrambled mosaic. You were right.

---

### Interleave handling: no code existed, now there is a fix (local branch, not yet pushed)

To answer your question directly: before my fix, there was no interleave handling at all. `RowInterleaveFactor` (0xC71F) was defined in `TiffTag.h` but never read; `ColumnInterleaveFactor` (0xCD43) was not even defined. Any 2x2-quadrant-interleaved CFA JXL DNG decoded scrambled.

I've implemented a de-interleave pass on a local branch (not pushed yet). Key design decisions:

**Whole-frame post-pass, not per-tile.** The de-interleave runs after all tiles are assembled, not inside the per-tile decode loop. The reason: Adobe's own canonical 2x2-interleaved test vector, `03_jxl_bayer_raw_integer.dng` from the [adobe-dng-sdk sample files](https://github.com/Autoenhance-ai/adobe-dng-sdk/tree/main/sample_files) you linked (10240x7168, 1280x1440 tiles, R=C=2), has its row-field boundary at row 3584. That is not a multiple of 1440, so a single JXL tile straddles two color planes. Per-tile de-interleave is therefore geometrically impossible; the interleave is a property of the assembled stored frame.

**Index mapping.** Forward map stored->final: `fy = within_field_row * R + field_row_index`, symmetric for columns with C. This matches the DNG 1.7.1.0 spec and the reference implementation in `dng_sdk`'s `dng_row_interleaved_image::MapRow`. Final dimensions equal stored dimensions (interleave reorders, never resizes). Non-divisible field sizes are handled the `dng_sdk` way (earlier fields absorb the remainder row/column).

**Ordering relative to crop.** De-interleave runs before `ActiveArea`/`DefaultCropOrigin` is applied, matching `dng_sdk`'s ordering.

**Verified on both sample types:**

| File | Layout | Result before fix | Result after fix |
|---|---|---|---|
| `03_jxl_bayer_raw_integer.dng` (Adobe) | 1280x1440 tiles, R=C=2 | scrambled 4-quadrant mosaic | coherent Bayer mosaic, quadrant seam energy collapses |
| toucan DNG (strip-based) | single strip, R=C=2 | scrambled | coherent |
| `01_jxl_linear_raw_integer.dng` (Adobe) | LinearRaw, no interleave | (unaffected) | unchanged, fast-path no-op |

A unit test covers the index math, including the spec's 4x4 / R=C=2 worked example and non-divisible field sizes.

I've put the fix up as a branch stacked on this PR: https://github.com/MaykThewessen/rawspeed/pull/1 . Happy to fold it straight into this PR instead if that's cleaner for review; whatever fits how you and the maintainers want to handle it.

---

### Bad pixels (your point 1): confirmed, nothing to fix in this PR

Your read is correct. `PanasonicV4Decompressor` collects zero-valued pixels into `mBadPixelPositions` when `zero_is_bad` is true (which it is for all current Panasonic cameras). The DNG path, including this PR's `JpegXlDecompressor`, never marks zeros, so `mBadPixelPositions` stays empty and `fixBadPixels()` is a no-op. The defect correction is lost because Adobe DNG Converter doesn't translate the camera's zero-marked bad pixels into `OpcodeList1 FixBadPixels` opcodes; it's a converter omission, not a decode bug in this PR. Nothing to address here.

---

### Threading (your point 3): orthogonal to correctness

Single-tile multithreaded JXL decode via `JxlResizableParallelRunner` is a reasonable performance enhancement. rawspeed already parallelizes across tiles via OpenMP in the DNG decompressor, so per-tile threading would mainly benefit single-tile or few-tile images. Makes sense as a follow-up; not a correctness issue for this PR.

---

### Separate finding: float JXL DNGs assert

While testing I hit a pre-existing issue unrelated to interleave: the current decoder writes decoded tiles through the UINT16 accessor unconditionally. Adobe sample `02_jxl_linear_raw_float.dng` is a float LinearRaw DNG; it fails with an assertion. The fix needs an F32 output path: `JXL_TYPE_FLOAT` output format and the F32 buffer accessor when the image is float. Noting it here so it doesn't get lost, but it's a separate fix from interleave.

dholth:
I was able to compile darktable with this patch and look at my .DNG photo, it seems to work great.

MaykThewessen:
Woew great that it worked! Can't see your .dng yet

erdostamasa:
I tested this branch with darktable on Fedora using a Samsung Galaxy S23 Expert RAW file:

* DNG 1.7
* JPEG XL compression (`Compression = 52546`)
* 4000 × 3000
* 16-bit, three-channel `LinearRaw`

After building darktable with this RawSpeed branch and `libjxl`, the file imported successfully, opened in darkroom, could be edited normally, and exported successfully. I did not observe obvious channel-order, scaling, strip-layout, or image-corruption problems in this sample.

So from a Samsung Galaxy S23 compatibility perspective, I can confirm that the implementation works for at least this real-world file. 

But I would  encourage  careful maintainer review before merging. Several commits explicitly credit Claude Opus as a co-author, and to me it looks like all comments and replies by the author are fully AI generated. The working result is valuable, but generated code can look convincing while hiding incorrect assumptions or poorly understood edge cases.

I think this should receive normal line-by-line human review, maintainability review, and automated coverage for the supported DNG layouts rather than relying primarily on extensive explanatory comments or successful decoding of a few samples. The earlier correction regarding the interleaved CFA sample is a useful example of why decoding without an error is not sufficient evidence of correctness.

For the S23 LinearRaw case I tested, however, the current branch works.

MaykThewessen:
Thanks for testing, and the criticism is fair. I took it as a reason to re-check the
decoder rather than defend it, and it turned up a real bug that your S23 file could not
have caught.

`JxlDecoderSetImageOutBitDepth` was never called, so libjxl applied its default
`JXL_BIT_DEPTH_FROM_PIXEL_FORMAT` and rescaled every decoded tile to fill the full uint16
range. The DNG's BlackLevel/WhiteLevel are keyed to BitsPerSample, so a 12-bit codestream
decoded about 16x too bright, a 14-bit one about 4x. Your sample is 16-bit, which is
precisely the case where that default is a no-op. It is a good illustration of your point:
the file imported, opened, edited and exported cleanly, and it would have done all of that
with the bug present at any other depth.

Fixed in 24775359. The integer path is now pinned to `JXL_BIT_DEPTH_FROM_CODESTREAM`, and
a codestream whose depth disagrees with the DNG's BitsPerSample is rejected with both
numbers in the message instead of decoded at a guessed scale. Two more on the same path:
the tile extent was computed with unsigned subtraction, so an origin at or past the image
bounds wrapped to ~4e9 rather than being rejected, and a codestream carrying extra
channels passed the channel check and had them silently dropped by libjxl.

On coverage: the geometry and the codestream-vs-DNG agreement rules now live in a pure
header with no libjxl dependency, with 20 unit tests that run on every build and need no
sample file. They cover both regressions above, edge and corner tile truncation,
channel-count and extra-channel rejection, and buffer sizing. That is not the same as
end-to-end coverage of the supported DNG layouts, which needs files in the rawspeed sample
set, and I would rather that be maintainer-directed than something I add unilaterally.

On the AI point: I do use it, and the trailers say so. The code is mine to defend, and I
would rather be judged on whether it holds up than on how it was produced, so line-by-line
review is welcome. I will keep my comments shorter.

erdostamasa:
If you can't even be bothered to write your own comments i think this PR should be closed and deleted. But I think you are just an automated account so there is no one to talk to

MaykThewessen:
I let Claude Code help me, sorry for that

da-phil:
> If you can't even be bothered to write your own comments i think this PR should be closed and deleted. But I think you are just an automated account so there is no one to talk to


@MaykThewessen I really appreciate your enthusiasm and work to address open issues in this repo, I hope you will not take offense by expressing my similar feelings in the following reply to @erdostamasa 

Although I also use LLMs to help me get things done, I find it off-putting when people even use them to write replies in conversations. To me it feels like genuine human interaction gets lost in all the AI slop around us.
I'm already chatting with LLMs at work far too often, and the last thing I want from human interactions is to end up talking to LLMs in disguise. Just my 2 cents (and I believe that the rawspeed maintainer has an even stronger feeling about this).

martinus-523:
Is there any update on this? JXL support would be awesome!

PR #972, opened 2026-06-27
ARW: add Sony "ARW 6.0" lossy decoder (compression 32766; A7R VI, A7 V)

Kab1r:
## Summary

Adds `SonyArw6Decompressor`, implementing Sony's "ARW 6.0" lossy codec
(TIFF Compression `32766`) used by the **A7R VI** (ILCE-7RM6) and **A7 V**
(ILCE-7M5) in their *Compressed* / *Compressed HQ* RAW modes. These formats were
previously undecodable.

## What this adds

- `SonyArw6Decompressor` (+ `SonyArw6LogToLinear` LUT) — the codec itself.
- `compression == 32766` dispatch in `ArwDecoder`, the active-area crop, and a
  dimension-ceiling bump for the 66.8 MP sensor (which also enables the A7R VI
  lossless tiled LJpeg path).
- `ILCE-7RM6` camera entry in `cameras.xml`. (The `ILCE-7M5` entry already
  exists.)
- A fuzz target for the new decompressor.

## Pipeline

Container tiles → per-row GCLI bitplane entropy decode → inverse reversible
5/3 wavelet (LL0 + 3 detail levels) → log→linear LUT → per-line colour
conversion → YCC→RGGB mosaic compose.

Coefficients are `int16`; tiles and their (up to four) components decode in
parallel under OpenMP (a no-op when it's unavailable). The per-tile sub-header
marker carries a 1-byte format tag that varies by body (`'0'` on the A7R VI,
`'A'` on the A7 V), so only its fixed `"000"` suffix is matched; the
black/white-level domain (1024/32800) and the `DefaultCrop`-derived active area
are codec-universal.

## Reverse-engineering & validation

The format was reverse-engineered clean-room by observing the decoder's
input/output behavior; the 4096-entry log→linear table was recovered as a
black-box input→output sampling, not transcribed.

The full pipeline is validated **byte-exact (MAE 0)** against the reference
decoder output (Adobe DNG) on both bodies, full-frame and APS-C, in both lossy
and HQ modes.
Malformed input is rejected via `ThrowRDE` rather than crashing,
hanging, or invoking undefined behavior (a fuzz target is included).

I have also personally tested the decoder with a dirty build of Darktable built against this branch and was able to edit images successfully.

## AI usage

This contribution was produced with
substantial AI assistance. The AI was used as a tool by me; the codec was
reverse-engineered from, and validated stage-by-stage against, the reference
decoder output, and I take full responsibility for the change and stand behind
its correctness.

PR #983, opened 2026-08-22
Implement Sony ARW6 decoder

### LINKED ISSUE #968: A7R VI camera support
I've uploaded the 6 sample images to pixls in full frame and aps-c mode

Settings 1/50 f10 ISO 100

*Raw lossless compressed
*Raw compressed HQ
*Raw compressed

abbradar:
This is an LLM-assisted (Claude Fable) port of my [ARW6 reference decoder](https://github.com/abbradar/arw6_decode) to rawspeed. The reference decoder's initial implementation was LLM-assisted too, but I've completely reviewed it and rewrote most parts by hand.

The comments were preserved as is from the reference decoder, but I restructured the code to live in a single .cpp, as other decoders here do.

One questionable thing is a custom `Plane` class which is an owning Array2DRef; I didn't find anything similar in the existing code, but I may be wrong. I've also introduced OpenMP parallelism for tiles and channels. It's clunky because of the exception passing; I didn't use OpenMP before, but AFAIU that's a well-known wart; I'd also love to be corrected.

Fixes https://github.com/darktable-org/rawspeed/issues/968


abbradar:
Formatted and also cleaned up a bit; decided against an arrow proxy clunkiness after all, and removed the identity LUT.

kmilos:
x-ref https://github.com/darktable-org/rawspeed/pull/972

abbradar:
Ouch, failed to find it while searching before, thanks

PR #985, opened 2026-08-28
FujiDecompressor: add lossy-compressed RAF decoding support

### LINKED ISSUE #232: Fujifilm X-T4 (non-lossless) "compressed" raw support
I have now uploaded three .RAF files (on a X-rite colorchecker classic):
1) uncompressed
2) Compressed lossless
3) Compressed

_Originally posted by @frli8848 in https://github.com/darktable-org/darktable/issues/4943#issuecomment-625797717_

### LINKED ISSUE #255: Fujifilm X-S10 (non-lossless) "compressed" raw support
Much like #232, the same appears to happen for X-S10 (254)

### LINKED ISSUE #332: Fujifilm X-E4 (non-lossless) "compressed" raw support
Probably the same problem as other recent Fujifilm non-lossless compressed raw support, like https://github.com/darktable-org/rawspeed/issues/255 and https://github.com/darktable-org/rawspeed/issues/232 
There is a file with non-lossless compression from X-E4 in the raw files repository.

### LINKED ISSUE #366: Fujifilm GFX100 (non-lossless) "compressed" raw support
Same as #232/#255/#332

igoreskub:
FujiDecompressor previously only handled uncompressed and lossless-compressed Fuji RAF files; lossy-compressed ("compressed" quality setting) files were rejected at header validation.

Ports the lossy decode path from dnglab/rawler's Rust implementation (rawler/src/decoders/fuji_decompressor.rs), which already unifies lossless and lossy decoding in one algorithm, reconciled against rawspeed's existing lossless-only architecture:

- Parse the per-line quantization base (q_base) array that follows the block-offset table in lossy files.
- Add FujiQTable, generalizing the single quantization table into four: a "main" table rebuilt whenever q_base changes between lines, plus three static tables (fixed q_base 0/1/2) selected per-pixel by local gradient magnitude.
- Wire table selection into the sample-decode hot path (fuji_decode_interpolation_even/odd_inner, fuji_decode_sample, fuji_quant_gradient), gated throughout by header.isLossless() so lossless-file behavior is unchanged.

Cross-validated pixel-exact against dnglab/rawler's independent decoder across 449 real-world files (one X-series body, both lossless- and lossy-compressed modes across a firmware change) at 25 spot-check coordinates per file.

Fixes a related regression found during that testing: passing q_bases.data() directly into Array1DRef crashed on every lossless file, since an empty vector's .data() can be null (observed on libc++) and Array1DRef's invariant check requires non-null even at size 0. Fixed with a static sentinel byte.

Includes an opt-in pixel spot-check diagnostic (RAWSPEED_FUJI_DIAG env var, off by default) used to produce the validation results above; zero overhead when unset. Also includes
src/utilities/dnglab-crossvalidate/, the standalone script (with README) used to run that cross-validation, so others can reproduce or extend it.

Closes #232
Closes #255
Closes #332
Closes #366
Closes #406

### DIFF
--- a/src/librawspeed/decompressors/FujiDecompressor.cpp
+++ b/src/librawspeed/decompressors/FujiDecompressor.cpp
@@ -46,9 +46,11 @@
 #include <bit>
 #include <cassert>
 #include <cstdint>
+#include <cstdio>
 #include <cstdlib>
 #include <cstring>
 #include <string>
+#include <tuple>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


kmilos:
x-ref https://github.com/darktable-org/rawspeed/pull/972#issuecomment-4820206677

igoreskub:
Thanks for the context, @kmilos — no pressure on timeline.

This represents real, tested work (449 files cross-validated against an independent decoder, two compression modes, one regression caught along the way), so happy to have it sit until someone has bandwidth. I'm also glad to stick around and maintain it going forward — respond to feedback, fix issues, extend to other bodies if needed.

Is there anyone else with merge rights who might take a look, or best to just leave this open for now?

